PR #9, opened 2024-02-27
Moved NEMO_SKILLS_CODE env variable definition

i-vainn:
Now `NEMO_SKILLS_CODE` env variable is automatically determined as `NEMO_SKILLS_CODE = str(Path(__file__).absolute().parents[1])` in `pipeline/launcher.py`.


Kipok (approved):
Thanks!

PR #10, opened 2024-02-29
Openai model

i-vainn:
Added `OpenAIModel` class to `nemo_skills/inference/server/model.py` and prompt config `nemo_skills/inference/prompt/openai_chat.yaml` for it.

### DIFF
--- a/nemo_skills/inference/prompt/code_sfted.yaml
+++ b/nemo_skills/inference/prompt/code_sfted.yaml
@@ -4,6 +4,6 @@ few_shot_examples:
 system: "\
   You're an expert Python programmer and mathematician. \
   Help the user to solve this problem using code when necessary. \
-  Make sure to put the answer (and only answer) inside \\boxed{{}}."
+  Make sure to put the answer (and only answer) inside \\boxed{}."
 user: "{question}\n\n{context}"
 prompt_template: "System:\n{system}\n\nUser:\n{user}Assistant:\n{generated_solution}"
--- a/nemo_skills/inference/prompt/code_sfted.yaml
+++ b/nemo_skills/inference/prompt/code_sfted.yaml
@@ -1,9 +1,9 @@
 few_shot_examples:
   template: ""
-  delimiter: "\n\n\n\n\n\n"
 system: "\
   You're an expert Python programmer and mathematician. \
   Help the user to solve this problem using code when necessary. \
   Make sure to put the answer (and only answer) inside \\boxed{}."
 user: "{question}\n\n{context}"
 prompt_template: "System:\n{system}\n\nUser:\n{user}Assistant:\n{generated_solution}"
+stop_phrases: ["\n\n\n\n\n\n"]
--- a/nemo_skills/inference/prompt/code_sfted.yaml
+++ b/nemo_skills/inference/prompt/code_sfted.yaml
@@ -1,7 +1,9 @@
-delimiter: "\n\n\n\n\n\n"
-prefix: "\
-  System:\n\
+few_shot_examples:
+  template: ""
+  delimiter: "\n\n\n\n\n\n"
+system: "\
   You're an expert Python programmer and mathematician. \
   Help the user to solve this problem using code when necessary. \
-  Make sure to put the answer (and only answer) inside \\boxed{}.\n\n"
-template: "User:\n{question}\n\n{context}Assistant:\n{generated_solution}"
+  Make sure to put the answer (and only answer) inside \\boxed{{}}."
+user: "{question}\n\n{context}"
+prompt_template: "System:\n{system}\n\nUser:\n{user}Assistant:\n{generated_solution}"
--- a/nemo_skills/inference/prompt/openai_chat.yaml
+++ b/nemo_skills/inference/prompt/openai_chat.yaml
@@ -1,7 +1,7 @@
 delimiter: "\n\n\n\n\n\n"
-prefix: "\
-  [{\"role\": \"system\", \"content\": \"\
+system: "\
   You're an expert Python programmer and mathematician. \
   Help the user to solve this problem using code when necessary. \
-  Make sure to put the answer (and only answer) inside \\\\boxed{}.\"}, "
-template: "{{\"role\": \"user\", \"content\": \"{question}\\n\\n{context}\"}}]"
+  Make sure to put the answer (and only answer) inside \\boxed{}.\n\n"
+prompt_template: "System:\n{system}\n\n{examples}"
+example_template: "User:\n{question}\n\n{context}Assistant:\n{generated_solution}"
--- a/nemo_skills/inference/prompt/openai_chat.yaml
+++ b/nemo_skills/inference/prompt/openai_chat.yaml
@@ -1,7 +1,0 @@
-delimiter: "\n\n\n\n\n\n"
-system: "\
-  You're an expert Python programmer and mathematician. \
-  Help the user to solve this problem using code when necessary. \
-  Make sure to put the answer (and only answer) inside \\boxed{}.\n\n"
-prompt_template: "System:\n{system}\n\n{examples}"
-example_template: "User:\n{question}\n\n{context}Assistant:\n{generated_solution}"
--- a/nemo_skills/inference/prompt/openai_chat.yaml
+++ b/nemo_skills/inference/prompt/openai_chat.yaml
@@ -1,0 +1,7 @@
+delimiter: "\n\n\n\n\n\n"
+prefix: "\
+  [{\"role\": \"system\", \"content\": \"\
+  You're an expert Python programmer and mathematician. \
+  Help the user to solve this problem using code when necessary. \
+  Make sure to put the answer (and only answer) inside \\\\boxed{}.\"}, "
+template: "{{\"role\": \"user\", \"content\": \"{question}\\n\\n{context}\"}}]"
--- a/nemo_skills/inference/prompt/utils.py
+++ b/nemo_skills/inference/prompt/utils.py
@@ -14,7 +14,7 @@
 
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional
+from typing import Any, Dict, List, Optional
 
 import yaml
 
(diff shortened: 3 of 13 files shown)


nemo_skills/inference/prompt/openai_chat.yaml
```diff
@@ -0,0 +1,7 @@
+delimiter: "\n\n\n\n\n\n"
```
Kipok, line 1: How about we make our code more general to handle cases like this. E.g. one way to do that is to add a new "system" parameter that's reserved for cases when system message is well defined. And to keep supporting our code where system is just a prefix, we can add a new "prompt_template" parameter that will be something like this:

```
prompt_template: "System: {system}\n\n{examples}"
```
where examples is what we currently compose in utils that uses `template` and `delimiter`. This way we can customize prefix/suffix of prompt with more flexibility and also get to know an actual system message in case it needs to be used parsed out like when we use openai or when we prepare nemo chat datasets.

Maybe also rename the current `template` to `example_template` to make it clearer

Kipok (changes requested):

nemo_skills/inference/prompt/openai_chat.yaml
Kipok, line 1: So it looks like this format is still not general enough to support all of the use-cases we have. Here are the things
that we need to be able to represent:

1. No "chat" format

2. NeMo chat format where explicit tokens need to be specified

3. Non-nemo chat models, where system/user might have different separator format, e.g.
   ```
       "<<SYS>>\n{system}\n<</SYS>>\n\n"
       "<|system|>\n{system}</s>\n"
       "[INST] {text} [/INST]",
       "<|user|>\n{text}\n<|assistant|>\n",
  
   ```

3. OpenAI (or other cloud providers) where you don't need to list tokens, but need the message separated into system/user

And in all cases we want a convenient support for few-shot examples, so that prompt can be constructed from them easily.

Here is what I propose to use and configs that are equivalent to currently supported use-cases. Might be necessary
to make some adjustments.

Current code_base.yaml:

```
# template for few-shot examples. 
# We need to separate the last part (actual question that's missing a generated_solution), because
# it needs to follow-up with special tokens for chat models that we have to explicitly handle
few_shot_examples:
  template: "Question:\n{question}\n\n{context}My solution:\n{generated_solution}"
  delimiter: "\n\n\n\n\n\n"
# will generate {examples} field based on the above
# examples = template.format(example1_info) + delimiter + ... + template.format(exampleN_info)

# system part of cloud LLM api call
system: ""

# user part of cloud LLM api call
user: "\
    Here are some examples of questions and solutions followed by a new question that you need to solve.\n\
    Make sure to put the answer (and only answer) inside \\boxed{}.\n\n\
    {examples}\n\n\n\n\n\n\
    Question:\n{question}\n\n{context}"

# ignored for cloud LLMs, but used for NeMo or other local models to explicitly format turn separators
prompt: {system}{user}My solution:\n
```

Current code_sfted.yaml. Note that few-shot examples is not used here, since we never want to show few-shots
for finetuned models (it does not make it any better, but takes up the context)

```
# system part of cloud LLM api call
system: "You're an expert Python programmer and mathematician. \
  Help the user to solve this problem using code when necessary. \
  Make sure to put the answer (and only answer) inside \\boxed{}."

# user part of cloud LLM api call
user: "{question}\n\n{context}"

# ignored for cloud LLMs, but used for NeMo or other local models to explicitly format turn separators
prompt: System:\n{system}\n\nUser:\n{user}Assistant:\n
```

From igitman/chat-format branch code_base.yaml

```
# template for few-shot examples. 
# We need to separate the last part (actual question that's missing a generated_solution), because
# it needs to follow-up with special tokens for chat models that we have to explicitly handle
few_shot_examples:
  template: "Question:\n{question}\n\n{context}My solution:\n{generated_solution}"
  delimiter: "\n\n\n\n\n\n"
# will generate {examples} field based on the above
# examples = template.format(example1_info) + delimiter + ... + template.format(exampleN_info)

# system part of cloud LLM api call
system: ""

# user part of cloud LLM api call
user: "\
    Here are some examples of questions and solutions followed by a new question that you need to solve.\n\
    Make sure to put the answer (and only answer) inside \\boxed{}.\n\n\
    {examples}\n\n\n\n\n\n\
    Question:\n{question}\n\n{context}"

# ignored for cloud LLMs, but used for NeMo or other local models to explicitly format turn separators
prompt: <extra_id_0>System\n{system}\n<extra_id_1>User\n{user}<extra_id_1>Assistant\nMy solution:\n
```

Kipok (changes requested):

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+
+@dataclass
+class PromptConfig:
+    few_shot_examples: FewShotExamples = MISSING
+    prompt_template: str = MISSING
+    user: str = MISSING
+    system: str = MISSING
     context_type: str = "empty"
     examples_type: Optional[str] = None
     num_few_shots: int = 5
```
Kipok, line 55: Can we put this inside few shot class?

Kipok: And examples_type as well

nemo_skills/inference/prompt/code_sfted.yaml
```diff
@@ -1,7 +1,9 @@
-delimiter: "\n\n\n\n\n\n"
-prefix: "\
-  System:\n\
+few_shot_examples:
+  template: ""
+  delimiter: "\n\n\n\n\n\n"
```
Kipok, line 3: How about we explicitly create a new top-level key called stop_phrases, so that we can fully remove delimiter from here? By the way, I think you don't really need a delimiter in this config since the model will just produce end-of-text token when needed. But it can also help us to customize chat format models by stopping on <extra_id_1> token

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
 
 
+@dataclass
+class Prompt:
+    config: PromptConfig
+    input_dict: Dict[str, Any]
+    example_dicts: List[Dict[str, Any]] = None
```
Kipok, line 62: Optional?

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+    context_template: Optional[str] = None
+    generated_solution = ""
```
Kipok, line 64:
```suggestion
    generated_solution: str = ""
```

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+
+    def __post_init__(self):
+        """Initialize example_dicts if not provided."""
```
Kipok, line 67:
```suggestion
        """Initialize example_dicts/context_template if not provided."""
```

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+            if filled_examples
+            else ""
+        )
+        context = self.build_context(self.input_dict)
+        user = self.config.user.format(examples=examples, context=context, **self.input_dict)
+        return user
+
+    def build_structured_examples(self) -> List[Dict[str, str]]:
+        """Builds a structured representation of the prompt."""
+        structured_prompt = [{"role": "system", "content": self.config.system}] if self.config.system else []
```
Kipok, line 97: Is this how it's supposed to be used? Empty list instead of empty content when system is missing? If empty content works, it will be simpler and more consistent, like this
```suggestion
        structured_prompt = [{"role": "system", "content": self.config.system}]
```

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+        examples = (
+            self.config.few_shot_examples.delimiter.join(filled_examples) + self.config.few_shot_examples.delimiter
+            if filled_examples
+            else ""
+        )
+        context = self.build_context(self.input_dict)
+        user = self.config.user.format(examples=examples, context=context, **self.input_dict)
+        return user
+
+    def build_structured_examples(self) -> List[Dict[str, str]]:
```
Kipok, line 95:
```suggestion
    def build_chat_prompt(self) -> List[Dict[str, str]]:
```

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+
 def get_prompt_config(prompt_type: str) -> PromptConfig:
     # reading prompt format from the yaml file, if not running through hydra
-    with open(Path(__file__).parent / f"{prompt_type}.yaml", "rt", encoding="utf-8") as fin:
-        return PromptConfig(**yaml.safe_load(fin))
+    config_path = Path(__file__).parent / f"{prompt_type}.yaml"
+    with open(config_path, "rt", encoding="utf-8") as fin:
+        # Load the YAML file using OmegaConf
+        loaded_config = OmegaConf.load(fin)
+        prompt_config = OmegaConf.structured(PromptConfig(**loaded_config))
```
Kipok, line 126: You might need to call OmegaConf.to_object instead or in addition to in order to force all inner dataclasses to be properly built as well

nemo_skills/inference/prompt/utils.py
```diff
@@ -61,19 +136,7 @@ def get_prompt_config(prompt_type: str) -> PromptConfig:
...
-
-    filled_examples = []
-    for example_dict in examples:
-        filled_examples.append(prompt_config.template.format(context=context.format(**example_dict), **example_dict))
-    filled_examples.append(
-        prompt_config.template.format(context=context.format(**input_dict), **input_dict, generated_solution="")
-    )
-
-    return prompt_config.prefix + prompt_config.delimiter.join(filled_examples)
+def get_prompt(prompt_config: PromptConfig, input_dict: Dict[str, Any]) -> Prompt:
```
Kipok, line 139: Is this used anywhere?

nemo_skills/inference/server/model.py
```diff
@@ -309,13 +313,92 @@ def _single_call(
...
+    def _single_call(
+        self,
+        prompts,
+        tokens_to_generate,
+        temperature,
+        top_p,
+        repetition_penalty,
+        random_seed,
+        stop_phrases: List[str],
+        top_k=0,  # does not supported by OpenAI
```
Kipok, line 345:
```suggestion
        top_k=0,  # is not supported by OpenAI
```

nemo_skills/inference/server/model.py
```diff
@@ -309,13 +313,92 @@ def _single_call(
...
+            top_p=top_p,
+            max_tokens=tokens_to_generate,
+            presence_penalty=repetition_penalty,
+            seed=random_seed,
+            stop=stop_phrases,
+            messages=messages,
+        ).choices[0]
+        content = response.message.content
+
+        # OpenAI remove stop tokens so we need to add them back
```
Kipok, line 387:
```suggestion
        # OpenAI removes stop tokens so we need to add them back
```

nemo_skills/inference/prompt/utils.py
```diff
@@ -37,19 +39,92 @@
...
+
+    def build_structured_examples(self) -> List[Dict[str, str]]:
+        """Builds a structured representation of the prompt."""
+        structured_prompt = [{"role": "system", "content": self.config.system}] if self.config.system else []
+        structured_prompt.append({"role": "user", "content": self.build_examples()})
+        if self.generated_solution:
+            structured_prompt.append({"role": "assistant", "content": self.generated_solution})
+        return structured_prompt
+
+    def __add__(self, other: str):
```
Kipok, line 103: I'm confused by this logic - why is this needed? Why do you need to hold to the generated_solution?

Kipok (changes requested):

nemo_skills/inference/prompt/utils.py
i-vainn: No, providing empty system prompt leads to bad generations like this `'generated_solution': '\\boxed{27}'`

nemo_skills/inference/prompt/utils.py
i-vainn: This is needed because of `new_outputs[idx]['full_prompt'] += output` in `BaseModel`. You are right, I think better to rewrite this explicitly as `new_outputs[idx]['full_prompt'].generated_solution += output`

Kipok (approved):
Thanks @i-vainn, great work! 

I think we might need to do another iteration on the API here, because in the future we might want to support multi-turn dialogs and the current prompt format cannot really do that in a straightforward way. But let's keep the current version for now and revisit this in the future.

PR #16, opened 2024-03-09
Retry on error

i-vainn:
This PR allows to recover from errors by re-generating failed code blocks and do majority voting on their outputs. Using `error_recovery_attempts=10` and `stop_on_code_error=False` we can improve OpenMath-Mistral-7b results from 44.5 to 46.56 on math.

### DIFF
--- a/nemo_skills/inference/server/model.py
+++ b/nemo_skills/inference/server/model.py
@@ -194,23 +194,23 @@ class BaseModel(abc.ABC):
                         if result['error_message']:
                             new_outputs[idx]['error_message'] = result['error_message']
                             if self.stop_on_code_error:
-                                new_outputs[idx]['full_prompt'].generate_solutions += output
+                                new_outputs[idx]['full_prompt'].generated_solutions += output
                                 continue
                             text_only_part = output.split(CODE_SEPARATORS[0])[0]
-                            new_outputs[idx]['full_prompt'].generate_solutions += text_only_part
+                            new_outputs[idx]['full_prompt'].generated_solutions += text_only_part
                             code_output = self.__recover_from_error(request, new_outputs[idx], executor)
                             # if re-generation did not help
                             if code_output is None:
                                 code_output = result["result"]
-                                new_outputs[idx]['full_prompt'].generate_solutions += output[len(text_only_part) :]
+                                new_outputs[idx]['full_prompt'].generated_solutions += output[len(text_only_part) :]
                         else:
-                            new_outputs[idx]['full_prompt'].generate_solutions += output
+                            new_outputs[idx]['full_prompt'].generated_solutions += output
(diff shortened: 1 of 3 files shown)


nemo_skills/inference/server/model.py
```diff
@@ -224,6 +236,53 @@ def __call__(
             )
         return outputs
 
+    def __recover_from_error(self, request, new_output, executor):
```
Kipok, line 239: Do you really need a double underscore? :)

nemo_skills/inference/server/model.py
```diff
@@ -224,6 +236,53 @@ def __call__(
...
+        recovery_request = {key: value for key, value in request.items() if key != 'prompts'}
+        recovery_request['prompts'] = [new_output['full_prompt']]
+
+        # if greedy, setting parameters to random
+        if request['temperature'] == 0 or request['top_k'] == 1:
+            recovery_request['temperature'] = 0.7
```
Kipok, line 245: Can we take this as a new data class called error_recovery and then you can add attempts as well as the temperature and other parameters there. + an option to enable/disable majority voting, so that it all can be customized?

nemo_skills/inference/server/model.py
```diff
@@ -78,6 +79,7 @@ def __init__(
         max_code_output_characters=1000,
         code_execution_timeout=10.0,
         max_code_executions=3,
+        error_recovery_attempts=0,
```
Kipok, line 82: Can you please add the documentation?

Kipok (changes requested):
Thanks @i-vainn looks great! Just a few minor things

nemo_skills/inference/server/model.py
Kipok: Also can you please add some logic to NeMo model that will error out when this new parameter is used (as well as stop_on execution error and other things that are only supported in external loop mode)? So that people don't try to use them and be surprised that they don't work, until we add the logic to handle it.

nemo_skills/inference/server/model.py
```diff
@@ -365,11 +367,19 @@ def __init__(
...
+            unsupported_arguments = [
+                'max_code_output_characters',
+                'code_execution_timeout',
+                'max_code_executions',
+                'error_recovery',
+                'stop_on_code_error',
+            ]
+            for arg in unsupported_arguments:
+                if arg in kwargs:
+                    raise ValueError(f"`{arg}` is not supported by NemoModel.")
```
Kipok, line 382:
```suggestion
                    raise ValueError(f"`{arg}` is not supported by NemoModel if handle_code_execution=False. To use it, set handle_code_execution=True.")
```

Kipok (approved):
Thanks, looks good! Only need to make sure docs are visible in cmdline help. Feel free to push here or do it in another PR

PR #22, opened 2024-04-25
gnalbandyan/special tokens

gnalbandyan:
- Add a flag process_output in nemo_skills/inference/server/model.py . By passing it False, we can gather model raw predictions and apply postprocessing on our side. By default, process_output=True. No change should be done in reasoning pipeline.
- Add handling model.module weight name structure in NeMo->HF conversion. No change should be done in reasoning pipeline.

### DIFF
--- a/nemo_skills/conversion/nemo_to_hf.py
+++ b/nemo_skills/conversion/nemo_to_hf.py
@@ -128,11 +128,12 @@ def convert(input_nemo_path, output_hf_path, hf_model_name, max_shard_size, prec
     heads_per_group = head_num // num_query_groups
     qkv_total_dim = head_num + 2 * num_query_groups
 
-    for weight_base_name in ['model', 'model.module']:
-        if f'{weight_base_name}.embedding.word_embeddings.weight' in model.state_dict():
-            break
+    if 'model.embedding.word_embeddings.weight' in model.state_dict():
+        weight_base_name = 'model'
+    elif 'model.module.embedding.word_embeddings.weight' in model.state_dict():
+        weight_base_name = 'model.module'
     # Embedding
-    embed_weight = model.state_dict()[f'{weight_base_name}.embed_tokens.weight']
+    embed_weight = model.state_dict()[f'{weight_base_name}.embedding.word_embeddings.weight']
     embed_weights_base_name = f'model.embed_tokens.weight'
     checkpoint[embed_weights_base_name] = param_to_weights(embed_weight)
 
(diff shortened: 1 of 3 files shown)


nemo_skills/conversion/nemo_to_hf.py
```diff
@@ -128,15 +128,18 @@ def convert(input_nemo_path, output_hf_path, hf_model_name, max_shard_size, prec
     heads_per_group = head_num // num_query_groups
     qkv_total_dim = head_num + 2 * num_query_groups
 
+    weight_base_name = 'model'
+    if f'{weight_base_name}.embedding.word_embeddings.weight' not in model.state_dict():
+        weight_base_name = 'model.module'
```
Kipok, line 133: For which models is this necessary? I would ideally avoid changing this script much since this is mostly a copy-paste from nemo and if we add new things here, it would be harder to merge upstream changes later

nemo_skills/inference/server/model.py
```diff
@@ -165,14 +167,18 @@ def __call__(
...
             outputs = self._single_call(**request)
-            outputs = [
-                {
-                    'generated_solution': remove_stop_tokens(output, stop_phrases),
-                    'predicted_answer': extract_answer(output),
-                    'error_message': extract_error_message(output),
-                }
-                for output in outputs
-            ]
+            # if process_output is False, return the raw outputs
```
Kipok, line 170: Can you explain a bit more what exactly you need from the output? While we can certainly add this flag, I'm trying to understand better what's the use-case is and whether there is another way to get to what you want. The problem is that those fields are also used in other scripts, so basically disabling process_output is guaranteed to break anything else you do with our pipeline and while it might be fine for your case, since it seems that you just want to run inference and then not use anything else, it might be confusing for other cases, where someone might enable this flag and still expect the other scripts to work. Plus, there are other places where the inference can happen and so we should make sure this flag is available there as well.

Happy to sync offline as well if that's easier.

nemo_skills/inference/server/model.py
```diff
@@ -165,6 +165,9 @@ def __call__(
         if not self.handle_code_execution:
             request["prompts"] = prompts
             outputs = self._single_call(**request)
+            # if process_output is False, return the raw outputs
```
Kipok, line 168: remove the comment?

Kipok:
Thanks!

Kipok (approved):

PR #23, opened 2024-05-02
Add piston option for code execution + refactoring of local sandbox to share the code

Kipok:
Also added tests workflow and code exec tests


Kipok:
Let me merge this to unblock some other coding work I'm doing - please still review when you get a chance and post any issues (will fix them in a follow up PRs)

PR #28, opened 2024-05-29
PR to add initial vLLM support

titu1994:
Initial pr to add vLLM inference support.

# Warning
There is a difference in metrics computed by vLLM vs metrics computed using NeMo or TRTLLM - despite best efforts in maintaining exact same arguments during greedy inference.

This can lead to potentially worse results on benchmarks depending on which inference server was used.

# Changelog
- Adds vllm Dockerfile
- Add vllm inference engine
- Update some pipelines to support inference via pipeline


nemo_skills/inference/server/model.py
```diff
@@ -493,10 +493,200 @@ def _send_request(
...
+        )
+
+        self.server_type = "openai"
+        self.oai_client = None
+        self.oai_client = self.prepare_openai(self.server_host, self.server_port)  # type: openai.OpenAI
+
+        self.model_name_server = self.get_model_name_from_server()
+        self.model = self.model_name_server
+
+        print(f"Model hosted by {self.server_type} server:", self.model)
```
Kipok, line 533:
```suggestion
        LOG.info("Model hosted by %s server: %s", self.server_type, self.model)
```

nemo_skills/inference/server/model.py
```diff
@@ -493,10 +493,200 @@ def _send_request(
...
+            request["top_k"] = 1
+            request["top_p"] = 1.0
+
+        outputs = self.prompt_api(
+            prompt=request['prompt'],
+            max_tokens=request['max_tokens'],
+            temperature=request['temperature'],
+            top_p=request['top_p'],
+            top_k=request['top_k'],
+            num_generations=request.get('num_generations', 1),
```
Kipok, line 582: I guess defaults are not necessary since you explicitly specify everything above? So you can also make it simpler by just using something like

```
outputs = self.prompt_api(**request, parse_response=True)
```

nemo_skills/inference/server/model.py
```diff
@@ -493,10 +493,200 @@ def _send_request(
...
+        logit_bias: dict = None,
+        seed: int = None,
+        parse_response: bool = True,
+    ) -> Union["openai.types.Completion", List[str]]:
+
+        if isinstance(prompt, (list, tuple)):
+            prompt = [str(prmt) for prmt in prompt]
+
+        if temperature == 0.0:
+            temperature = 1.0
```
Kipok, line 619: This should be done alongside setting top_k to 1 always? Otherwise you risk accidentally setting temp=1 when it was not intended to be so, if for some reason you get a request with t=0, top_k=32, e.g.

nemo_skills/inference/server/serve_vllm.sh
```diff
@@ -0,0 +1,82 @@
...
+
+python -m vllm.entrypoints.openai.api_server \
+  --model "${MODEL_PATH}" \
+  --trust-remote-code \
+  --seed=${SEED} \
+  --host="0.0.0.0" \
+  --port=${SERVER_PORT} \
+  --served-model-name "${MODEL_NAME}" \
+  --tensor-parallel-size=${NUM_GPUS} \
+  --max-num-seqs=1024 \
```
Kipok, line 80: What's this parameter controlling? Why hardcoded value?

pipeline/launcher.py
```diff
@@ -54,6 +54,15 @@ def get_server_command(server_type, num_gpus):
         # somehow on slurm nemo needs multiple tasks, but locally only 1
         if CLUSTER_CONFIG["cluster"] == "local":
             num_tasks = 1
+
+    elif server_type == 'vllm':
+        server_start_cmd = (
+            f"(NUM_GPUS={num_gpus} bash /code/nemo_skills/inference/server/serve_vllm.sh /model/ {model_name} "
```
Kipok, line 60: Can you please remind me why model_name needs to be explicit here? I remember you were telling me before, but cannot recall what's the reason now

pipeline/run_eval.py
```diff
@@ -123,14 +123,20 @@ def get_sampling_cmd(benchmark, random_seed, extra_eval_args="", extra_arguments
...
     extra_arguments = f'{" ".join(unknown)}'
 
     args.model_path = Path(args.model_path).absolute()
     args.output_dir = Path(args.output_dir).absolute()
 
-    server_start_cmd, num_tasks = get_server_command(args.server_type, args.num_gpus)
+    server_start_cmd, num_tasks = get_server_command(args.server_type, args.num_gpus, args.model_path.name)
+
+    if args.server_type == "vllm":
+        server_wait_string = "Uvicorn running"
```
Kipok, line 137: Can you return this from the ger_server_command, so that we keep it in one place?

Kipok (changes requested):
Thanks Som, looks great! Some minor comments after which I will test it on my side and if all looks good (subject to expected accuracy degradation), will merge

Kipok (approved):

Kipok:
Have not fully tested yet, but let me merge and I will test in another branch and do fixes there if necessary (as I'm going to create a large PR with many changes soon anyway)

nemo_skills/inference/server/serve_vllm.sh
titu1994: The largest executed batch size. The model will still store but not execute more than this at any given time. For 1000 80gb, this seems a reasonable limit

pipeline/launcher.py
titu1994: It doesn't need to be if we don't put the logic for prompt handling anymore. It's just good for the user logs to show which model is loaded at this point so that they don't mistake which model is loaded and waste compute

PR #29, opened 2024-05-29
LLM reasoning with tables (using TabMWP)

redoctopus:
First draft merging new LLM code with edits to get SFT, eval, etc. of tabular data working.

Most notably, added few shot examples and tables context options, plus a few minor changes to get eval/synthetic data generation/SFT to run.


datasets/tabmwp/prepare.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
```
Kipok, line 12: Please add more calls to this script inside dataseta/prepare_all.sh as I assume there are some subsets other than test that need to be loaded?

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
```diff
@@ -0,0 +1,217 @@
...
+$275 | 20,500 | 9,400
+$395 | 18,400 | 13,000
+$515 | 16,300 | 16,600
+$635 | 14,200 | 20,200
+""".strip(),
+        "reference_solution": "At the price of $155, the quantity demanded is greater than the quantity supplied. There is not enough of the good or service for sale at that price. So, there is a shortage.\n",
+        "generated_solution": """
+Let's solve this problem using Python code and the values from the table.
+{start_code}
+quantity_demanded_price_155 = 22600
```
Kipok, line 61: Note that you don't have to use code if the problem does not require it. In this case it looks like a more natural solution would be to just solve it in text

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
```diff
@@ -0,0 +1,217 @@
...
+{start_code}
+cans = [7, 4, 5, 8, 9]
+cans = sorted(cans)
+middle1 = (len(cans) - 1) // 2
+middle2 = len(cans) // 2
+result = (cans[middle1] + cans[middle2]) / 2
+result
+{end_code}
+{start_code_output}
+7
```
Kipok, line 98: I think the actual output that Python gives is "7.0". Suggest you double check it in all places to make sure the few-shot examples are more representative of what the model will see during real inference

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
```diff
@@ -0,0 +1,217 @@
...
+{start_code_output}
+yes
+{end_code_output}
+\\boxed{Yes}, Lorenzo does have enough money to buy a toy guitar and juggling balls.
+""".strip(),
+    },
+]
+
+
+only_code = [
```
Kipok, line 137: I think this will not work in the current code version - have you used this few-shots? If not, maybe remove them

pipeline/prepare_eval.py
```diff
@@ -64,6 +64,7 @@
     args.output_path.mkdir(exist_ok=True, parents=True)
 
     conversion_step = f"&& mv /training_folder/model-averaged.nemo /inference_folder/{model_name}"
+    conversion_step += f" && cp /nemo_model/model_config.yaml /inference_folder/"
```
Kipok, line 67: just wondering why is this needed?

pipeline/run_eval.py
```diff
@@ -98,6 +99,7 @@ def get_sampling_cmd(benchmark, random_seed, extra_eval_args="", extra_arguments
     wrapper_args.add_argument("--model_path", required=True)
     wrapper_args.add_argument("--server_type", choices=('nemo', 'tensorrt_llm'), default='tensorrt_llm')
     wrapper_args.add_argument("--output_dir", required=True)
+    wrapper_args.add_argument("--context_type", default="empty")
```
Kipok, line 102: You don't need to add an extra parameter here as you can always add `++prompt.context_type=table` to the end of the command and it will work. So let's remove it to not have redundant options

Kipok: (and put empty as the default back in the cmd above)

pipeline/run_eval.py
```diff
@@ -164,7 +167,11 @@ def get_sampling_cmd(benchmark, random_seed, extra_eval_args="", extra_arguments
     eval_cmds = [" ".join(eval_cmds[i :: args.num_nodes]) for i in range(args.num_nodes)]
 
     for idx, eval_cmd in enumerate(eval_cmds):
-        extra_sbatch_args = ["--parsable", f"--output={args.output_dir}/slurm_logs_eval{idx}.log"]
+        extra_sbatch_args = [
+            "--parsable",
+            f"--output={args.output_dir}/slurm_logs_eval{idx}.log",
+            f"--nv-meta ml-model.llm-reasoning ",
```
Kipok, line 173: This should go as part of the extra_sbatch_args to cluster config and not in the code

pipeline/run_labeling.py
```diff
@@ -62,7 +62,7 @@ def run_script(format_dict, seed, extra_arguments, partition=None, dependency=No
     format_dict["random_seed"] = seed
     format_dict["extra_arguments"] = extra_arguments
 
-    extra_sbatch_args = ["--parsable", f"--output={LOGS.format(**format_dict)}"]
+    extra_sbatch_args = ["--parsable", f"--output={LOGS.format(**format_dict)}", f"--nv-meta ml-model.llm-reasoning "]
```
Kipok, line 65: This should go as part of the extra_sbatch_args to cluster config and not in the code

pipeline/run_sft_and_eval.py
```diff
@@ -27,11 +27,10 @@
 
 def start_job(extra_sbatch_args: str, cmd: str) -> int:
     full_cmd = f'{extra_sbatch_args} {cmd}'
+
     if CLUSTER_CONFIG['cluster'] == 'local':
         subprocess.run(full_cmd, shell=True)
         return 0
-    job_id = subprocess.run(full_cmd, shell=True, check=True, capture_output=True).stdout.decode()
```
Kipok, line 33: Accidentally removed?

Kipok (changes requested):

pipeline/run_sft_and_eval.py
redoctopus: Yep, thanks for catching this!

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
redoctopus: Oh, nice catch, thanks. Just ran all the example solutions to double-check the outputs and found a few more issues.

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
redoctopus: Do we have a way of mixing code/text solutions with the model being able to choose what to use?

In this case the end goal is to have the values loaded programmatically so I figure these examples should all involve translating the problem to code to begin with.

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
redoctopus: Odd, it seemed to be working for me a few weeks back--I used it to make prompts for synthetic data generation. When I checked there was definitely a difference between the results from `only_code` vs. `text_with_code`.

What is expected behavior for the current version?

pipeline/run_eval.py
redoctopus: Ah, I was originally worried that having it specified twice in the command would cause an error.

Just to make sure--will the `++prompt.context_type=table` additional arg overwrite the existing `prompt.context_type=empty` if both are specified?

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
Kipok: The model is supposed to learn that. So if you generate some examples that use code and some that don't and train on those, the hope is that it will be able to differentiate between these two cases at inference time. But you need to be showing this in few shots (either in the same set of few shots show both or have example with/without code and run multiple labeling jobs)

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
Kipok: the code you generate is not going to be run unless you wrap it with the special tokens. Since there are no such tokens here, the model will still learn to produce the code, but will never see its output

pipeline/run_eval.py
Kipok: yes, it will overwrite

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
redoctopus: Gotcha, makes sense. I'll keep these with code examples for now since we're focusing on code-generation, but good to keep in mind.

pipeline/prepare_eval.py
redoctopus: Some part of the eval process where the model was being restored wanted this file in the same directory as the checkpoint, but I can't remember exactly where it was that was erroring out without this...

nemo_skills/inference/prompt/few_shot_examples/examples_tabmwp.py
redoctopus: Yeah, that checks out. I don't think we use this format so I'll remove these examples.

pipeline/prepare_eval.py
Kipok: It shouldn't be needed, so let's remove it for now. If you get an error again, please let me know

pipeline/run_labeling.py
redoctopus: Done

pipeline/run_eval.py
redoctopus: Done

Kipok (approved):
looks good, thanks!

Kipok:
Actually looks like one of the tests is failing because something is wrong with dataset creation. Can you please fix that? Also add other splits into the test file so that we check that they are created as well

redoctopus:
> Actually looks like one of the tests is failing because something is wrong with dataset creation. Can you please fix that? Also add other splits into the test file so that we check that they are created as well

Ah whoops, looks like I was saving them as .json instead of the expected .jsonl. Fixed.

PR #30, opened 2024-05-31
Fixes of data explorer bugs

dgtm777:
Fix bugs:
- Allow save dataset multiple times
- Fix samples filtering
- Allow the same files naming
- Remove regular expression which may cause exponential backtracking
- Fix jupyter-notebook's output visualization
- Fix changing models in comparison mode

Add tests
Add base_generation variable for filtering
Update readme and filtering instructions


visualization/Readme.md
```diff
@@ -32,7 +32,7 @@ This page enables the analysis of model answers based on different parameters. I
...
 
 ## Analyze page
-To use the Analyze page, specify paths to the generations you want to use (if not obtained through the "Inference" page). You can pass parameters via the command line with `++visualization_params.model_prediction.generation1='/some_path/generation1/output-greedy.jsonl'` or add them in an additional config file.
+To use the Analyze page, specify paths to the datasets you want to use (if not obtained through the "Inference" page). You can pass parameters via the command line with `++visualization_params.model_prediction.dataset1='/some_path/generation1/output-greedy.jsonl'` or add them in an additional config file.
```
Kipok, line 35: I'm a bit confused by usage of "dataset" vs "generation" in this doc. Can you just pick one and consistently use it everywhere?

.github/workflows/tests.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         python -m pip install --upgrade pip
         pip install .
         pip install pytest pytest-cov
+        pip install -r visualization/requirements.txt
```
Kipok, line 28: I think it's best if you move this to a separate testing pipeline where you'd only run visualization tests. Your current tests fail most likely because you need to use some other plugin / base image to support selenium (I guess the current one does not have necessary requirements for browser-based tests). So it could be necessary to move these away anyway, but another reason to do that is that visualization is optional and we don't want to accidentally miss some import errors by installing these requirements for main tests.

visualization/utils/strategies/whole_dataset_mode.py
```diff
@@ -52,10 +54,22 @@ def get_query_input_layout(self, dataset) -> List[dbc.AccordionItem]:
...
 
     def run(self, utils: Dict, params: Dict) -> html.Div:
-        utils = {key.split(SEPARATOR_ID)[-1]: value for key, value in utils.items()}
+        utils = {
+            key.split(SEPARATOR_ID)[-1]: value
+            for key, value in utils.items()
+            if key != RETRIEVAL and key not in RETRIEVAL_FIELDS
+        }
+        logging.info(f"Whole dataset mode utils: {utils}")
+        exmaples_type = utils.pop("examples_type", None)
```
Kipok, line 63:
```suggestion
        examples_type = utils.pop("examples_type", None)
```

visualization/utils/strategies/whole_dataset_mode.py
```diff
@@ -52,10 +54,22 @@ def get_query_input_layout(self, dataset) -> List[dbc.AccordionItem]:
...
+        example_dicts = get_examples().get(
+            exmaples_type,
```
Kipok, line 65:
```suggestion
            examples_type,
```

Kipok (approved):
Thanks! Great work!

PR #31, opened 2024-05-31
New datasets

dgtm777:
Add datasets


tests/test_is_output_correct.py
```diff
@@ -24,9 +24,15 @@
...
         ("\\frac{1}{2}", 0.5),
+        ("\\frac {1}{2}", 0.5),
         ("x^2+2x+1", "x^2 + 2*x + 1"),
         ("x^2+2x+1", "x^2 + 2*x - (-1)"),
         ("x^2+2x+1", "2x+ 1+x^2"),
+        ("17\\text{ any text}", "17"),
+        ("\$10", "10"),
+        ("10%", "1/10"),
+        ("56,\\!01,\\!78,\\!95,\\!760", "56017895760"),
+        ("any text=12", "12"),
```
Kipok, line 35: I'm not sure we necessarily want to evaluate this as correct. The model should instead put boxed around a number only

datasets/prepare_all.sh
```diff
@@ -6,18 +6,26 @@ python3 ${SCRIPT_DIR}/gsm8k/prepare.py --split_name validation
 python3 ${SCRIPT_DIR}/gsm8k/prepare.py --split_name test
 # train + validation - should be used for the final run only
 python3 ${SCRIPT_DIR}/gsm8k/prepare.py --split_name train_full
+echo "Preparing gsm-plus"
```
Kipok, line 9: Please include all of the new datasets in here https://github.com/Kipok/NeMo-Skills/blob/main/tests/test_datasets.py#L63

tests/test_is_output_correct.py
```diff
@@ -24,9 +24,15 @@
         (5, 5.0),
         ("1/2", 0.5),
         ("\\frac{1}{2}", 0.5),
+        ("\\frac {1}{2}", 0.5),
         ("x^2+2x+1", "x^2 + 2*x + 1"),
         ("x^2+2x+1", "x^2 + 2*x - (-1)"),
         ("x^2+2x+1", "2x+ 1+x^2"),
+        ("17\\text{ any text}", "17"),
+        ("\$10", "10"),
+        ("10%", "1/10"),
```
Kipok, line 33: I think this is a bit hard to support in our current implementation. So let's maybe ignore errors like that. It will work if you replace 1/10 with 0.1

tests/test_is_output_correct.py
```diff
@@ -24,9 +24,15 @@
         (5, 5.0),
         ("1/2", 0.5),
         ("\\frac{1}{2}", 0.5),
+        ("\\frac {1}{2}", 0.5),
```
Kipok, line 27: Please remove all of these test cases - I've ported them to another branch where they will be supported (except the two for which I left comments).

Kipok (changes requested):

Kipok (approved):

PR #33, opened 2024-06-06
Shtoshni/standard prompts

shtoshni:
Hey @Kipok,

Here are the standard examples used for few-shot prompting. There are based off lm-eval-harness which derives these examples from the Jason Wei paper for GSM8K and Minerva paper for MATH.


Kipok (approved):
Thanks!

PR #35, opened 2024-06-09
Fix data explorer tool after large changes in pipeline

dgtm777:


visualization/utils/common.py
```diff
@@ -451,3 +462,27 @@ def run_subprocess(command: str) -> Tuple[str, bool]:
         success = False
 
     return result.stdout.strip(), result.stderr.strip(), success
+
+
+def get_prompt_types():
```
Kipok, line 467: Can you reuse this from https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/inference/prompt/utils.py#L34?

Kipok (approved):

PR #37, opened 2024-06-12
SDP data preparation pipeline

i-vainn:
Changed `nemo-skills/finetuning/prepare_sft_data.py` to [SDP](https://github.com/NVIDIA/NeMo-speech-data-processor) pipeline


nemo_skills/finetuning/data_preparation_utils/sdp_configs/base_config.yaml
```diff
@@ -0,0 +1,51 @@
+processors_to_run: all
```
Kipok, line 1: It will be more user-friendly if most important parameters for different processors are available as top-level arguments. E.g. create a new argument `prediction_jsonl_files` and then in the first processor use it directly via `${prediction_jsonl_files}`. Same for output location or important filtering steps, like downsampling_method, etc. Basically, most of the parameters that we currently have in the original prepare script can be used here for convenience. You can also use boolean conditions to skip certain processors based on top-level arguments like here https://github.com/NVIDIA/NeMo-speech-data-processor/blob/main/dataset_configs/italian/mls/config.yaml#L102 or here https://github.com/NVIDIA/NeMo-speech-data-processor/blob/main/dataset_configs/italian/mls/config.yaml#L196

Kipok: If you find it convenient to skip processors via top-level args, you can potentially try to combine code/non-code flow into a single config and just pick different subparts based on the args. But creating multiple configs might be easier

nemo_skills/finetuning/prepare_sdp.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+import hydra
+from data_preparation_utils.run_processors import run_processors
+
+
+@hydra.main(version_base=None)
```
Kipok, line 21: Let's set the config folder location to be default here, so that you can just use --config-name=... and it works?

nemo_skills/finetuning/prepare_sdp.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: Let's also remove the old file to not keep two versions of the same logic

nemo_skills/finetuning/data_preparation_utils/preprocessing.py
```diff
@@ -0,0 +1,222 @@
...
+        else:
+            raise NotImplementedError(f"Sampling method {self.sampling_method} not implemented")
+
+        with open(self.output_manifest_file, "wt", encoding="utf-8") as fout:
+            for instance in output_instances:
+                fout.write(json.dumps(instance) + "\n")
+
+
+class WriteFinalSftManifest(BaseProcessor):
+    def __init__(self, prompt_type: str, chat_format: bool = False, metadata: Dict = {}, **kwargs):
```
Kipok, line 194: Please don't use mutable defaults like `{}`, instead set to None and check inside if it's None, then create an empty dict

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,283 @@
...
+from nemo_skills.code_execution import CODE_OUTPUT_SEPARATORS, CODE_SEPARATORS
+from nemo_skills.synthetic_arithmetic.solve_expression import solve_expression
+
+PATTERN_ANS = re.compile(r"\\boxed\{([^}]*)\}")
+PATTERN_CODE = re.compile(CODE_SEPARATORS[0])
+
+
+class DropMultiBoxed(BaseParallelProcessor):
+
+    def __init__(self, text_key: str = "generation", **kwargs):
```
Kipok, line 37: Maybe call it "solution_key" instead?

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,283 @@
...
+        ans_match = PATTERN_ANS.search(data_entry[self.text_key])
+        code_match = PATTERN_CODE.search(data_entry[self.text_key])
+        if not ans_match or not code_match or ans_match.start() > code_match.start():
+            return [DataEntry(data=None)]
+
+        return [DataEntry(data=data_entry)]
+
+
+class DropBrokenCode(BaseParallelProcessor):
+    PATTERN_ANS = re.compile(r"\\boxed\{([^}]*)\}")
```
Kipok, line 63: duplicate definition?

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,283 @@
...
+
+        trimmed_output = "\n".join(output_lines[: stop_idx + 1])
+        data_entry[self.text_key] = trimmed_output
+
+        return [DataEntry(data=data_entry)]
+
+
+class SplitArithmetic(BaseParallelProcessor):
+
+    def __init__(self, remove_incorrect: bool = False, text_key: str = "generation", **kwargs):
```
Kipok, line 124: why False as default? Seems like we usually want to do this?

Kipok (changes requested):

nemo_skills/finetuning/data_preparation_utils/preprocessing.py
```diff
@@ -0,0 +1,236 @@
...
+        random.seed(self.random_seed)
+        if self.sampling_method == "random":
+            output_instances = list(chain(*groupped_samples))
+            if self.do_shuffle:
+                random.shuffle(output_instances)
+            output_instances = output_instances[: self.num_samples]
+        elif self.sampling_method == "fair":
+            soln_counter = 0
+            output_instances = []
+            while self.num_samples is not None:
```
Kipok, line 186: why is this a "while"? Also this does not work if num_samples > data size, it hangs forever

Kipok (changes requested):

Kipok (commented):
Somehow I'm getting more entries when running the new script. Can it be that we don't deduplicate properly?

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,306 @@
...
+
+from nemo_skills.code_execution import CODE_OUTPUT_SEPARATORS, CODE_SEPARATORS
+from nemo_skills.synthetic_arithmetic.solve_expression import solve_expression
+from nemo_skills.synthetic_arithmetic.utils import extract_expressions
+
+PATTERN_ANS = re.compile(r"\\boxed\{([^}]*)\}")
+PATTERN_CODE = re.compile(CODE_SEPARATORS[0])
+
+
+class BaseFilter(BaseParallelProcessor):
```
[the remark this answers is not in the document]
i-vainn: This don't work for some reason. Drop... are freezing for some reason when in_memory_chunksize is 10M.

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,305 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def process_dataset_entry(self, data_entry) -> List:
+        raise NotImplementedError
+
+    def test(self):
```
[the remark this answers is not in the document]
i-vainn: Yeah, I agree. This used to make sense in old code with separate configs for each task. But now since we have single config and modify it only through cli, this seems unnecessary.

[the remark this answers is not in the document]
Kipok: But why do you need an extra logic for that? SDP should already not test any processors that are disabled with should_run argument

nemo_skills/finetuning/data_preparation_utils/filters.py
```diff
@@ -0,0 +1,306 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
```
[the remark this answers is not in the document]
i-vainn: I asked SDP devs, this one will be hopefully added to SDP soon.

Kipok (approved):

Kipok:
Looks great, let's merge this!

PR #38, opened 2024-06-12
Filtering out duplicate solutions

shtoshni:
This patch filters out duplicate solutions for a given problem.

### DIFF
--- a/nemo_skills/finetuning/filtering_utils.py
+++ b/nemo_skills/finetuning/filtering_utils.py
@@ -30,19 +30,19 @@ def process_bad_solutions(
     code_solns = []
     text_solns = []
     # Solution set to ensure uniqueness of solutions filtered through
-    solution_set = set() 
+    solution_set = set()
     for sample in samples:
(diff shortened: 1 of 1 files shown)


nemo_skills/finetuning/filtering_utils.py
```diff
@@ -29,15 +29,25 @@ def process_bad_solutions(
...
     text_solns = []
+    # Solution set to ensure uniqueness of solutions filtered through
+    solution_set = set()
     for sample in samples:
         if should_remove(sample['generation'], solution_filters):
             continue
         if should_trim:
             sample['generation'] = trim_output(sample['generation'])
+
+        if sample['generation'] in solution_set:
```
Kipok, line 40: So we actually do have deduplication initially, but it looks like you're getting some more duplicates after trimming. @i-vainn let's make sure that in your refactoring this deduplication is an explicit processor that we call last right before producing the final file to avoid this in the future.

Kipok (approved):

PR #41, opened 2024-06-14
Math fixes

shtoshni:
This commit handles the multiple boxed entries that are present in the reference solution of the MATH dataset.


Kipok (approved):
thanks!

PR #42, opened 2024-06-16
Provide an way to launch a HF model on VLLM directly by HF name

titu1994:
# Changelog

- Provide a way to simply use "HF" name of a model for VLLM codepath, so that local development can conveniently switch models without having to apriori download the model directory explicitly

### DIFF
--- a/pipeline/start_server.py
+++ b/pipeline/start_server.py
@@ -19,6 +19,7 @@ import sys
 import time
 from argparse import ArgumentParser
 from pathlib import Path
+from huggingface_hub import get_token
 
 # adding nemo_skills to python path to avoid requiring installation
 sys.path.append(str(Path(__file__).absolute().parents[1]))
@@ -45,7 +46,7 @@ else \
     {server_start_cmd}; \
 fi \
 """
-MOUNTS = "{NEMO_SKILLS_CODE}:/code,{model_path}:/model"
+MOUNTS = "{NEMO_SKILLS_CODE}:/code"
 JOB_NAME = "interactive-server-{server_type}-{model_name}"
 
 # TODO: nemo does not exit on ctrl+c, need to fix that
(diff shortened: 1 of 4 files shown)


pipeline/start_server.py
```diff
@@ -20,6 +20,8 @@
 from argparse import ArgumentParser
 from pathlib import Path
 
+from huggingface_hub import get_token
```
Kipok, line 23: Can you please add a try/except here and define get_token to just return os.getenv("HF_TOKEN", "") in case it cannot be imported (with some warning if you'd like)? We try to keep the pipeline scripts running without any extra packages installed as they are run on slurm clusters and it's best not to install extra things there

pipeline/start_server.py
```diff
@@ -73,26 +75,36 @@
     )
     args = parser.parse_args()
 
-    args.model_path = Path(args.model_path).absolute()
+    if os.path.exists(args.model_path):
```
Kipok, line 78: Maybe better to make model_path optional argument and add another argument model_name for your usecase? And if it's provided, we still do all the checks below. Otherwise if we provide incorrect model path, there will be no error, until the model is actually loaded inside slurm job

Kipok (changes requested):
Thanks, just a couple of small changes are needed

pipeline/start_server.py
titu1994: Good point will do.

pipeline/start_server.py
titu1994: Hmm that's an option but it opens up too many combinations and we have to sit and reject half the combinations that don't work. I can make it more robust by saying model path split is length two and that the file does not exist, which should be sufficient to assert that it's a HF path

Kipok: why is that sufficient? You can currently specify relative model path as `trt_models/some_model_name`, which is what I do often. Feels safer to have an extra parameter and then just a check if both are specified or none are specified, we raise an error. Especially given that in some of the servers (e.g. trtllm), the error message when it can't find the model is super obscure, I think it just fails with segfault and you'd spend some time trying to figure out what happened while you simply had a path typo

titu1994: Oof ok makes sense

titu1994: Made the changes

pipeline/start_server.py
```diff
@@ -20,6 +20,11 @@
 from argparse import ArgumentParser
 from pathlib import Path
 
+try:
+    from huggingface_hub import get_token
+except (ImportError, ModuleNotFoundError):
+    get_token = lambda: os.environ.get('HF_TOKEN', None)
```
Kipok, line 26:
```suggestion
    get_token = lambda: os.environ.get('HF_TOKEN', '')
```

Kipok (commented):
Just a small fix, but otherwise looks good!

Kipok (approved):

PR #50, opened 2024-06-27
Revert commit af9bd3b: Edge case for trt llm out-of-vocabulary

yinglin-nv:
This PR reverts changes introduced in this commit: https://github.com/Kipok/NeMo-Skills/commit/af9bd3b9d5590918350a35cb9b8b549f14486040, which adds lines that handle out-of-range IDs which are not needed now as the TensorRT-LLM bug has been fixed. Also, `tokenizer.vocab_size` doesn't take special tokens into consideration and thus may impact the generation of special tokens.


Kipok (approved):
Thanks!

PR #51, opened 2024-06-28
Add code evaluations and mmlu

Kipok:
Will add mbpp and refactor it a bit more tomorrow


Kipok:
All should be ready and working now. The mmlu version here is generative, so we don't check the log probs, thus the results will be different from "standard" eval. As a follow up PR later on I will add support for arbitrary metrics, since it does not make sense to use current wrong/correct/no answer that we use for math there. Also, will need to update docs, they are a bit out-of-date now.

Here is an example of how to run eval on multiple benchmarks that require different prompts (change nemo_skills/inference/prompt/llama3/eval_map.py or add it to another folder to customize)

```
python pipeline/run_eval.py \
    --model_path /mnt/datadrive/models/Meta-Llama-3-8B-Instruct \
    --server_type vllm \
    --output_dir test-evals \
    --benchmarks human-eval:0 mbpp:0 gsm8k:0 mmlu:0 \
    --num_gpus 2 \
    --num_nodes 1 \
    --num_jobs 1 \
    --prompt_folder llama3 \
    --model_version instruct \
    ++split_name=test \
    ++batch_size=256 \
    ++inference.tokens_to_generate=2048
```

PR #52, opened 2024-06-28
Rounding script

dgtm777:
Script to add rounding instructions to questions


Kipok (changes requested):
Sorry, can you please rebase on top of latest main? The prepare_all.sh is now prepare.py, so some slight changes are needed

datasets/add_rounding_instructions.py
```diff
@@ -0,0 +1,27 @@
+import json
```
Kipok, line 1: Would it be easier and more straightforward to define it as a function inside datasets/utils.py and then import it, so that you don't need to call submodule commands and also don't need to read from the disk and can modify data in-memory?

datasets/gsm-plus/prepare.py
```diff
@@ -48,16 +53,23 @@
             "reversing_operation",
         ],
     )
+    parser.add_argument("--add_rounding_instructions", type=bool, default=True)
```
Kipok, line 56: I think if you specify it in this way, it will be impossible to disable? E.g. how do you run this script without that argument set to true

Kipok (approved):

PR #58, opened 2024-07-10
Fixed OpenAI API

i-vainn:


nemo_skills/inference/prompt/openai/chat.yaml
```diff
@@ -0,0 +1,24 @@
+few_shot_examples:
+  template: "Example question:\n{question}\n\n{context}Example solution:\n{generation}\n\n\n\n\n\n"
```
Kipok, line 2: I changed the format here a bit, so the \n\n should go after {context} - can you please make it consistent with other prompt examples?

nemo_skills/inference/server/code_execution_model.py
```diff
@@ -220,6 +220,17 @@ def _recover_from_error(self, request, new_output, executor):
...
         return most_common
 
+    def _handle_stop_words(self, output: str):
+        """
+        OpenAI chat API remove stop word from the output, so we need to add it back
+        to enable code execution.
+        """
+        if not isinstance(self.model, OpenAIModel):
+            return output
+        if output.find(CODE_SEPARATORS[0]) > output.find(CODE_SEPARATORS[-1]):
```
Kipok, line 230: Can you please check if we can do it in the same way as for vllm? The api should return which stop words caused it to finish and we can directly add that instead of hardcoding it for our separators

nemo_skills/inference/prompt/openai/chat.yaml
```diff
@@ -0,0 +1,24 @@
...
+user: |-
+    Here are some examples of questions and solutions followed by a new question that you need to solve.
+
+    {examples}Question:
+    {question}
+
+    {context}
+
+prompt_template: |-
+  <system_start>{system}<system_end>
```
Kipok, line 20: Can you please add a comment here for clarity that these are not used directly but will be parsed out by our api?

nemo_skills/inference/server/model.py
```diff
@@ -272,7 +271,7 @@ def _send_request(
         random_seed: int,
         stop_phrases: list[str],
     ) -> str:
-        messages = prompt.build_structured(input_dict)
+        messages = self._parse_prompt(prompt)
```
Kipok, line 274: Let's remove build_structured from prompt class as we don't need it anymore. We can always bring it back in the future if there is a reason to use it

Kipok (changes requested):

nemo_skills/inference/server/code_execution_model.py
i-vainn: Unfortunately I did not find any way to get the exact stop word. Looks like this is only possible with legacy completions API.

nemo_skills/inference/prompt/openai/chat.yaml
i-vainn: I removed `\n\n` at all because the prompt will anyway be parsed and passed into openai api as a structured list

Kipok (approved):

PR #59, opened 2024-07-15
Fix latex

dgtm777:
Fix visualization of align*


Kipok:
Can we start adding tests for these components?

Kipok (approved):

PR #60, opened 2024-07-15
Changing evalplus install commit to support MBPP v0.2

shtoshni:
Evalplus installation commit to support MBPPv0.2

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 backoff
-evalplus
+git+https://github.com/evalplus/evalplus@c91370f
 huggingface_hub
 hydra-core
 numpy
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 backoff
-evalplus
+evalplus @ git+https://github.com/evalplus/evalplus@c91370f
```
Kipok, line 2: Is their latest package broken? Why do we need to pin it to specific commit? It's probably a good idea to do it in general, since we don't want to have a changed benchmark suddenly, but just trying to understand why it's required for the 0.2 version

shtoshni: They have not released their latest package in a while. 
The last release was on April 2 - https://github.com/evalplus/evalplus/releases/tag/v0.2.1
The MBPP v0.2 change in the repo happened on April 17 - https://github.com/evalplus/evalplus/commit/3de226896daa67b54e2683eb6e5f6840ce66018a
This commit is from April 23. Since this later commit works, I didn't try other commits.

Kipok (approved):

PR #61, opened 2024-07-15
Nemotron fix

shtoshni:
Added eval map for Nemotron models.

### DIFF
--- a/nemo_skills/inference/prompt/nemotron/eval_map.py
+++ b/nemo_skills/inference/prompt/nemotron/eval_map.py
@@ -16,11 +16,10 @@
 # can have a default key that will be used if benchmark name is not in dict
 
 EVAL_MAP = {
-    'base': { 
+    'base': {
         'default': 'nemotron/fewshot',
     },
     'instruct': {  # nemotron-instruct
         'default': 'nemotron/zeroshot',
     },
-    
 }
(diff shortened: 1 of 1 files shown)


nemo_skills/inference/prompt/nemotron/eval_map.py
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# need to contain an EVAL_MAP dictionary mapping model version name to prompts for different benchmarks
+# can have a default key that will be used if benchmark name is not in dict
+
+EVAL_MAP = {
+    'base': {
+        'default': 'nemotron/fewshot',
```
Kipok, line 20: That's not really for the base model as it still uses the extra id tokens. Why do you need to add this file? If you only eval on math benchmarks, you don't need to specify the eval map stuff and our previous commands should work. If you want to eval on both math and other benchmarks, then you need to add other prompts

shtoshni: This file was added because Nemotron-340B math evaluations required this eval_map file. Or we can do without it as well?

Kipok: It should work without it (there was a bug, but it should be fixed now). Let me know if you're running into problems

Kipok: If you do want to merge this, then please also add the prompts for code benchmarks + a proper base prompt (or just remove base). If not, let's just not merge this for now to not confuse people since we only want to have that file when multi-area eval is actually possible

shtoshni: Let's not merge for now then since I haven't tested prompts with Nemotron models for different tasks.

PR #62, opened 2024-07-16
Addin IFEval + refactoring for reporting flexible metrics

Kipok:
Will probably do multiple rounds of refactoring here, but this should be ok to unblock us adding new benchmarks easily

### DIFF
--- a/nemo_skills/evaluation/evaluate_results.py
+++ b/nemo_skills/evaluation/evaluate_results.py
@@ -15,9 +15,11 @@
 import json
 import logging
 import shutil
+import subprocess
 import sys
 from argparse import Namespace
 from dataclasses import field
+from pathlib import Path
 from typing import Any
 
 import hydra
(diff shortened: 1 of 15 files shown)


pipeline/compute_metrics.py
```diff
@@ -28,94 +28,221 @@
...
-        correct_answer = total_correct / total * 100.0
-        no_answer = 0
-        wrong_answer = (total - total_correct) / total
-    else:
-        raise ValueError(f"Unsupported eval_type {eval_type}")
-    return correct_answer, wrong_answer, no_answer, total
+    return evaluator.get_metrics()
+
+
+math_benchmarks = [
```
shtoshni, line 220: Would be good to move these benchmark-related constants to a separate file

nemo_skills/evaluation/evaluate_results.py
```diff
@@ -54,56 +56,104 @@ def __post_init__(self):
...
+            samples = [preprocess_code(json.loads(line)) for line in f]
+        # all changes will be done with a new key "completion", so it's ok to write to the same file
+        with open(jsonl_file, "wt", encoding="utf-8") as f:
+            for sample in samples:
+                f.write(json.dumps(sample) + "\n")
+        eval_config = {
+            "samples": jsonl_file,
+            "base_only": False,
+            "parallel": None,
+            "i_just_wanna_run": False,
```
shtoshni, line 83: "i_just_wanna_run"??

Kipok: That's evalplus parameter :)

pipeline/compute_metrics.py
Kipok: Yeah, eventually, but maybe not in this PR?

nemo_skills/evaluation/evaluate_results.py
shtoshni: Hahaha

pipeline/compute_metrics.py
shtoshni: Sure, thing. May be we can add a ToDo

shtoshni:
The PR looks good to me. Thanks for refactoring the eval and metrics part. I have not tested it, but we can merge if it works with your setup.

shtoshni (approved):
Thanks for refactoring the eval part!

PR #63, opened 2024-07-16
Change reported metrics for if-eval

Kipok:
Changed the reported to standard prompt/instruction strict/loose accuracies. Tested that our numbers roughly match what's in open llm leaderboard

### DIFF
--- a/pipeline/compute_metrics.py
+++ b/pipeline/compute_metrics.py
@@ -28,94 +28,221 @@ from nemo_skills.utils import setup_logging, unroll_files
 LOG = logging.getLogger(__file__)
 
 
-def fill_up_missing(evaluations, log_probs, pred_answers, allow_incomplete):
-    if allow_incomplete:
-        evaluations.append(False)
-        log_probs.append(-1e9)
-        pred_answers.append(None)
-    else:
-        raise RuntimeError("Need to run evaluate_results.py before computing metrics!")
+class MathEval:
+    def __init__(self):
+        self.reset()
+
+    def fill_up_missing(self):
+        return {'predicted_answer': None, 'is_correct': False}
+
+    def is_incomplete(self, elem):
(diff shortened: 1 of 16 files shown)


pipeline/compute_metrics.py
```diff
@@ -140,9 +141,37 @@ def is_incomplete(self, elem):
         if incomplete:
             return False
```
shtoshni, line 142: Should it be True or False?

Kipok: Good catch, should be True

pipeline/compute_metrics.py
```diff
@@ -138,11 +139,39 @@ def fill_up_missing(self):
...
         incomplete = 'loose_eval' not in elem or 'strict_eval' not in elem
         if incomplete:
-            return False
+            return True
         return (
-            'follow_all_instructions' not in elem['loose_eval'] or 'follow_all_instructions' not in elem['strict_eval']
+            'follow_instruction_list' not in elem['loose_eval']
+            or 'follow_instruction_list' not in elem['strict_eval']
+            and 'instruction_id_list' not in elem['loose_eval']
+            or 'instruction_id_list' not in elem['strict_eval']
```
shtoshni, line 147: This is a complex clause imo. There is "or" and "and" mixed in without parenthesis which can make it hard to parse the true condition. I think you're trying to cover for loose and strict eval conditions.  

This might be a cleaner way to get to it:

if 'follow_instruction_list' in elem['loose_eval'] and 'instruction_id_list' in elem['loose_eval']:
     return False
if 'follow_instruction_list' in elem['strict_eval'] and 'instruction_id_list' in elem['strict_eval']:
     return False

return True

Kipok: thanks, in fact there were even more mistakes related to that logic, hopefully it's correct now

shtoshni (approved):

PR #67, opened 2024-07-18
Integrating arena-hard benchmark

Kipok:
Arena hard is the first benchmark where we need to run an LLM to do evaluation, so this PR has quite a bit of changes. For future LLM-as-a-judge benchmarks we should be able to reuse a lot of the logic add here.

Some important details.
1. I'm not getting exactly the same scores as in the original repo, but the relative difference between llama3-8b and 70b is the same. My scores are 26 / 52 and their scores are 20 / 46. There are two potential reasons - one is they have a retry mechanism when gpt-4 doesn't produce an answer and I currently didn't implement that, so there are 2-3% of missing grades. Second, they use different temperature for different categories, and we only use greedy for simplicity and reproducibility.
2. To run the evaluation you'd need to define OPENAI_API_KEY env var. What's going to happen by default is we will submit a batch request which is going to send all the data at once to openai to avoid rate limits and also decrease the cost. Then when you run summarize results, we will check if the request is finished and stop if not. If it is finished the script will fetch the data and include in the generation file, so that it's complete and proceed to report metrics. The problem here is that it requires openai/python10 to be installed where you're running this and so it might not work on slurm login nodes. So either we need to run it in an interactive job or by pulling data to some local machine.
3. Another way to run the evaluation is through NGC hosted models, e.g. with llama3-70b. I found that the scores are very different than with gpt-4, so that might not be a good idea, but can still serve as some quick and cheap way to get the numbers. To do that, the following additional arguments need to be provided. 
```
--extra_eval_args "\
  ++eval_config.judge_model="meta/llama3-70b-instruct" \
  ++eval_config.base_url="https://integrate.api.nvidia.com/v1" \
  ++eval_config.use_batch_api=False \
"
```
This will run the evaluation synchronously, but it's very fast since we submit 100 requests in parallel and this can be increased further (but might run into rate limits at some point). For this to work you also need to define NVIDIA_API_KEY variable

### DIFF
--- a/datasets/arena-hard/prepare.py
+++ b/datasets/arena-hard/prepare.py
@@ -47,6 +47,5 @@ if __name__ == "__main__":
             data['baseline_answer'] = baseline_answers[data['question_id']]
             # will be filled by the evaluation script
             data['judgements'] = []
-            data['judge_scores'] = []
 
             fout.write(json.dumps(data) + "\n")
--- a/datasets/arena-hard/prepare.py
+++ b/datasets/arena-hard/prepare.py
@@ -1,0 +1,48 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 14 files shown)


Kipok:
Merging without review and let's fix issues if any in follow-ups

PR #70, opened 2024-07-22
GPU tests in CI

Kipok:


Kipok:
tests are in a somewhat broken state now in CI (but work locally). But let me merge anyway so that I don't need to keep this PR in sync with main. Should hopefully be fixed soon by changes in the ci configuration + upgraded containers

PR #76, opened 2024-07-26
Update vLLM

titu1994:
Update vLLM dockerfile to 0.5.3.post1 to support Llama 3.1 models


Kipok:
Let's close this for now and revisit when flash-infer is fully tested

PR #78, opened 2024-07-29
Remove `tokenizer_type` when loading a tokenizer using AutoTokenizer

yinglin-nv:
This PR removes the `tokenizer_type` passed to `AutoTokenizer.from_pretrained()`, which is no longer required by recent models such as Llama 3 and Mistral Nemo 12B.


Kipok (approved):
Thanks!

PR #79, opened 2024-07-31
Generate batch support

Alcray:
Added batch generation to generate_solutions and get_batch_results to summarize_results.py

### DIFF
--- a/nemo_skills/inference/generate_solutions.py
+++ b/nemo_skills/inference/generate_solutions.py
@@ -148,7 +148,7 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
 
                     # saving the request id to be able to retrieve results when they are ready
                     with open(cfg.output_file + '-batch-request-id', 'wt', encoding='utf-8') as batch_fout:
-                        batch_fout.write(json.dumps({'request_id': request_metadata.id}))
+                        batch_fout.write(json.dumps({'request_id': request_metadata.id, "generation_key": cfg.generation_key}))
                     LOG.info('Submitted batch evaluation request. Please wait for the results to be ready.')
                     LOG.info('The current status and final results can be accessed through summarize_results.py')
                     LOG.info('Request metadata: %s', str(request_metadata))
--- a/nemo_skills/inference/generate_solutions.py
+++ b/nemo_skills/inference/generate_solutions.py
@@ -133,8 +133,7 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
                 break
             data_points.append(data_point)
 
-            if len(data_points) == cfg.batch_size or idx == len(data) - 1:
-                if cfg.inference.use_batch_api:
+            if cfg.inference.use_batch_api and len(data_points) == len(data):
                     # Using batch API for generation
                     # Accessing the model directly since CodeExecutionWrapper does not support batch_generate
                     request_metadata = llm.model.batch_generate(
@@ -156,12 +155,13 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
 
                     # Clear data_points as we've submitted the batch
                     data_points = []
-                else:
+
+            elif not cfg.inference.use_batch_api and (len(data_points) == cfg.batch_size or idx  == len(data) - 1):
                     # batch-computing the outputs
                     outputs = llm.generate(
                         prompts=[prompt.build_string(data_point) for data_point in data_points],
                         stop_phrases=list(cfg.prompt.stop_phrases),
-                        **asdict(cfg.inference),
+                        **(asdict(cfg.inference).pop('use_batch_api', None)),
                     )
 
                     for output, original_data_point in zip(outputs, data_points):
@@ -176,21 +176,6 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
 
                         fout.write(json.dumps(output) + "\n")
                     data_points = []
-
-        # collecting the final batch
-        if len(data_points) > 0 and not cfg.inference.use_batch_api:
-            outputs = llm.generate(
-                prompts=[prompt.build_string(data_point) for data_point in data_points],
-                stop_phrases=list(cfg.prompt.stop_phrases),
-                **asdict(cfg.inference),
-            )
-            for output, original_data_point in zip(outputs, data_points):
-                output.update(original_data_point)
-                if 'error_message' not in output:
-                    output['error_message'] = extract_error_message(output['generation'])
-                if cfg.generation_key != "generation":
-                    output[cfg.generation_key] = output.pop("generation")
-                fout.write(json.dumps(output) + "\n")
 
 
 error_recovery_params = '\n' + get_fields_docstring(
--- a/nemo_skills/inference/generate_solutions.py
+++ b/nemo_skills/inference/generate_solutions.py
@@ -134,13 +134,18 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
 
             data_points.append(data_point)
 
-            if len(data_points) == cfg.batch_size:
+            if len(data_points) == cfg.batch_size or idx == len(data) - 1:
                 if cfg.inference.use_batch_api:
                     # Using batch API for generation
-                    request_metadata = llm.batch_generate(
+                    # Accessing the model directly since CodeExecutionWrapper does not support batch_generate
+                    request_metadata = llm.model.batch_generate(
                         prompts=[prompt.build_string(data_point) for data_point in data_points],
                         tokens_to_generate=cfg.inference.tokens_to_generate,
                     )
+                    with open(cfg.output_file, 'at', encoding='utf-8') as fout:
+                        for data_point in data_points:
+                            fout.write(json.dumps(data_point) + '\n')
+
                     # saving the request id to be able to retrieve results when they are ready
                     with open(cfg.output_file + '-batch-request-id', 'wt', encoding='utf-8') as batch_fout:
                         batch_fout.write(json.dumps({'request_id': request_metadata.id}))
@@ -150,7 +155,6 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
                     
                     # Clear data_points as we've submitted the batch
                     data_points = []
-                    continue
                 else:
                     # batch-computing the outputs
                     outputs = llm.generate(
--- a/nemo_skills/inference/generate_solutions.py
+++ b/nemo_skills/inference/generate_solutions.py
@@ -42,6 +42,7 @@ class InferenceConfig:
     random_seed: int = 0
     tokens_to_generate: int = 2048
     repetition_penalty: float = 1.0
+    use_batch_api: bool = False  # Whether to use batch API for generation
 
 
 @nested_dataclass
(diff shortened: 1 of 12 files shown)


nemo_skills/inference/prompt/few_shot_examples/examples_proof.py
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+        "question": "Denoting the sums of the first $n_1$, first $n_2$ and first $n_3$ terms of an arithmetic progression by $S_1$, $S_2$ and $S_3$, respectively, show that $$\\frac{S_1}{n_1}(n_2-n_3)+\\frac{S_2}{n_2}(n_3-n_1)+\\frac{S_3}{n_3}(n_1-n_2)=0.$$",
+        "classification": """
+        Classification: proof_question
+""".strip(),
+    },
+    {
+        "question": "Solve the following equations: $\\frac{\\sin(60^\\circ+x)+\\sin(60^\\circ-x)}{2} =\\frac{\\tan x}{(1+\\tan^2 x)^2}+\\frac{\\cot x}{(1+\\cot^2 x)^2}$",
+        "classification": """
```
shtoshni, line 28: Why couldn't this be "generation": "False"  and "generation": "True" (True for proof_question)

nemo_skills/inference/generate_solutions.py
```diff
@@ -133,29 +134,50 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
...
-                    output.update(original_data_point)
-                    if 'error_message' not in output:
-                        output['error_message'] = extract_error_message(output['generation'])
-
-                    if cfg.generation_key != "generation":
-                        output[cfg.generation_key] = output.pop("generation")
-
-                    fout.write(json.dumps(output) + "\n")
-                data_points = []
+            if len(data_points) == cfg.batch_size or idx == len(data) - 1:
```
Kipok, line 137: Doesn't this mean that if use_batch_api=True and then default batch_size=16, then we will submit many batch_generate requests, which will override the same file? Or do I misunderstand something here? If that's the case, probably best to put it out of loop to make sure it's always done with full data and batch_size will be ignored in this case

nemo_skills/inference/generate_solutions.py
```diff
@@ -133,29 +134,50 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
...
+                    )
+                    with open(cfg.output_file, 'at', encoding='utf-8') as fout:
+                        for data_point in data_points:
+                            fout.write(json.dumps(data_point) + '\n')
+
+                    # saving the request id to be able to retrieve results when they are ready
+                    with open(cfg.output_file + '-batch-request-id', 'wt', encoding='utf-8') as batch_fout:
+                        batch_fout.write(json.dumps({'request_id': request_metadata.id}))
+                    LOG.info('Submitted batch evaluation request. Please wait for the results to be ready.')
+                    LOG.info('The current status and final results can be accessed through summarize_results.py')
```
Kipok, line 153: Since this logic is decoupled from evaluation (there are cases where we run generation, but there are no metrics to compute), how about we add a new script inside pipeline, called `check_batch_gen.py` or something like that, which you will run with the same jsonl file list and it will do the check and fuse the results back into the original files

nemo_skills/inference/generate_solutions.py
```diff
@@ -133,29 +134,50 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
...
-                    if cfg.generation_key != "generation":
-                        output[cfg.generation_key] = output.pop("generation")
-
-                    fout.write(json.dumps(output) + "\n")
-                data_points = []
+            if len(data_points) == cfg.batch_size or idx == len(data) - 1:
+                if cfg.inference.use_batch_api:
+                    # Using batch API for generation
+                    # Accessing the model directly since CodeExecutionWrapper does not support batch_generate
+                    request_metadata = llm.model.batch_generate(
```
Kipok, line 141: This is a bit brittle since not all models support batch generate, only openai and we want to be explicit that we are not using code gen wrapper as users might expect it to be used. How about instead we add a new parameter `use_code` that default to False and then if it's False we don't wrap the model in CodeExecutionWrapper. Please also add a `batch_generate` method to the base model that will by default raise NotImplementedError, so that it's clear that this functionality is not implemented when someone tries to use it with models other than openai

nemo_skills/inference/generate_solutions.py
```diff
@@ -133,29 +134,50 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
...
+
+                    for output, original_data_point in zip(outputs, data_points):
+                        # to make it easier to follow up with evaluation and limit accidental errors, we are adding
+                        # all of the ground-truth data to the output file alongside the generated solutions
+                        output.update(original_data_point)
+                        if 'error_message' not in output:
+                            output['error_message'] = extract_error_message(output['generation'])
+
+                        if cfg.generation_key != "generation":
+                            output[cfg.generation_key] = output.pop("generation")
```
Kipok, line 174: We also need to use the right generation key for the batch api, can you please save it to the batch-request-id so that it can later be used to properly write data in the same way as non-batch api?

nemo_skills/inference/prompt/few_shot_examples/examples_proof.py
Kipok: Do you think it will work better or just for simplicity? I think it probably does not matter what we use, but potentially we might extend it to more categories in the future (e.g. questions with images), in which case we'd need to use something like what's currently here anyway

pipeline/summarize_results.py
```diff
@@ -31,6 +32,52 @@
...
+                with open(batch_request_file, 'rt', encoding='utf-8') as fin:
+                    request_id = json.load(fin)['request_id']
+
+                from nemo_skills.inference.server.model import get_model
+
+                llm = get_model(server_type='openai', model='gpt-4-1106-preview')
+                metadata, outputs = llm.get_batch_results(request_id)
+
+                if outputs is None:
+                    LOG.warning(f"Judgements are not ready yet for {jsonl_file}! Current status: {metadata}")
```
Kipok, line 53:
```suggestion
                    LOG.warning("Batch generations are not ready yet for %s! Current status: %s", jsonl_file, metadata)
```

pipeline/summarize_results.py
```diff
@@ -31,6 +32,52 @@
...
+                    continue
+
+                # Read existing data from the jsonl file
+                with open(jsonl_file, 'rt', encoding='utf-8') as fin:
+                    data = [json.loads(line) for line in fin]
+
+                # Update data with judgements
+                for data_point, output in zip(data, outputs):
+                    data_point['judgement'] = output['generation']
```
Kipok, line 62: Let's use the "generation_key" here that you would add into the batch-request-id file

pipeline/summarize_results.py
```diff
@@ -31,6 +32,52 @@
...
+                with open(jsonl_file, 'wt', encoding='utf-8') as fout:
+                    for data_point in data:
+                        fout.write(json.dumps(data_point) + '\n')
+
+                batch_request_file.unlink()
+                LOG.info(f"Processed batch results for {jsonl_file}")
+            except PermissionError:
+                user = os.getenv('USER', 'your_username')
+                LOG.error(f"Permission denied when trying to access {jsonl_file} or {batch_request_file}")
+                print(f"Permission denied. Try running the following command to change file ownership:")
```
Kipok, line 74: Let's put it all into the LOG.error for consistency

pipeline/summarize_results.py
```diff
@@ -31,6 +32,52 @@
 from nemo_skills.evaluation.metrics import MathEval
 from nemo_skills.utils import setup_logging
 
+LOG = logging.getLogger(__name__)
+
+
+def process_batch_results(prediction_jsonl_files):
```
Kipok, line 38: Can you please reuse this logic in here https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/evaluation/metrics.py#L346 and in here https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/evaluation/metrics.py#L54

Kipok (changes requested):
Thanks! Looks good overall, but let's do a bit of refactoring to make it more consistent and easier to use

nemo_skills/inference/generate_solutions.py
Alcray: It does, I am not sure if openai has any restriction on the batch size, but for my experiments I was using 1024 so that the whole dataset would be covered in a single batch. 
If openai does not have any restriction, then I can modify the code so that it takes batch size nearest power of 2 that is greater then the actual dataset size

pipeline/summarize_results.py
Alcray: Done

nemo_skills/inference/generate_solutions.py
Alcray: Do you mean to handle batch result getting in a `check_batch_gen.py` instead of `summarize_results.py`?

pipeline/summarize_results.py
Alcray: Do you want me to rewrite `def setup(self, prediction_jsonl_files):` the way I did `def process_batch_results(prediction_jsonl_files):`, or vice versa?

nemo_skills/inference/generate_solutions.py
Alcray: Done

nemo_skills/inference/generate_solutions.py
Kipok: Why does it need to be a power of 2? Let's just always send the whole data. If there is some restriction or error, users can always manually split the file. If we find ourselves doing this too often in the future, we can automate that, but I'd suggest to use a simpler approach for now

nemo_skills/inference/generate_solutions.py
Kipok: Yes, and the summarize_results.py (or actually better the compute_metrics function) will call that script first

Alcray: Done

nemo_skills/inference/generate_solutions.py
Alcray: Done

nemo_skills/inference/generate_solutions.py
```diff
@@ -130,15 +131,37 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
         for idx, data_point in tqdm(enumerate(data), initial=starting_idx, total=len(data) + starting_idx):
             if idx >= cfg.max_samples:
                 break
-
             data_points.append(data_point)
 
-            if len(data_points) == cfg.batch_size:
+            if cfg.inference.use_batch_api and len(data_points) == len(data):
```
Kipok, line 136: Let's move this outside of the loop

nemo_skills/inference/generate_solutions.py
```diff
@@ -130,15 +131,37 @@ def generate_solutions(cfg: GenerateSolutionsConfig):
...
+                        json.dumps({'request_id': request_metadata.id, "generation_key": cfg.generation_key})
+                    )
+                LOG.info('Submitted batch evaluation request. Please wait for the results to be ready.')
+                LOG.info('The current status and final results can be accessed through summarize_results.py')
+                LOG.info('Request metadata: %s', str(request_metadata))
+
+                # Clear data_points as we've submitted the batch
+                data_points = []
+
+            elif not cfg.inference.use_batch_api and (len(data_points) == cfg.batch_size or idx == len(data) - 1):
```
Kipok, line 159: Let's use cfg.max_samples instead of len(data)

Alcray: Done

Kipok:
If we need this in the future, let's reopen and update

PR #80, opened 2024-07-31
Shtoshni/llama3.1 ft

shtoshni:
This PR does:
- Fix preprocessing error of not using the skip_first argument to skip the validation dataset + other minor nits
- Adds a maths SFT prompt.
- Fix prompts for Llama3 and Nemotron


nemo_skills/finetuning/data_preparation_utils/prepare_sft_data.yaml
```diff
@@ -18,6 +18,9 @@ prompt_type: openmathinstruct/sft
 chat_format: false  # whether to use NeMo's chat format
 generation_suffix: ""  # suffix to add to the end of expected response (e.g. <|eot_id|>)
 
+add_eos: false
```
Kipok, line 21: This is already supported with a `generation_suffix` parameter above

nemo_skills/inference/prompt/llama3/math_sft.yaml
```diff
@@ -0,0 +1,15 @@
+system: ""
+
+user: |- 
+  Help the user to solve the given math problem. Make sure to put the answer (and only answer) inside \boxed{{}}.
+
+  {question}{context}
+
+prompt_template: |-
+  <|start_header_id|>user<|end_header_id|>
```
Kipok, line 9:
```suggestion
  <|begin_of_text|><|start_header_id|>user<|end_header_id|>
```

Kipok (changes requested):

nemo_skills/inference/prompt/llama3/eval_map.py
```diff
@@ -36,5 +36,8 @@
         'human-eval': 'llama3/codegen',
         'mbpp': 'llama3/codegen',
         'mmlu': 'llama3/mmlu',
+        # Math benchmarks
+        'gms8k': 'llama3/math_sft',
```
Kipok, line 40: Can you please update default to be math_sft and then explicitly set llama3/sft for ifeval/arena-hard instead? We have a ton of math benchmarks and it's better if we don't need to specify a specific prompt for each of them if we want to evaluate later

nemo_skills/inference/prompt/llama3/eval_map.py
```diff
@@ -36,5 +36,8 @@
         'human-eval': 'llama3/codegen',
```
Kipok, line 36: Do these need to be changed to remove system message as well?

nemo_skills/inference/prompt/llama3/mmlu.yaml
```diff
@@ -1,20 +1,22 @@
+# Prompt used by Llama-base/instruct models for MMLU benchmark 
+# TODO - Seems to be different from https://huggingface.co/datasets/meta-llama/Meta-Llama-3.1-8B-evals/viewer/Meta-Llama-3.1-8B-evals__mmlu__details?row=3
```
Kipok, line 2: Let's switch to the one they use? No reason not to use it, I guess

nemo_skills/inference/prompt/nemotron/base.yaml
```diff
@@ -0,0 +1,26 @@
...
+  {examples}Question:
+  {question}{context}
+
+  Don't forget that your final answer should be inside \boxed{{}}!
+
+prompt_template: |-
+  <extra_id_0>System
+  {system}
+  <extra_id_1>User
+  {user}
```
Kipok, line 21: Is newline expected here? Not fully sure, but good to double check

nemo_skills/inference/prompt/nemotron/codegen.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # Your codes here
+  ```
+
+  @@ Response
+
+prompt_template: |-
+  <extra_id_0>System
+  {system}
+  <extra_id_1>User
+  {user}<extra_id_1>Assistant
```
Kipok, line 28: Let's be consistent, either always have new line or never have it

nemo_skills/inference/prompt/nemotron/instruct.yaml
```diff
@@ -0,0 +1,18 @@
+# Generic prompt for Nemotron instruct
```
Kipok, line 1: This is not generic, but math specific right? The generic one should probably just have question with no other instructions

nemo_skills/inference/prompt/nemotron/mmlu.yaml
```diff
@@ -0,0 +1,20 @@
+system: ""
+
+user: |-
+    Answer the following multiple choice question. The last line of your response should be of the following format: 'Answer: \boxed{{$LETTER}}' (without quotes) where LETTER is one of ABCD. Think step by step before answering.
```
Kipok, line 4: We can actually remove the \boxed{} thing from mmlu prompts now - can you please add a new parameter here https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/evaluation/settings.py#L61 with something like this

```
'mmlu': '++eval_config.extract_from_boxed=False ++eval_config.extract_regex="Answer: (.+)$"'
```

The boxed format is obviously not what's typically used here and we should probably not be using it as well

tests/test_prompts.py
```diff
@@ -109,9 +109,7 @@ def test_llama3_instruct_prompt():
     config.few_shot_examples.num_few_shots = 2
     prompt = Prompt(config=config)
 
-    expected_prompt = """<|begin_of_text|><|start_header_id|>system<|end_header_id|>
-
-<|eot_id|><|start_header_id|>user<|end_header_id|>
+    expected_prompt = """<|start_header_id|>user<|end_header_id|>
```
Kipok, line 112: There are some tests that are failing because of the newly added bos tokens, can you please fix references there?

nemo_skills/inference/prompt/llama3/instruct.yaml
```diff
@@ -1,23 +1,21 @@
+# Generic prompt for Llama instruct
```
Kipok, line 1: math specific?

nemo_skills/inference/prompt/nemotron/eval_map.py
```diff
@@ -0,0 +1,31 @@
...
+# need to contain an EVAL_MAP dictionary mapping model version name to prompts for different benchmarks
+# can have a default key that will be used if benchmark name is not in dict
+
+EVAL_MAP = {
+    'base': {
+        'default': 'openmathinstruct/base',
+    },
+    'instruct': {  # nemotron-instruct
+        'default': 'nemotron/instruct',
+        'fewshot': 'nemotron/fewshot',
```
Kipok, line 24: this only works for benchmark names, so specifying fewshot here has no effect

nemo_skills/inference/prompt/nemotron/fewshot_instruct.yaml
```diff
@@ -1,16 +1,17 @@
+# Generic prompt for Nemotron-instruct fewshot
```
Kipok, line 1: math-specific?

nemo_skills/inference/prompt/nemotron/mmlu.yaml
```diff
@@ -0,0 +1,21 @@
+system: ""
+
+user: |-
+  Answer the following multiple choice question. The last line of your response should be of the following format: 'Answer: \boxed{{$LETTER}}' (without quotes) where LETTER is one of ABCD. Think step by step before answering.
```
Kipok: Let's use the same as for llama?

tests/test_prompts.py
```diff
@@ -452,17 +445,55 @@ def test_llama3_math_prompt():
...
         prompt.build_string(
             {
                 'question': "Find the sum of all complex values of $a,$ such that the polynomial $x^4 + (a^2 - 1) x^2 + a^3$ has exactly two distinct complex roots."
             }
         )
+        == expected_prompt
     )
+
+
+def test_llama3_math_prompt():
```
Kipok, line 458:
```suggestion
def test_llama3_mmlu_prompt():
```

nemo_skills/inference/prompt/nemotron/eval_map.py
```diff
@@ -0,0 +1,31 @@
...
+
+# need to contain an EVAL_MAP dictionary mapping model version name to prompts for different benchmarks
+# can have a default key that will be used if benchmark name is not in dict
+
+EVAL_MAP = {
+    'base': {
+        'default': 'openmathinstruct/base',
+    },
+    'instruct': {  # nemotron-instruct
+        'default': 'nemotron/instruct',
```
Kipok, line 23: math by default?

nemo_skills/inference/prompt/nemotron/eval_map.py
```diff
@@ -0,0 +1,30 @@
...
+    'base': {
+        'default': 'openmathinstruct/base',
+    },
+    'instruct': {  # nemotron-instruct
+        'default': 'nemotron/math',
+        'gsm8k': 'nemotron/math',
+        'math': 'nemotron/math',
+        'human-eval': 'nemotron/codegen',
+        'mbpp': 'nemotron/codegen',
+        'mmlu': 'nemotron/mmlu',
```
Kipok, line 28: can you add ifeval and arena-hard also?

Kipok (approved):

Kipok:
Thanks!

PR #81, opened 2024-08-01
Multiple fixes

Kipok:
1. Remove "context" parameter as it's not really used anymore and is just confusing for people.
2. Update trtllm arguments in the docs
3. Add llama3 chat format support to sft and data preparation. This is a bit hacky at the moment and will be fixed when this is properly supported in nemo
4. Remove system message from llama3 sft config
5. Fix gpu tests and upgrade finetuning test to llama3


docs/reproducing-results.md
```diff
@@ -45,11 +45,11 @@ documentation to learn how to make inference more efficient.
...
-      --gpt_attention_plugin bfloat16 \
-      --gemm_plugin bfloat16 \
-      --context_fmha "enable" \
-      --max_input_len 4096 \
-      --max_output_len 512 \
+      --gpt_attention_plugin bfloat16
+      --use_paged_context_fmha enable
+      --max_input_len 3584 \
+      --max_seq_len 4096 \
+      --max_num_tokens 4096 \
```
shtoshni, line 52: Is this right? Isn't `max_num_tokens` = `max_input_len + max_seq_len `

shtoshni:
I'll finish the review by 3 pm PST today.

docs/reproducing-results.md
Kipok: No, max_num_tokens is actually more like batch size. It's the maximum number of generation tokens the engine can do at the same time (across all prompts it's working on). Not really sure why we need to specify both that and the batch size, but somehow that's the current way to do it in trtllm. That being said, these particular parameters are certainly not optimal, but probably good enough and close to what we actually used with prior trtllm version

shtoshni (approved):

docs/reproducing-results.md
shtoshni, line 52: Maybe a good idea to add a pointer in this doc to the original TRTLLM script?

nemo_skills/inference/prompt/llama3/base.yaml
shtoshni, line 1: Can you also remove `context` from:
- `nemo_skills/inference/prompt/nemotron/math.yaml`
- `nemo_skills/inference/prompt/nemotron/sft.yaml`
- `nemo_skills/inference/prompt/llama3/math_sft.yaml`

docs/reproducing-results.md
Kipok: There is a pointer to their repo in the checkpoint conversion doc, this one is just for reproducing our results, so it seems proper to just include the command there that we used.

In general, there is unfortunately not really any good documentation page in trtllm repo, so we can't link to something more specific than just their github repository

nemo_skills/inference/prompt/llama3/base.yaml
Kipok: done, thanks!

PR #82, opened 2024-08-01
Optimizing SFT recipe

Kipok:
Adding more optimal parameters for training llama3 8B model + fixing a bug with incorrect timeout


Kipok:
Let me merge this one as it contains some important fixes. We can follow up with another PR if there are any problems.

PR #86, opened 2024-08-08
Fix fill output args

yinglin-nv:
`batch_input_ids_list` is now required by the `_fill_output()` function in ModelCppRunner. In this PR, we pass this arg to `runner._fill_output()` in `_stream()`.


Kipok:
Thanks @yinglin-nv! I ported this to #87 along with some other changes related to upgrade, so let's close this one.

PR #91, opened 2024-08-21
Rename the tool

dgtm777:
1. Rename the tool
2. Remove context_template logic
3. Add github tests


visualization/requirements.txt
```diff
@@ -1,8 +1,8 @@
-ansi2html
-dash
-dash_bootstrap_components
-hydra-core
-joblib
-pandas
-pygments
-sshtunnel_requests
+ansi2html==1.9.2
```
Kipok, line 1: Let's not fix versions unless there is a good reason to. If we find errors because of package upgrade in the future, we can upper-bound it, but on reason to limit the installed versions otherwise

visualization/test-requirements.txt
```diff
@@ -1,2 +1,3 @@
-dash[testing]
-webdriver_manager
+dash[testing]==2.17.0
```
Kipok, line 1: maybe we can add a new requirements folder where we put all relevant req files. Let's have "main", "inspector" and "tests" there. With tests containing what you need here + pytest and pytest-cov that we currently install in the actions pipeline

visualization/tests/README.md
```diff
@@ -1,7 +1,8 @@
-To launch tests firstly install requirements for the tool and for tests
+To launch tests firstly install all the requirements
 ```
+pip install -r requirements.txt
 pip install -r visualization/requirements.txt
-pip install -r visualization/tests/requirements.txt
+pip install -r visualization/test-requirements.txt
```
Kipok, line 5: Does it make sense to rename "visualization" to "nemo_inspector" then?

Kipok (changes requested):

visualization/test-requirements.txt
dgtm777: What about datasets/requirements? Do we want to move them to the folder as well?

Kipok: Yes!

Kipok (approved):

PR #98, opened 2024-08-30
Modify sandbox code execution to capture stderr

aleksficek:
Can we make this change so that both the stdout and stderr outputs are captured in sandbox code execution? This also resembles the return object from Piston. Additionally, this makes code execution when the code fails much faster as it does not wait for timeout to occur and fails immediately. 

I'm not sure what NeMo-Skills code will break with this change so let me know what uses this and I can make the changes if needed.

### DIFF
--- a/nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
+++ b/nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
@@ -34,8 +34,21 @@ def execute_code_subprocess(generated_code, queue):
     resource.setrlimit(resource.RLIMIT_STACK, (limit, limit))
 
     sys.stdout = StringIO()
-    exec(generated_code, {})
-    queue.put(sys.stdout.getvalue())
+    sys.stderr = StringIO()
+
+    try:
+        exec(generated_code, {})
+    except Exception as e:
+        return_dict = {
+            "stdout": sys.stdout.getvalue(),
+            "stderr": repr(e),
+        }
+    else:
+        return_dict = {
+            "stdout": sys.stdout.getvalue(),
+            "stderr": sys.stderr.getvalue(),
+        }
+    queue.put(return_dict)
 
 
 @app.route("/execute", methods=["POST"])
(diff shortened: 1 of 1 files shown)


Kipok:
Yeah, this will break a bunch of downstream code. But you can already get error message if you use the Sandbox class (which you should be using as it also provides a unified interface between piston and local exec) https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/code_execution/sandbox.py#L224. Here are some examples how you can use it https://github.com/Kipok/NeMo-Skills/blob/main/tests/test_code_execution.py#L39

Also please note that this logic is in the process of being refactored and eventually (in a couple of weeks) we will switch to this version that does provide you stdout/stderr explicitly https://github.com/Kipok/NeMo-Skills/pull/89. But hopefully for your purposes the current sandbox api is already good enough (it shouldn't wait for timeout btw, if the code has an error, it will finish immediately). Let me know if you still need to make some changes, happy to discuss that

aleksficek:
> Yeah, this will break a bunch of downstream code. But you can already get error message if you use the Sandbox class (which you should be using as it also provides a unified interface between piston and local exec) https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/code_execution/sandbox.py#L224. Here are some examples how you can use it https://github.com/Kipok/NeMo-Skills/blob/main/tests/test_code_execution.py#L39
> 
> Also please note that this logic is in the process of being refactored and eventually (in a couple of weeks) we will switch to this version that does provide you stdout/stderr explicitly #89. But hopefully for your purposes the current sandbox api is already good enough (it shouldn't wait for timeout btw, if the code has an error, it will finish immediately). Let me know if you still need to make some changes, happy to discuss that

Got it thanks for clearing that up. So I have something using my own local custom setup currently working, I'll merge and use that until the refactors come in and then switch NeMo-Codegen to the new code in NeMo-Skills for consistency. Ping me on slack when the changes get in and I can make the updates.

Kipok:
sounds good, thanks @aleksficek. Let me close this for now then

PR #99, opened 2024-08-31
Fix regex escape for parsing of end tokens

titu1994:
Fixes a minor issue with regex escape for certain special stop words (like `?`) used for magpie continuations.

### DIFF
--- a/nemo_skills/inference/server/model.py
+++ b/nemo_skills/inference/server/model.py
@@ -31,7 +31,9 @@ def remove_stop_phrases(text: str, stop_phrases: list[str]) -> str:
     """Removes everything after the last stop token."""
     if not stop_phrases:
         return text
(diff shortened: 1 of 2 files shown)


nemo_skills/inference/server/model.py
```diff
@@ -31,7 +31,9 @@ def remove_stop_phrases(text: str, stop_phrases: list[str]) -> str:
     """Removes everything after the last stop token."""
     if not stop_phrases:
         return text
-    return re.split("|".join([sp.replace('|', '\\|') for sp in stop_phrases]), text, maxsplit=1)[0]
+    # Escape all special characters in stop phrases
+    escaped_stop_phrases = [re.escape(sp) for sp in stop_phrases]
```
Kipok, line 35: This function is also duplicated inside serve_trt.py, can you please add this fix there as well?

titu1994: Done, thanks ! Missed that..

Kipok (approved):

PR #102, opened 2024-09-10
Added dataset groups to prepare.py

i-vainn:


Kipok (commented):
Thanks! Can you please add a test for this (and also fix the existing test for dataset prepation that's likely broken)?

Kipok (approved):

PR #104, opened 2024-09-11
Added uncommitted code check

i-vainn:


Kipok:
Integrated into nemo-run directly and merged through #112

PR #105, opened 2024-09-11
Summarize results from cluster

i-vainn:
summarize_results can now fetch results from cluster


Kipok:
Merged a modified version in #111

PR #107, opened 2024-09-11
Updated prompts

i-vainn:
Placed all config files inside `configs` dir and changed `num_few_shots` handling.


Kipok:
Merged with fixes in #113

PR #129, opened 2024-09-27
Add support for env variable passthrough and env variable mounting

titu1994:
This pr adds a few flags for processing by the server creation engine before it executes

The server config can have the following additional args

- `env_vars`: Optional environ variables that will be copied as is from user environment to execution environment 
- `required_env_vars`: Mandatory environ variables that will be checked for existence before being copied to execution environment
- `env_mounts`: A user defined list mapping from `<STR | ENV_VAR>:<STR | ENV_VAR>`. The engine will resolve the ENV_VAR at runtime, and consider it a mandatory argument to be passed by the user.

This enables for dynamic mounting by the user without the hardcoding of local and slurm mount paths - all users can share the same config file, and simply populate the mount paths via personal environment variables provided in their private .bashrc or .profile (or even explicit export). 

This will reduce the number of copy pasted user-defined versions of the example server configs and prevent users from having to manually maintain and update their configs as and when the core repo updates with newer containers.

# Example

```yaml
# cluster_configs/local.yaml
...

# optional env variables
env_vars: ['NEMO_SKILLS_CONFIG']
# required env variables
required_env_vars: ['HF_HOME', 'HF_TOKEN']

# mounting determined by environment variables
env_mounts:
  - "{HF_HOME}:/cache/huggingface"
```

### DIFF
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import os.path
+import os
 from argparse import ArgumentParser
 
 import nemo_run as run
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -25,7 +25,7 @@ if __name__ == "__main__":
     parser.add_argument("--config_dir", default=None, help="Path to the cluster_configs dir")
     parser.add_argument("--log_dir", required=False, help="Can specify a custom location for slurm logs")
     parser.add_argument("--cluster", required=True, help="One of the configs inside cluster_configs")
-    parser.add_argument("--model", required=False, default=None, help="Path to the model.")
+    parser.add_argument("--model", required=True, help="Path to the model.")
     parser.add_argument(
         "--server_type",
         choices=('nemo', 'trtllm', 'vllm'),
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import os.path
 from argparse import ArgumentParser
 
 import nemo_run as run
@@ -25,7 +25,7 @@ if __name__ == "__main__":
     parser.add_argument("--config_dir", default=None, help="Path to the cluster_configs dir")
     parser.add_argument("--log_dir", required=False, help="Can specify a custom location for slurm logs")
     parser.add_argument("--cluster", required=True, help="One of the configs inside cluster_configs")
-    parser.add_argument("--model", required=True, help="Path to the model.")
+    parser.add_argument("--model", required=False, default=None, help="Path to the model.")
     parser.add_argument(
         "--server_type",
         choices=('nemo', 'trtllm', 'vllm'),
@@ -51,7 +51,12 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     cluster_config = get_cluster_config(args.cluster, args.config_dir)
-    check_if_mounted(cluster_config, args.model)
+
+    if args.server_type != "vllm":
+        check_if_mounted(cluster_config, args.model)
+    elif args.server_type == "vllm" and os.path.exists(args.model):
+        check_if_mounted(cluster_config, args.model)
+
     if args.log_dir:
         check_if_mounted(cluster_config, args.log_dir)
 
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import os
 from argparse import ArgumentParser
 
 import nemo_run as run
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -52,11 +52,6 @@ if __name__ == "__main__":
 
     cluster_config = get_cluster_config(args.cluster, args.config_dir)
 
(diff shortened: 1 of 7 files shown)


nemo_skills/pipeline/utils.py
```diff
@@ -266,6 +267,100 @@ def get_packager():
...
+    Returns:
+        dict: dictionary of environment
+    """
+    env_vars = {}
+    # Check for user requested env variables
+    required_env_vars = cluster_config.get("required_env_vars", [])
+    for env_var in required_env_vars:
+        if env_var not in os.environ:
+            raise ValueError(f"Required environment variable {env_var} not found.")
+        env_vars[env_var] = os.environ[env_var]
```
Kipok, line 290: Let's add a log here as well, so that all added variables are printed

nemo_skills/pipeline/start_server.py
```diff
@@ -51,7 +51,12 @@
     args = parser.parse_args()
 
     cluster_config = get_cluster_config(args.cluster, args.config_dir)
-    check_if_mounted(cluster_config, args.model)
+
+    if args.server_type != "vllm":
+        check_if_mounted(cluster_config, args.model)
+    elif args.server_type == "vllm" and os.path.exists(args.model):
```
Kipok, line 57: I think this check is not going to work, since args.model has to be the mounted path, not real path on your file system. So we cannot check if it exists

Kipok: What we can do instead is check that it starts with / since all mounted paths have to be formatted that way

nemo_skills/pipeline/start_server.py
```diff
@@ -51,7 +51,12 @@
     args = parser.parse_args()
 
     cluster_config = get_cluster_config(args.cluster, args.config_dir)
-    check_if_mounted(cluster_config, args.model)
+
+    if args.server_type != "vllm":
```
Kipok, line 55: Let's move this logic inside `get_server_command` so that it automatically applies anywhere server is used. Please also remove the mount check for args.model from generate.py, eval.py, check_contamination.py and llm_math_judge.py (we should unify the common code in those scripts, but not done yet)

nemo_skills/pipeline/utils.py
```diff
@@ -266,6 +267,100 @@ def get_packager():
...
+    # if env_vars is None, we will get the env_vars from the cluster config
+    if env_vars is None:
+        env_vars = get_env_variables(cluster_config)
+
+    # if there are no env_mounts, we will return the mounts as is
+    if 'env_mounts' not in cluster_config:
+        return mounts
+
+    # if there are env_mounts, we will add the mounts from the env_mounts
+    env_mounts = cluster_config['env_mounts']
```
Kipok, line 329: I think it's better to just support it in the mounts field. So we'd only have a single mounts field that can be either normal mounts or from env vars like in your current logic

nemo_skills/pipeline/utils.py
```diff
@@ -280,12 +375,17 @@ def get_executor(
...
+    env_vars = get_env_variables(cluster_config)
+    config_mounts = get_mounts_from_config(cluster_config, env_vars)
+
     mounts = mounts or config_mounts
     packager = get_packager()
     if cluster_config["executor"] == "local":
         if num_nodes > 1:
             raise ValueError("Local executor does not support multi-node execution")
+
+        env_vars["PYTHONUNBUFFERED"] = "1"  # this makes sure logs are streamed right away]
```
Kipok, line 387:
```suggestion
        env_vars["PYTHONUNBUFFERED"] = "1"  # this makes sure logs are streamed right away
```

Kipok (changes requested):
Thanks! Just a couple of small comments

nemo_skills/pipeline/utils.py
titu1994: Done

nemo_skills/pipeline/start_server.py
titu1994: Done

nemo_skills/pipeline/start_server.py
titu1994: I see, good point

nemo_skills/pipeline/utils.py
```diff
@@ -266,6 +277,100 @@ def get_packager():
...
+        if ":" not in mount:
+            raise ValueError(f"Invalid mount format: {mount}. The mount path must be separated by a colon.")
+
+        mount_source, mount_target = mount.split(":")
+
+        if mount_source[0] == "{" and mount_source[-1] == "}":
+            # Resolve the environment variable for the mount source
+            mount_source = mount_source[1:-1]
+
+            if mount_source not in env_vars:
```
Kipok, line 349: We don't need to restrict it to what in the config here, right? Let's just check os.environ for that, and the env_vars field will be there only to pass env vars inside of a job

nemo_skills/pipeline/utils.py
```diff
@@ -266,6 +277,100 @@ def get_packager():
...
+                    f"Required environment variable {mount_source} not found in env variables passed in cluster configs."
+                )
+
+            mount_source = env_vars[mount_source]
+
+        if mount_target[0] == "{" and mount_target[-1] == "}":
+            # Resolve the environment variable for the mount target
+            mount_target = mount_target[1:-1]
+
+            if mount_target not in env_vars:
```
Kipok, line 360: same here

nemo_skills/pipeline/start_server.py
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import os
```
Kipok, line 14: unused?
```suggestion
```

Kipok (commented):
Just a few small changes more and good to merge

nemo_skills/pipeline/utils.py
titu1994: Hmm technically we dont need to restrict. But its very easy to miss the value. So ill instead check for the value from os.environ and raise an error if it doesn't exist there.

nemo_skills/pipeline/utils.py
titu1994: Same as above

Kipok (approved):
Thanks!

PR #138, opened 2024-09-30
add new math base

wedu-nvidia:
fix the math base prompt for better generation quality


nemo_skills/prompt/config/generic/math-base.yaml
```diff
@@ -0,0 +1,16 @@
+# default prompt for all math benchmarks (e.g. gsm8k, math)
```
Kipok, line 1:
```suggestion
# prompt that can be used with base models (or instruct models with base template)
```

Kipok (approved):

PR #163, opened 2024-10-04
Minor fix to the dependency job id calculation

titu1994:
Minor fix to the temporary function used to resolve the latest job directory - now it accepts explicit job directory demarcated by a integer string separated by _ after the `expname`. It then bypasses the `latest` calculation and just uses the job directory directly.

Further more, a minor fix in handler calculation - as when there are multiple jobs scheduled, the parameter is called `handlers` instead and represents a list (so we extend it instead)


Kipok (approved):

Kipok:
Thanks!

PR #175, opened 2024-10-08
Fill-in some of the missing pieces in repro docs

Kipok:


wedu-nvidia (approved):
Looks good to me, thanks for adding such detailed and informative content.

PR #177, opened 2024-10-08
Fixing the default top_k to -1

Kipok:


titu1994 (approved):
Thanks for catching this bug

PR #180, opened 2024-10-09
Nemo conversion for qwen from hf to nemo

wedu-nvidia:
this is my test code 
```
ns convert \
    --cluster="draco-oci" \
    --input_model=/hf_models/Qwen2.5-Math-7B \
    --output_model=/exps/qwen2.5-math-7b-nemo-covert \
    --convert_from=hf \
    --convert_to=nemo \
    --model_type=qwen \
    --num_gpus=8 \
    --hf_model_name=Qwen/Qwen2.5-Math-7B
```
I also tested the converted model and it works as expected.


nemo_skills/conversion/hf_to_nemo_qwen.py
```diff
@@ -0,0 +1,306 @@
...
+        required=False,
+        help="Name of HF model we are converting to (e.g. mistralai/Mistral-7B-v0.1)",
+    )
+    parser.add_argument("--override", action="store_true", help="Override existing output directory if it exists.")
+    args = parser.parse_args()
+    return args
+
+
+def load_config(args, qwen_config):
+    nemo_config = OmegaConf.load(os.path.join(os.path.dirname(__file__), 'qwen2_nemo_config.yaml')).model
```
Kipok, line 61:
```suggestion
    nemo_config = OmegaConf.load(os.path.join(os.path.dirname(__file__), 'nemo_config_qwen.yaml')).model
```

nemo_skills/conversion/qwen2_nemo_config.yaml
Kipok, line 1: let's name nemo_config_qwen.yaml for consistency

nemo_skills/pipeline/convert.py
```diff
@@ -165,9 +172,9 @@ def convert(
         pass
 
     # TODO: add support for qwen nemo conversion
-    if model_type == "qwen":
-        if convert_from == "nemo" or convert_to == "nemo":
-            raise ValueError("NeMo conversion for Qwen models is not supported yet")
+    # if model_type == "qwen":
```
Kipok, line 175: Let's keep the if, but only for the from part, since that's still not supported

nemo_skills/pipeline/convert.py
```diff
@@ -82,17 +82,24 @@ def get_hf_to_trtllm_cmd(
 def get_hf_to_nemo_cmd(
     input_model, output_model, model_type, hf_model_name, dtype, num_gpus, num_nodes, extra_arguments
 ):
+    # Check if the model_type is "nemo"
+    if model_type == "qwen":
+        conversion_command = "python -m nemo_skills.conversion.hf_to_nemo_qwen"
```
Kipok, line 87: Let's either change this to be with just `script = ht_to_nemo_qwen` same as for trtllm or change the other one to have this conversion_command style. Either way works, but it's best to keep things consistent

Kipok (changes requested):
Thanks Wei, just a couple of small changes and we can merge

nemo_skills/pipeline/convert.py
wedu-nvidia: ok, got it. I can add a comment then

nemo_skills/conversion/qwen2_nemo_config.yaml
wedu-nvidia: ok

wedu-nvidia:
fixed the above issues, can you please take a look again?

PR #182, opened 2024-10-10
SFT data processing commands 

shtoshni:
Added decontamination and SFT data preparation commands.


docs/reproducing-results.md
```diff
@@ -362,9 +362,59 @@ run inference through Nvidia NIM API.
...
 
-6. Now all the data is generated and you can follow up by converting it to the SFT format.
-
-   TBD
+6. Check for test set contamination.
+   We test against GSM8K, MATH, AMC 2023, and AIME 2024.  
+
+   First we use this command to get all the unique problems:
+   ```
+   cat /workspace/new-problems-solution-augmentation/*/*/output-rs0.jsonl | jq '.problem' | sort -u | shuf | jq -c -R '{"problem": .}' > /workspace/new-problems-solution-augmentation/unique_problems_all.jsonl
```
Kipok, line 370: This shouldn't be needed, I think - the retrieve_similar already builds a set, so it will only operate on unique problems

docs/reproducing-results.md
```diff
@@ -362,9 +362,59 @@ run inference through Nvidia NIM API.
...
+   ```
+   python -m nemo_skills.inference.retrieve_similar \
+      ++retrieve_from=/workspace/new-problems-solution-augmentation/unique_problems_all.jsonl \
+      ++compare_to="./nemo_skills/dataset/gsm8k/test.jsonl ./nemo_skills/dataset/math/test.jsonl ./nemo_skills/dataset/amc23/test.jsonl ./nemo_skills/dataset/aime24/test.jsonl" \
+      ++output_file=/workspace/new-problems-solution-augmentation/contamination-retrieved.jsonl \
+      ++top_k=5
+   ```
+   > **_NOTE:_** Currently the above command doesn't run inside docker, so you will need to install additional packages.
+
+   Next, you need to run LLM inference to check those closest found problems from the output file. Here is an example using Llama-405B from Nvidia API catalog, but you can replace it with OpenAI models or self-hosted models.
```
Kipok, line 383: Since this is about reproducing our results, I think we don't need to mention that so that it's clear that we used 405b. But we should put this inside decontamination doc

docs/reproducing-results.md
```diff
@@ -362,9 +362,59 @@ run inference through Nvidia NIM API.
...
+    ```
+    ns check_contamination \
+        --cluster=local \
+        --input_file=/workspace/new-problems-solution-augmentation/contamination-retrieved.jsonl \
+        --output_file=/workspace/new-problems-solution-augmentation/contamination-llm.jsonl \
+        --server_type=openai \
+        --model=meta/llama-3.1-405b-instruct \
+        --server_address=https://integrate.api.nvidia.com/v1 \
+        ++check_both_ways=True \
+        ++generation_key=contaminated;
```
Kipok, line 394:
```suggestion
```
That's the default

docs/reproducing-results.md
```diff
@@ -362,9 +362,59 @@ run inference through Nvidia NIM API.
...
+    ```
+    
+   Identify all the problems for which the `contaminated` key has the output True. 
+   Add the entry `"contaminated": True` in all the generation files in `/workspace/new-problems-solution-augmentation/`
```
Kipok, line 398: Let's provide a short script that does that?

docs/reproducing-results.md
```diff
@@ -362,9 +362,59 @@ run inference through Nvidia NIM API.
...
+
+7. Now all the data is generated and you can follow up by converting it to the SFT format.
+   We remove the problems marked as contaminated. 
+   We also remove solutions with length > 1024 Llama tokens.
+   To avoid the models from generating extremely short solutions, we remove solutions shorter than 200 characters.    
+   ```
+   python -m nemo_skills.training.prepare_sft_data \
+      ++prompt_template=llama3-instruct \
+      ++prompt_config=generic/math \
+      ++input_files="/workspace/solution-augmentation/*/output-rs*.jsonl /workspace/new-problems-solution-augmentation/*/*/output-rs*.jsonl" \
```
Kipok, line 408:
```suggestion
      ++input_files="/workspace/solution-augmentation/**/output-rs*.jsonl /workspace/new-problems-solution-augmentation/**/output-rs*.jsonl" \
```

nemo_skills/training/data_preparation_utils/prepare_sft_data.yaml
```diff
@@ -37,7 +41,9 @@ majority_filter:
 
 filters:
   drop_multi_boxed: true
+  remove_contamindated: false
```
Kipok, line 44: why false by default? I'd say should be true by default and we just make sure it doesn't do anything if the key isn't present

Kipok (changes requested):
Thanks, just a couple of small changes. It might also be a good idea to add a test for the small subset of openmathinstruct-2 sft conversion to make sure it's always supported in our code, like we did previously for openmathinstruct-1 here https://github.com/Kipok/NeMo-Skills/blob/c72baa85953315b3a310ad29fa8b0ce4a4e067c9/tests/test_data_preparation.py#L42

nemo_skills/training/data_preparation_utils/prepare_sft_data.yaml
shtoshni: Sure, making the default true.

docs/reproducing-results.md
shtoshni: Good point. Let me remove it.

docs/reproducing-results.md
shtoshni: Sure, let me remove it.

docs/reproducing-results.md
shtoshni: Cool!

docs/reproducing-results.md
shtoshni: Didn't know about **. Good one!

docs/reproducing-results.md
```diff
@@ -362,9 +362,53 @@ run inference through Nvidia NIM API.
...
-
-   TBD
+6. Check for test set contamination.
+   We test against GSM8K, MATH, AMC 2023, and AIME 2024.  
+ 
+   Retrieve top-5 similar items from the test sets
+   ```
+   python -m nemo_skills.inference.retrieve_similar \
+      ++retrieve_from="./nemo_skills/dataset/gsm8k/test.jsonl ./nemo_skills/dataset/math/test.jsonl ./nemo_skills/dataset/amc23/test.jsonl ./nemo_skills/dataset/aime24/test.jsonl" \
+      ++compare_from="/workspace/new-problems-solution-augmentation/**/output-rs0.jsonl" \
```
Kipok, line 372:
```suggestion
      ++compare_to="<path to /workspace>/new-problems-solution-augmentation/**/output-rs0.jsonl" \
```

docs/reproducing-results.md
```diff
@@ -362,9 +362,53 @@ run inference through Nvidia NIM API.
...
+      ++output_file=/workspace/new-problems-solution-augmentation/contamination-retrieved.jsonl \
```
Kipok, line 373:
```suggestion
      ++output_file=<path to /workspace>/new-problems-solution-augmentation/contamination-retrieved.jsonl \
```

nemo_skills/training/data_preparation_utils/filters.py
```diff
@@ -250,20 +288,23 @@ def process_dataset_entry(self, data_entry: str) -> str:
 
 
 class CodeTextFilter(BaseParallelProcessor):
-    def __init__(self, filter_type, solution_key='generation', **kwargs):
+    def __init__(self, filter_type, prompt_config, prompt_template, solution_key='generation', **kwargs):
```
Kipok, line 291: I guess it's better to only accept the template, since config is not needed here. And you can use the `load_config` from prompt/utils.py directly to load it

shtoshni:
Resolved all the comments. Let me know if there are any remaining issues.

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+      ++top_k=5
+   ```
+   > **_NOTE:_** Currently the above command doesn't run inside docker, so you will need to install additional packages.
+
+   Next, you need to run LLM inference to check those closest found problems from the output file. We use the Llama3.1-405B-Instruct model for this, and here's one way of doing it via Nvidia API catalog.
+
+    ```
+    ns check_contamination \
+        --cluster=local \
+        --input_file=<path to workspace>/new-problems-solution-augmentation/contamination-retrieved.jsonl \
```
Kipok, line 383:
```suggestion
        --input_file=/workspace/new-problems-solution-augmentation/contamination-retrieved.jsonl \
```
the reason we need <path to workspace> is that because retrieve-similar doesn't yet work as a normal pipeline script and so we need "unmounted" path there. But all other commands should still use /workspace

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+        --output_file=<path to workspace>/new-problems-solution-augmentation/contamination-llm.jsonl \
+        --server_type=openai \
+        --model=meta/llama-3.1-405b-instruct \
+        --server_address=https://integrate.api.nvidia.com/v1 \
+        ++check_both_ways=True;
```
Kipok, line 388:
```suggestion
        ++check_both_ways=True
```

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+   ```
+   > **_NOTE:_** Currently the above command doesn't run inside docker, so you will need to install additional packages.
+
+   Next, you need to run LLM inference to check those closest found problems from the output file. We use the Llama3.1-405B-Instruct model for this, and here's one way of doing it via Nvidia API catalog.
+
+    ```
+    ns check_contamination \
+        --cluster=local \
+        --input_file=<path to workspace>/new-problems-solution-augmentation/contamination-retrieved.jsonl \
+        --output_file=<path to workspace>/new-problems-solution-augmentation/contamination-llm.jsonl \
```
Kipok, line 384:
```suggestion
        --output_file=/workspace/new-problems-solution-augmentation/contamination-llm.jsonl \
```

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+      ++prompt_config=generic/math \
+      ++input_files="/workspace/solution-augmentation/**/output-rs*.jsonl /workspace/new-problems-solution-augmentation/**/output-rs*.jsonl" \
+      ++output_path=/workspace/sft_data.jsonl \
+      ++filters.remove_contamindated=true \
+      ++filters.remove_len_outlier_solutions=true \
+      ++use_chars_for_min_length=true \
+      ++min_solution_length=200 \
+      ++hf_model_name="meta-llama/Meta-Llama-3.1-8B" \
+      ++max_solution_length=1024 \
+      ++generation_suffix='"<|eot_id|>"';
```
Kipok, line 445:
```suggestion
      ++generation_suffix='"<|eot_id|>"'
```

nemo_skills/training/data_preparation_utils/filters.py
```diff
@@ -250,20 +288,21 @@ def process_dataset_entry(self, data_entry: str) -> str:
 
 
 class CodeTextFilter(BaseParallelProcessor):
-    def __init__(self, filter_type, solution_key='generation', **kwargs):
+    def __init__(self, filter_type, prompt_config, solution_key='generation', **kwargs):
```
Kipok, line 291: Let's call it template, since that's how we reference this "config" in the rest of the codebase and docs. The prompt_config is typically referred to the "config" part that contains user/system messages

Kipok, line 291: Alternatively, just accept the token right away without config file

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+   Identify all the problems for which the `contaminated` key has the output True. 
+   Add the entry `"contaminated": True` in all the generation files in `<path to workspace>/new-problems-solution-augmentation/`. Here is a sample python script for this:
+
+    ```python
+    def load_contaminated_problems(jsonl_file):
+        contaminated_problems = set()
+        with open(jsonl_file, 'r') as f:
+            for line in f:
+                data = json.loads(line)
+                if data.get('contaminated', False):
```
Kipok, line 400:
```suggestion
                if data['contaminated']:
```
Should always have this key as far as I understand. Better not have the default that might silently disable this filter if for some reason previous step didn't write these keys

docs/reproducing-results.md
```diff
@@ -362,9 +362,88 @@ run inference through Nvidia NIM API.
...
+
+    def update_output_files(directory, contaminated_problems):
+        file_pattern = str(Path(directory) / '**' / 'output-rs*.jsonl')
+        for file_path in glob.glob(file_pattern, recursive=True):
+            temp_file_path = Path(file_path).with_suffix('.temp')
+            
+            with open(file_path, 'r') as input_file, open(temp_file_path, 'w') as output_file:
+                for line in input_file:
+                    data = json.loads(line)
+                    if data.get('problem') in contaminated_problems:
```
Kipok, line 412:
```suggestion
                    if data['problem'] in contaminated_problems:
```

Kipok (changes requested):
sorry, a few more changes

docs/reproducing-results.md
shtoshni: Thanks for the detailed checks! Added the suggested changes.

docs/reproducing-results.md
shtoshni: Good point!

nemo_skills/training/data_preparation_utils/filters.py
shtoshni: Oh yeah, thanks for the suggestion.

shtoshni: Better to use the template in my opinion given we might be switching models.

shtoshni:
Let me know how it looks now.

nemo_skills/training/data_preparation_utils/filters.py
```diff
@@ -298,6 +337,8 @@ def process(self):
         self.prepare()
         os.makedirs(os.path.dirname(self.output_manifest_file), exist_ok=True)
         metrics = []
+        prompt = load_config(self.prompt_template)
```
Kipok, line 340:
```suggestion
        prompt = load_config(self.prompt_template, Path(__file__).absolute().parents[2] / 'prompt' / 'template')
```

PR #189, opened 2024-10-14
Updated code execution

i-vainn:
Added support for multiple code output formats.


nemo_skills/code_execution/utils.py
```diff
@@ -16,16 +16,27 @@
...
-        output += f"\n[stderr]\n{execution_dict['stderr']}\n[/stderr]"
+    if code_output_format == 'llama':
+        output = execution_dict["process_status"]
+        if execution_dict['stdout']:
+            output += f"\n[stdout]\n{execution_dict['stdout']}\n[/stdout]"
+        if execution_dict['stderr']:
+            output += f"\n[stderr]\n{execution_dict['stderr']}\n[/stderr]"
+        output = f"{code_output_begin}\n\n{output}{code_output_end}\n\n"
+    elif code_output_format == 'openmathinstruct':
+        output = f"\n{execution_dict['stdout']}\n" if execution_dict['stdout'] else f"\n{execution_dict['stderr']}\n"
```
Kipok, line 31: We should do concatenation of stdout and stderr here and below for qwen

Kipok: Actually let's remove openmathinstruct format, it's almost the same as qwen anyway, right?

nemo_skills/inference/server/code_execution_model.py
```diff
@@ -117,7 +118,7 @@ def generate(
                 # checking if any of the outputs need code execution and submitting requests in parallel
                 futures = [None] * len(prompts)
                 for idx, output in zip(remaining_ids, outputs):
-                    if output.strip().endswith(code_end):
+                    if output.endswith(code_end) and output.rfind(code_begin) > output.rfind(code_end, 0, -1):
```
Kipok, line 121: why do we need "0, -1"?

nemo_skills/prompt/config/qwen/math-code-fewshot.yaml
```diff
@@ -0,0 +1,27 @@
+few_shot_examples:
+  prefix: ""
+  template: |-
+    Problem: {problem}
+    {user_end}{assistant_begin}{solution}{assistant_end}{user_begin}
```
Kipok, line 5: Why do we need user/assistant tokens here? Better to not use them and simply show examples in the first user message

nemo_skills/prompt/config/qwen/math-code-fewshot.yaml
```diff
@@ -0,0 +1,27 @@
...
+  3. For Python code use the following template:
+    ```python
+    # Your Python code
+    ```
+  4. Put the final answer in \boxed{{}}.
+
+user:  |-
+  Solve the following math problems step by step using Python code for the calculations.
+  
+  {examples}
```
Kipok, line 26: {examples} might be empty, so it's better not to have a newline after to ensure consistent prompting

Kipok: You can use prefix/suffix to make it work with and without examples, see https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/prompt/config/generic/math.yaml#L3

nemo_skills/prompt/config/qwen/math-code-zeroshot.yaml
```diff
@@ -0,0 +1,19 @@
+system: |-
```
Kipok, line 1: Let's unify this and fewshot prompt

nemo_skills/prompt/template/llama3-base-fewshot.yaml
```diff
@@ -0,0 +1,31 @@
+# Prompt specification for Llama3-base model
```
Kipok, line 1: Why do we need this? This should be done as part of few shot examples in the config and not a separate template

nemo_skills/prompt/template/llama3-instruct-code.yaml
```diff
@@ -0,0 +1,32 @@
+# Prompt specification for the original Llama3-instruct model
```
Kipok, line 1: Let's call it `llama3-instruct-qwen-code`

nemo_skills/prompt/template/llama3-instruct-code.yaml
```diff
@@ -0,0 +1,32 @@
...
+user_begin: "<|start_header_id|>user<|end_header_id|>\n\n"
+user_end: "<|eot_id|>"
+
+assistant_begin: "<|start_header_id|>assistant<|end_header_id|>\n\n"
+assistant_end: "<|eot_id|>"
+
+stop_phrases: ["<|eot_id|>"]
+
+# used to execute code within these tags
+code_begin: "```python"
```
Kipok, line 25: Should we have a newline here as well?

nemo_skills/prompt/utils.py
```diff
@@ -96,6 +98,7 @@ class PromptTemplate:
     # used to extract the code output
     code_output_begin: str = '<llm-code-output>'
     code_output_end: str = '</llm-code-output>'
+    code_output_format: str = 'llama'
```
Kipok, line 101:
```suggestion
    code_output_format: str = 'qwen'
```

nemo_skills/prompt/utils.py
```diff
@@ -121,16 +124,28 @@ def build_filled_example(self, example_dict: Dict[str, Any]) -> str:
...
             example_dict["solution"] = example_dict["solution"].replace(
                 "{code_begin}", self.config.template.code_begin
             )
-            example_dict["solution"] = example_dict["solution"].replace("{code_end}", self.config.template.code_end)
-            example_dict["solution"] = example_dict["solution"].replace(
-                "{code_output_begin}", self.config.template.code_output_begin
-            )
             example_dict["solution"] = example_dict["solution"].replace(
-                "{code_output_end}", self.config.template.code_output_end
+                "{code_end}", self.config.template.code_end.strip()
```
Kipok, line 145: Why strip()? The newline might be important

Kipok (changes requested):
Thanks! A couple of changes are needed. Please also add tests here https://github.com/Kipok/NeMo-Skills/blob/main/tests/test_code_execution.py with different formats

nemo_skills/inference/server/code_execution_model.py
i-vainn: Sometimes ``` can be used for purposes other than code blocks, such as in regular Markdown or other syntax. The code snippet `output.rfind(code_begin) > output.rfind(code_end, 0, -1)` checks whether the last `code_end` actually corresponds to the previous `code_begin`. If there's another `code_end` between the `code_begin` and the last `code_end`, this condition will return false.

nemo_skills/code_execution/utils.py
```diff
@@ -16,16 +16,28 @@
...
+        output = execution_dict["process_status"]
+        if execution_dict['stdout']:
+            output += f"\n[stdout]\n{execution_dict['stdout']}\n[/stdout]"
+        if execution_dict['stderr']:
+            output += f"\n[stderr]\n{execution_dict['stderr']}\n[/stderr]"
+        output = f"{code_output_begin}\n\n{output}{code_output_end}\n\n"
+    elif code_output_format == 'qwen':
+        output = ""
+        if execution_dict['stdout']:
+            output += f"[stdout]\n{execution_dict['stdout']}\n[/stdout]"
```
Kipok, line 33: This shouldn't have any [stdout] tags, right?

nemo_skills/inference/server/code_execution_model.py
Kipok: I still don't understand why we need 0, -1. Why do we use different code to check for the last code_begin vs the last code_end?

nemo_skills/prompt/config/qwen/math-code-nostepbystep.yaml
```diff
@@ -0,0 +1,25 @@
+few_shot_examples:
+  prefix: "Here are some examples of problems and solutions you can refer to.\n\n"
+  template: "Problem:\n{problem}\n\nSolution:\n{solution}\n\n\n\n\n\n"
+  suffix: "Problem:\n"
+  examples_type: math_text_with_code
+
+system: |-
+  You are a math problem solver that uses Python code as an integral part of your reasoning. Show your work clearly and concisely.
```
Kipok, line 8: Let's instead try to directly use the default prompt from here https://huggingface.co/Qwen/Qwen2.5-Math-72B-Instruct

Kipok: Few-shot examples shouldn't be necessary as well, since the model should know how to solve the problems

nemo_skills/prompt/config/qwen/math-code.yaml
```diff
@@ -0,0 +1,25 @@
+few_shot_examples:
```
Kipok, line 1: Is this the same as the other prompt?

Kipok (changes requested):

nemo_skills/prompt/config/qwen/math-code.yaml
i-vainn: Yeah

nemo_skills/code_execution/utils.py
```diff
@@ -16,16 +16,28 @@
...
+            output += f"\n[stdout]\n{execution_dict['stdout']}\n[/stdout]"
+        if execution_dict['stderr']:
+            output += f"\n[stderr]\n{execution_dict['stderr']}\n[/stderr]"
+        output = f"{code_output_begin}\n\n{output}{code_output_end}\n\n"
+    elif code_output_format == 'qwen':
+        output = ""
+        if execution_dict['stdout']:
+            output += f"{execution_dict['stdout']}"
+        if execution_dict['stderr']:
+            output += f"{execution_dict['stderr']}"
```
Kipok, line 35: Do we need to be adding a newline if both stdout and stderr are present?

nemo_skills/inference/server/code_execution_model.py
```diff
@@ -117,7 +118,9 @@ def generate(
                 # checking if any of the outputs need code execution and submitting requests in parallel
                 futures = [None] * len(prompts)
                 for idx, output in zip(remaining_ids, outputs):
-                    if output.strip().endswith(code_end):
+                    # .rfind(code_end, 0, -1) searches for the previous occurrence of code_end and checks
```
Kipok, line 121:
```suggestion
                    # .rfind(code_end, 0, -1) searches for the second-to-last occurrence of code_end and checks
```

nemo_skills/inference/server/code_execution_model.py
```diff
@@ -126,7 +129,9 @@ def generate(
                             session_id=new_outputs[idx]['session_id'],
                         )
                 for idx, output in zip(remaining_ids, outputs):
-                    if output.strip().endswith(code_end):
+                    # .rfind(code_end, 0, -1) searches for the previous occurrence of code_end and checks
```
Kipok, line 132:
```suggestion
                    # .rfind(code_end, 0, -1) searches for the second-to-last occurrence of code_end and checks
```

nemo_skills/prompt/template/llama3-instruct-qwen-code.yaml
```diff
@@ -0,0 +1,32 @@
...
+stop_phrases: ["<|eot_id|>"]
+
+# used to execute code within these tags
+code_begin: "```python\n"
+code_end: "```\n"
+# used to extract the code output
+code_output_begin: "```output\n"
+code_output_end: "```\n"
+# used to post-process code output
+# choices: openmathinstruct, llama, qwen
```
Kipok, line 31:
```suggestion
# choices: llama, qwen
```

tests/test_code_execution.py
```diff
@@ -194,7 +194,7 @@ def test_real_generations(sandbox_type):
 @pytest.mark.parametrize(
     "code_begin,code_end,code_output_begin,code_output_end",
     [
-        # ('<llm-code>', '</llm-code>', '<llm-code-output>', '</llm-code-output>'),
+        # ('<llm-code>\n', '</llm-code>\n', '<llm-code-output>\n', '</llm-code-output>\n'),
```
Kipok, line 197: Can we enable this test?

tests/test_code_execution.py
```diff
@@ -223,10 +223,10 @@ def test_few_shots(sandbox_type, code_begin, code_end, code_output_begin, code_o
                 session_id = None
                 for code_snippet, expected_output in zip(code_snippets, expected_outputs):
                     output, session_id = sandbox.execute_code(code_snippet, session_id=session_id)
+                    execution_dict = {"process_status": "completed", "stdout": expected_output.strip(), "stderr": None}
```
Kipok, line 226: Do we have to do .strip()?

tests/test_code_execution.py
Kipok: And I guess add one more for qwen format here and also in the explicit few-shot test below?

Kipok (commented):
Thanks, just a couple of small changes and should be good to merge!

Kipok:
Can you please also for the future create PRs from the branch and not fork? That way we'd be able to launch the GPU tests CI

tests/test_code_execution.py
i-vainn: Yes, because `extract_code_output` returns everything between `code_output_begin` and `code_output_end` including `\n`'s while `sandbox.execute_code` returns just code output without any `\n`'s

tests/test_code_execution.py
Kipok: Right, but it also does extraction of code from those tags. E.g. if you use some special symbol in the tokens that the regex doesn't parse correctly or there is some other bug, it's good if we can catch it here

nemo_skills/code_execution/utils.py
```diff
@@ -16,16 +16,28 @@
...
+        output = execution_dict["process_status"]
+        if execution_dict['stdout']:
+            output += f"\n[stdout]\n{execution_dict['stdout']}\n[/stdout]"
+        if execution_dict['stderr']:
+            output += f"\n[stderr]\n{execution_dict['stderr']}\n[/stderr]"
+        output = f"{code_output_begin}\n\n{output}{code_output_end}\n\n"
+    elif code_output_format == 'qwen':
+        output = ""
+        if execution_dict['stdout']:
+            output += f"{execution_dict['stdout']}\n"
```
Kipok, line 33: I don't think we should always be adding a new line, right? It's only needed if both stdout and stderr are present, but otherwise we just take whatever is returned. 

It's probably a minor detail, but it's good to get this as consistent as possible, since LLMs might get off-track if there is an extra newline somewhere where it's not supposed to be

Kipok: I guess an alternative is that we don't include newline and actually raise an assert if both stderr and stdout are non-empty. I think it should never be the case with our current sandbox and if it is, then we don't really know what to do, since stdout and stderr might have been interleaved in the output, but we extracted them separately

tests/test_prompts.py
```diff
@@ -282,12 +282,191 @@ def test_generic_math_prompt_code_examples():
 # we will solve for b and c
 solutions = sp.solve((point_1,point_2), (b, c))
 solutions[c]
-</llm-code><llm-code-output>
+</llm-code>
```
Kipok, line 285: Can you please add one more test for qwen format? Good to have them explicit somewhere so that we can use tests as a quick reference of how the code tokens/output looks like

nemo_skills/code_execution/utils.py
i-vainn: > I don't think we should always be adding a new line, right? It's only needed if both stdout and stderr are present, but otherwise we just take whatever is returned.

Right now the output will look like
\`\`\`output
number
\`\`\`
if stdout equals to number. 

And if both stdout and stderr are empty we return 
\`\`\`output
\`\`\`
without any content.

Isn't it what we expect?

i-vainn: >  I think it should never be the case with our current sandbox

What abut the code printing something before an error? Isn't this possible with our current sandbox?

Kipok: yes, but the issue is that if you have a bunch of prints to stderr and then to stdout and they are mixed. It's unclear how they should be treated. But it's not really a problem right now because for some reason (that's actually a bug but I don't know how to fix it) even errors go to stdout. E.g. you can see that in execution tests we always assert that stderr is empty

Kipok: But the newlines are already in the output, right? E.g. if you do print(1) it will be "1\n" and so your current change would make it "1\n\n"?

i-vainn: Right, this is a good point

Kipok (approved):
thanks!

PR #190, opened 2024-10-14
Adding support for dependent jobs to generate.py

Kipok:


Kipok:
Depends on https://github.com/NVIDIA/NeMo-Run/pull/77
@titu1994 FYI

PR #194, opened 2024-10-17
Nemo -> hf conversion for qwen models

Kipok:


wedu-nvidia (approved):
All look good to me. Thanks

PR #196, opened 2024-10-22
nemo with oai api

arendu:


Kipok:
An equivalent functionality was merged in https://github.com/Kipok/NeMo-Skills/pull/224, so let's close this one.

arendu:
One part of this deals with mamba serving... ill make a new PR for that.

PR #198, opened 2024-10-22
Adding antlr version check

shtoshni:


nemo_skills/code_execution/math_grader.py
```diff
@@ -87,6 +87,15 @@
 from math import isclose
 from typing import Union
 
+from pkg_resources import DistributionNotFound, VersionConflict, require
```
Kipok, line 90: If it's easy, can we switch to the importlib.resources api? Seems like pkg_resources is deprecated https://setuptools.pypa.io/en/latest/pkg_resources.html.

Also, let's actually fail with exception, so that it's not just warning to the user, but we stop execution if the version doesn't match. Otherwise it's too easy to miss this.

Finally, just to be safe, let's put this check inside `math_equal` function as we might accidentally import this file somewhere where we don't actually call math_equal and get the failure

Kipok (changes requested):

nemo_skills/code_execution/math_grader.py
shtoshni: Fair enough!

shtoshni: Created a function to not overcomplicate the `math_equal` function

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,9 +84,33 @@
 import contextlib
 import re
 import signal
+from importlib.metadata import PackageNotFoundError, version
 from math import isclose
 from typing import Union
 
+from rich import print as rprint
```
Kipok, line 91: Let's not use rich as it's not available inside the sandbox where we actually run this code. While we might be able to modify our own setup to include it, we don't always have control over the sandbox environment if using cloud solutions e.g.

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,9 +84,33 @@
...
+    try:
+        installed_version = version(PACKAGE_NAME)
+        if installed_version != REQUIRED_VERSION:
+            rprint(
+                f"[bold red]Package version mismatch: {installed_version} (required: {REQUIRED_VERSION})[/bold red]"
+            )
+            sys.exit(1)
+    except PackageNotFoundError:
+        rprint(f"[bold red]Package {PACKAGE_NAME} not found[/bold red]")
+        sys.exit(1)
```
Kipok, line 109:
```suggestion
        raise
```
Better to reraise the exception after printing the instruction on what to do, since it will give people an exact traceback of where the error is coming from

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,9 +84,33 @@
...
+    except Exception as e:
+        rprint(f"[bold red]Error checking version: {e}[/bold red]")
+        sys.exit(1)
```
Kipok, line 112:
```suggestion
        raise
```

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,9 +84,33 @@
...
+
+    try:
+        installed_version = version(PACKAGE_NAME)
+        if installed_version != REQUIRED_VERSION:
+            rprint(
+                f"[bold red]Package version mismatch: {installed_version} (required: {REQUIRED_VERSION})[/bold red]"
+            )
+            sys.exit(1)
+    except PackageNotFoundError:
+        rprint(f"[bold red]Package {PACKAGE_NAME} not found[/bold red]")
```
Kipok, line 108: Can you also include the version that we need here, so that it's easier for people to understand what to do? Something like `please install antlr4-python3-runtime==4.11.0`

nemo_skills/code_execution/math_grader.py
Kipok: I guess we don't need this blank exception at all? If something fails other than our check, that's a bug and we'd rather see the full stack trace

Kipok (changes requested):

nemo_skills/code_execution/math_grader.py
shtoshni: Sure. Added this print statement.

nemo_skills/code_execution/math_grader.py
shtoshni: Yeah not sure if that would be triggered at all.

nemo_skills/code_execution/math_grader.py
shtoshni: That's a good point. Replaced print + sys.exit with a raise command.

nemo_skills/code_execution/math_grader.py
shtoshni: That's a good point.

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,10 +84,25 @@
...
+def _check_antlr_version():
+    "Function for checking the antlr package version."
+    # Check antlr version
+    PACKAGE_NAME = 'antlr4-python3-runtime'
+    REQUIRED_VERSION = '4.11.0'
+
+    try:
+        installed_version = version(PACKAGE_NAME)
+        if installed_version != REQUIRED_VERSION:
+            raise Exception(f"Package version mismatch: {installed_version} (required: {REQUIRED_VERSION})")
```
Kipok, line 101:
```suggestion
            raise RuntimeError(f"Package {PACKAGE_NAME} version mismatch: {installed_version} (required: {REQUIRED_VERSION})")
```

nemo_skills/code_execution/math_grader.py
```diff
@@ -84,10 +84,25 @@
...
+    except PackageNotFoundError:
+        raise Exception(f"Package {PACKAGE_NAME} not found. Please install antlr4-python3-runtime==4.11.0.")
```
Kipok, line 103:
```suggestion
        raise RuntimeError(f"Package {PACKAGE_NAME} not found. Please install antlr4-python3-runtime==4.11.0.")
```

Kipok (approved):
Just a few small changes for better error reporting and let's merge after that

shtoshni:
Thanks for the thorough review, @Kipok!

PR #200, opened 2024-10-23
Update train pipeline to new DAG api from NeMo Run

titu1994:
# Changelog
- Updates sequential experiment run with new DAG API

### DIFF
--- a/nemo_skills/pipeline/train.py
+++ b/nemo_skills/pipeline/train.py
@@ -256,7 +256,7 @@ def train(
             num_tasks=1,
             num_gpus=num_gpus,
             run_after=run_after,
-            task_dependencies=[prev_job_id]
+            task_dependencies=[prev_job_id],
         )
 
         run_exp(exp, cluster_config)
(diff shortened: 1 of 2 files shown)


nemo_skills/pipeline/train.py
```diff
@@ -219,8 +219,9 @@ def train(
     )
 
     with run.Experiment(expname) as exp:
+        prev_job_id = None
```
Kipok, line 222: It's not really id, but a task object, so better to rename, but not too important

nemo_skills/pipeline/train.py
```diff
@@ -254,9 +256,10 @@ def train(
             num_tasks=1,
             num_gpus=num_gpus,
             run_after=run_after,
+            task_dependencies=[prev_job_id],
```
Kipok, line 259:
```suggestion
            task_dependencies=[prev_job_id] if prev_job_id is not None else None,
```
We sometimes run this script with 0 training jobs to just re-average checkpoints, so this might actually be the first job

Kipok (changes requested):
Thanks, just a small change

nemo_skills/pipeline/train.py
titu1994: Ah ok will do

titu1994: Fixed

PR #201, opened 2024-10-23
fix checkpoints averaging

dgtm777:
List type is not working when launching through the command line. Changed to the string type

### DIFF
--- a/docs/openmathinstruct2/training.md
+++ b/docs/openmathinstruct2/training.md
@@ -81,7 +81,7 @@ ns train \
     --nemo_model=/workspace/llama3.1-8b-nemo \
     --num_nodes=8 \
     --num_gpus=8 \
-    --average_steps 10000 20000 30000 40000 50000 60000 \
+    --average_steps=10000,20000,30000,40000,50000,60000 \
     --training_data=/workspace/openmathinstruct2-sft.jsonl \
     ++model.data.train_ds.micro_batch_size=4 \
     ++model.tensor_model_parallel_size=4 \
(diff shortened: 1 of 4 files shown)


docs/openmathinstruct2/training.md
```diff
@@ -81,7 +81,7 @@ ns train \
     --nemo_model=/workspace/llama3.1-8b-nemo \
```
Kipok, line 81: Please also update https://github.com/Kipok/NeMo-Skills/blob/main/docs/pipelines/training.md

nemo_skills/pipeline/train.py
```diff
@@ -160,7 +160,9 @@ def train(
     disable_wandb: bool = typer.Option(False, help="Disable wandb logging"),
     with_sandbox: bool = typer.Option(False, help="If sandbox is required for code generation"),
     partition: str = typer.Option(None, help="Specify partition for jobs"),
-    average_steps: list[int] = typer.Option(None, help="List of checkpoint steps to average"),
+    average_steps: str = typer.Option(
```
Kipok, line 163: Can you please add a check that if there is a " " inside average_steps we will raise an error and ask to check if it's formatted properly with the comma? Not really needed long-term, but to make it easier for people to transition from the previous way (which did work through python api), so that they don't just get their training jobs failing but instead see an error at the submission time

And if you can do the same for benchmarks to cover https://github.com/Kipok/NeMo-Skills/issues/153, that would be great

Kipok (changes requested):

Kipok (approved):
Thanks!

PR #207, opened 2024-10-24
feat: Add Reward Model Training

gwarmstrong:
Augments the training pipeline to support Reward Model Training with NeMo-Aligner


nemo_skills/pipeline/train.py
```diff
@@ -31,6 +32,135 @@
...
+    extra_arguments: str = ""
+    logging_params: str = ""
+
+
+def train_sft(params: TrainingParams) -> str:
+    if params.config_name is None:
+        config_name = "sft_config"
+    else:
+        config_name = params.config
+    config_params = f"--config-name={config_name} --config-path={params.config_path} "
```
Kipok, line 62: Since we are not sharing the main cmd, why not include these 3 arguments directly in the command below instead of creating them here and then formatting inside cmd?

nemo_skills/pipeline/train.py
```diff
@@ -31,6 +32,135 @@
...
+        f"    {extra_arguments} "
+    )
+    return cmd
+
+
+def train_rm(params: TrainingParams) -> str:
+    if params.config_name is None:
+        config_name = "dpo_config"
+    config_params = f""
+    training_script = "python -u /opt/NeMo-Aligner/examples/nlp/gpt/train_reward_model.py"
```
Kipok, line 130: I think it's better to keep the entry script on our side (by copy-pasting from nemo aligner). While your approach also works in the majority of cases, we don't want to put any restriction on what kind of container people can use our repo with. So while this is unlikely, someone can in principle have aligner installed in a different place or maybe they'd want to mount things inside container and then point python path there

Kipok: And same for the config, which also gives people option to provide their own (which otherwise isn't possible, since it needs to be the first argument, and so we cannot use our default way of passing parameters through the pipeline wrapper)

tests/gpu-tests/test_train.py
```diff
@@ -140,3 +140,42 @@ def test_dpo():
...
     assert metrics['num_entries'] == 10
+
+
+@pytest.mark.gpu
+def test_rm():
+    model_path = os.getenv('NEMO_SKILLS_TEST_NEMO_MODEL')
+    if not model_path:
+        pytest.skip("Define NEMO_SKILLS_TEST_NEMO_MODEL to run this test")
+
+    model_path = os.getenv('NEMO_SKILLS_TEST_NEMO_MODEL')
```
Kipok, line 151: Duplicate lines?

tests/gpu-tests/test_train.py
```diff
@@ -140,3 +140,42 @@ def test_dpo():
...
+        expname="test-rm",
+        training_algo="rm",
+        output_dir="/tmp/nemo-skills-tests/test-rm",
+        nemo_model=model_path,
+        num_nodes=1,
+        num_gpus=1,
+        num_training_jobs=1,
+        training_data="/nemo_run/code/tests/data/small-rm-data.test",
+        disable_wandb=True,
+    )
```
Kipok, line 181: Can you add some assert that checkpoints are created (which we can later replace with an inference call when that's integrated)? The problem is that nemo-run unfortunately always returns 0 exit code from any commands even if they crashed and so this train function will never cause the test to fail

nemo_skills/pipeline/train.py
```diff
@@ -31,6 +32,135 @@
...
+        f"    exp_manager.exp_dir={params.output_dir}/training "
+        f"    ++exp_manager.max_time_per_run={params.timeout} "
+        f"    {extra_arguments} "
+    )
+    return cmd
+
+
+def train_rm(params: TrainingParams) -> str:
+    if params.config_name is None:
+        config_name = "dpo_config"
```
Kipok, line 128: I guess this is currently unused, but see another comment that we should add the config from nemo aligner here ideally

nemo_skills/pipeline/train.py
Kipok: Alternatively, if that cmd is fully shared across all 3 cases I think it's ok to move it inside a separate function and reuse that. It's less explicit but still close enough in the code that hopefully won't confuse anyone. Whichever you think makes more sense

gwarmstrong: Fair enough--yeah I refactored out the shared part of the script and threw all of the customization into the top layer command.

nemo_skills/pipeline/train.py
gwarmstrong: Makes sense. Done :)

tests/gpu-tests/test_train.py
gwarmstrong: Removed.

nemo_skills/pipeline/train.py
gwarmstrong: fixed in the most recent refactor of this.

nemo_skills/training/training_rm.yaml
```diff
@@ -0,0 +1,119 @@
...
+    overlap_grad_sync: False
+    contiguous_grad_buffer: True
+    lr: 9e-6
+    weight_decay: 0.1
+    betas:
+    - 0.9
+    - 0.98
+    sched:
+      name: CosineAnnealing
+      warmup_steps: 10
```
Kipok, line 98:
```suggestion
      warmup_steps: 0
```

nemo_skills/training/training_rm.yaml
```diff
@@ -0,0 +1,119 @@
...
+      constant_steps: 1000
```
Kipok, line 99:
```suggestion
      constant_steps: 0
```

nemo_skills/training/training_rm.yaml
```diff
@@ -0,0 +1,119 @@
...
+      min_lr: 9e-7
```
Kipok, line 100:
```suggestion
      min_lr: ${model.optim.lr} # No interpolation happening if the min_lr is not set to be different from lr
```

nemo_skills/training/training_rm.yaml
```diff
@@ -0,0 +1,119 @@
...
+    model_parallel_size: ${multiply:${model.tensor_model_parallel_size}, ${model.pipeline_model_parallel_size}}
+
+pretrained_checkpoint:
+  restore_from_path: null
+
+model:
+  mcore_gpt: True
+  share_embeddings_and_output_weights: False
+
+  reward_model_type: binary_ranking # ["binary_ranking, "regression"]
```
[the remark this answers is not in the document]
gwarmstrong: 👍

nemo_skills/pipeline/train.py
```diff
@@ -84,6 +137,37 @@ def get_training_cmd(
...
+        num_nodes=num_nodes,
+        expname=expname,
+        training_algo=training_algo,
+        disable_wandb=disable_wandb,
+        wandb_project=wandb_project,
+        timeout=timeout,
+        extra_arguments=extra_arguments,
+        logging_params=logging_params,
+    )
+    extra_arguments = get_extra_arguments[training_algo](training_params)
```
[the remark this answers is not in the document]
gwarmstrong: I moved it into the post_init

Kipok (approved):
Thanks!

gwarmstrong:
@Kipok the gpu test failure appears to have come from an unrelated test--any idea what that's about?

Kipok:
The tests are failing because of an issue introduced in another PR. Let me run the training tests locally to double check there are no issues there and we can merge after that

tests/gpu-tests/test_train.py
```diff
@@ -140,3 +140,40 @@ def test_dpo():
...
+
+    assert os.path.exists("/tmp/nemo-skills-tests/test-rm/training/checkpoints/model-averaged")
```
Kipok, line 179:
```suggestion
    assert os.path.exists("/tmp/nemo-skills-tests/test-rm/model-averaged-nemo")
```
We move the averaged model to a different location

PR #220, opened 2024-10-30
Length based filtering for questions

shtoshni:
Added support for length-based filtering of questions.

### DIFF
--- a/nemo_skills/training/data_preparation_utils/filters.py
+++ b/nemo_skills/training/data_preparation_utils/filters.py
@@ -148,44 +148,12 @@ class RemoveContaminated(BaseFilter):
         return [DataEntry(data=data_entry, metrics=dict(num_removed=0))]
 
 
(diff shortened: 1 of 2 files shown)


nemo_skills/training/data_preparation_utils/filters.py
```diff
@@ -148,6 +148,38 @@ def process_dataset_entry(self, data_entry) -> List:
         return [DataEntry(data=data_entry, metrics=dict(num_removed=0))]
 
 
+class RemoveLenOutlierProblems(BaseFilter):
```
Kipok, line 151: Isn't it the same as `RemoveLenOutlierSolutions`? If so, let's just name it more generally and then use as it is by changing the `problem_key` (which we should also name more generally)

shtoshni: Replaced these with a general function

Kipok (approved):
Thanks!

PR #221, opened 2024-10-30
Minor refactor to accomodate externalization of eval and inference

titu1994:


nemo_skills/prompt/config/generic/codegen_system.yaml
```diff
@@ -0,0 +1,13 @@
+# Default prompt that does not contain any extra instructions. Can be used for instruction following evals (e.g. ifeval)
```
Kipok, line 1: Comment is not up-to-date. Also should we replace the original prompt with this one instead of keeping both?

titu1994: Hmm actually its not working yet, so i don't think its worth upstreaming just yet.

nemo_skills/evaluation/evaluator.py
```diff
@@ -297,5 +298,20 @@ def eval_lean4(cfg):
...
+    if is_evaluator_registered(eval_type):
+        raise ValueError(f"Evaluator for {eval_type} already registered")
+
+    EVALUATOR_MAP[eval_type] = eval_fn
+
+
 def evaluate(cfg):
+    if cfg.eval_type not in EVALUATOR_MAP:
+        raise ValueError(
+            f"Evaluator not found for type: {cfg.eval_type}.\n" f"Supported types: {str(EVALUATOR_MAP.keys())}"
```
Kipok, line 315:
```suggestion
            f"Evaluator not found for type: {cfg.eval_type}.\nSupported types: {str(EVALUATOR_MAP.keys())}"
```

Kipok (approved):
Thanks!

PR #223, opened 2024-10-31
add setting for large-scale data training

wedu-nvidia:
add setting for large-scale data training 
```
++model.data.train_ds.hf_dataset=True \
++model.data.train_ds.index_mapping_dir=/data/your_data_cache/
```

### DIFF
--- a/docs/pipelines/training.md
+++ b/docs/pipelines/training.md
@@ -72,6 +72,17 @@ defaults for an 8B model size
     ++model.tensor_model_parallel_size=4
 ```
 
+
+
+To enhance training efficiency for large-scale data, we strongly recommend adding a cache folder to store the dataset indexing cache, which will speed up the training process. Additionally, set hf_dataset=True to prevent CPU out-of-memory (OOM) issues. This is already the default setting in sft_config.yaml. Use the following configuration:
+
+```bash
+    ++model.data.train_ds.hf_dataset=True \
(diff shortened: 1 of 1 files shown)


docs/pipelines/training.md
```diff
@@ -72,6 +72,17 @@ defaults for an 8B model size
     ++model.tensor_model_parallel_size=4
 ```
 
+
+
+To enhance training efficiency for large-scale data, we strongly recommend adding a cache folder to store the dataset indexing cache, which will speed up the training process. Additionally, set hf_dataset=True to prevent CPU out-of-memory (OOM) issues. This is already the default setting in sft_config.yaml. Use the following configuration:
+
+```bash
+    ++model.data.train_ds.hf_dataset=True \
+    ++model.data.train_ds.index_mapping_dir=/data/your_data_cache/
```
Kipok, line 81: Can we do it automatically? We already ask users for training data explicitly, so can we just use something like training_data.parent / 'cache' as the default cache path?

wedu-nvidia: Yes, we can. You suggest I directly modify the setting in the code script and set it as default and user do not have to worry about it? If that is the case, I can modify the training script then.

Kipok: Yes, let's do that. If someone wants to disable this, they can always override explicitly

PR #227, opened 2024-11-01
Added scipy and proper indentation

i-vainn:


requirements/code_execution.txt
```diff
@@ -19,6 +19,7 @@ ipython
 # this are requirements for code execution - put all packages that LLM should have access to in here
 numpy
 pandas
+scipy
```
Kipok, line 22: let's maybe also add scikit-learn as it might sometimes also be used I imagine?

Kipok:
Also looks like our current sandbox dockerfile fails because of some lean error. Can you please work with @Alcray to fix that (by pinning the version wherever necessary or any other fix that works)?

requirements/code_execution.txt
i-vainn: I am not sure if we need it. Qwen with TIR prompt is not using it. Maybe we'll need it later, but for now its kinda useless

PR #229, opened 2024-11-01
Updating docs with length-based question filtering updates

shtoshni:
Recently in a (github issue)[https://github.com/Kipok/NeMo-Skills/issues/214], a discussion about length-based filtering for questions came up. We had originally not done any such filtering for OpenMathInstruct-2, but found that doing such filtering doesn't affect the performance, and allows for a higher micro batch size. This PR updates the docs with this finding.


wedu-nvidia:
Looks good to me.

wedu-nvidia (approved):

PR #231, opened 2024-11-07
Avoid the streaming compression route if the command  is not available

shtoshni:
The `pv` utility may not be available on certain systems. This patch checks for whether the utility is available or not, and avoids the streaming compression version when it is not available.

### DIFF
--- a/nemo_skills/pipeline/utils.py
+++ b/nemo_skills/pipeline/utils.py
@@ -296,8 +296,16 @@ def cluster_download(tunnel: SSHTunnel, remote_dir: str, local_dir: str, remote_
     result = tunnel.run(f'du -sb {remote_dir} | cut -f1')
     total_size = int(result.stdout.strip())
     
-    # Certain systems may not have the `pv` command
-    streaming_possible = (shutil.which("pv") is not None)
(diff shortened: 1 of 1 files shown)


nemo_skills/pipeline/utils.py
```diff
@@ -294,16 +295,23 @@ def cluster_download(tunnel: SSHTunnel, remote_dir: str, local_dir: str, remote_
...
-        f'cd {remote_dir_parent} && '
-        f'tar -cf - {remote_dir_name} | '
-        f'pv -s {total_size} -p -t -e -b -F "Compressing Remote Directory: %b %t %p" | '
-        f'gzip > {remote_tar}'
-    )
 
-    # Run the remote compression command and stream the progress
-    result = tunnel.run(command, watchers=[OutputWatcher()], pty=True, hide=False)
+    # Certain systems may not have the `pv` command
+    streaming_possible = shutil.which("pv") is not None
```
titu1994, line 300: It's the cluster that didn't have pv, not the local machine. This check is just for local right ?

titu1994 (approved):
LGTM thanks !

PR #236, opened 2024-11-12
Remove RLIMIT_STACK in sandbox

darraghdog:
### Reason for Removing `resource.setrlimit(resource.RLIMIT_STACK)`

**Problem Description**: 
The execution environment is encountering a `ValueError` with the message `not allowed to raise maximum limit` when attempting to set the stack size limit using `resource.setrlimit(resource.RLIMIT_STACK, (limit, limit))`. This error occurs because the Python process lacks the necessary permissions to increase its stack limit beyond the system-imposed maximum.

**Solution**:
- By removing this line, we avoid unnecessary and potentially problematic attempts to modify the stack size limit. The system default or the existing limit will remain in place, which should suffice for the safe execution of the intended operations.

**Impact of Change**:
- Removing this line has shown to resolve the error without observable negative impact on memory management or program stability in the environments where this issue was encountered.
- It maintains the intention of memory constraints through the other limits (`RLIMIT_AS` and `RLIMIT_DATA`) which do not trigger permissions issues.

This change should improve compatibility across different environments, especially where permission to raise stack limits is restricted.


Kipok (approved):
Thanks!

PR #240, opened 2024-11-13
Add generic script processing

gwarmstrong:
Enables calling arbtirary modules as scripts  through NeMo-Skills, e.g.,

```bash
ns run_cmd --module nemo_skills.training.prepare_sft_data \
    --cluster=local \
    ++input_files=...
```

### DIFF
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -27,8 +27,7 @@ LOG = logging.getLogger(__file__)
 
 
 def get_cmd(module, extra_arguments):
-    cmd = "pip install transformers && "
-    cmd += f"time python -m {module} "
+    cmd = f"python -m {module} "
     cmd += f" {extra_arguments} "
     cmd = (
         f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && "
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import os
 from pathlib import Path
 
 import nemo_run as run
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -67,9 +67,9 @@ def script(
     script: str = typer.Option(
         None,
         help=(
-            "Path to the python script to run. "
-            "If specified, will run this script instead of the module. "
-            "Will be passed as the first argument to the module."
+            "Path to the python script to run. The script must in the git index "
+            "of the directory where this command is run. The relative path to the script "
+            "from the current working directory should be provided."
         ),
     ),
     num_gpus: int = typer.Option(1, help="Number of GPUs to use"),
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -26,15 +26,19 @@ from nemo_skills.utils import setup_logging
 LOG = logging.getLogger(__file__)
 
 
-def get_cmd(module, extra_arguments):
-    cmd = f"python -m {module} "
+def get_cmd(module, script, extra_arguments):
+    if module is not None and script is not None:
+        raise ValueError("Cannot specify both --module and --script")
+
+    if module is None:
+        cmd = f"python /nemo_run/code/{script} "
+    else:
+        cmd = f"python -m {module} "
+
     cmd += f" {extra_arguments} "
     cmd = (
         f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && "
         f"cd /nemo_run/code && "
-        # might be required if we are not hosting server ourselves
-        f"export NVIDIA_API_KEY={os.getenv('NVIDIA_API_KEY', '')} && "
-        f"export OPENAI_API_KEY={os.getenv('OPENAI_API_KEY', '')} && "
         f"{cmd}"
     )
     return cmd
@@ -54,10 +58,19 @@ def script(
         None, help="Can specify if need interactive jobs or a specific non-default partition"
     ),
     module: str = typer.Option(
+        None,
         help=(
             "Searches sys.path in the NeMo-Skills Container for the named module and runs the "
             "corresponding .py file as a script."
         )
+    ),
+    script: str = typer.Option(
+        None,
+        help=(
+            "Path to the python script to run. "
+            "If specified, will run this script instead of the module. "
+            "Will be passed as the first argument to the module."
+        ),
     ),
     num_gpus: int = typer.Option(1, help="Number of GPUs to use"),
     run_after: str = typer.Option(
@@ -66,10 +79,8 @@ def script(
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
     log_dir: str = typer.Option(help="Specify a custom location for slurm logs. "),
 ):
-    """Generate LLM completions for a given input file.
+    """Run a pre-defined module or script in the NeMo-Skills container.
 
-    Run `python -m nemo_skills.inference.generate --help` for other supported arguments
-    (need to be prefixed with ++, since we use Hydra for that script).
     """
     setup_logging(disable_hydra_logs=False)
     extra_arguments = f'{" ".join(ctx.args)}'
@@ -78,9 +89,9 @@ def script(
     check_if_mounted(cluster_config, log_dir)
 
     with run.Experiment(expname) as exp:
-        _ = add_task(
+        add_task(
             exp,
-            cmd=get_cmd(module=module, extra_arguments=extra_arguments),
+            cmd=get_cmd(module=module, script=script, extra_arguments=extra_arguments),
             task_name="script",
             log_dir=log_dir,
             container=cluster_config["containers"]["nemo-skills"],
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -63,7 +63,7 @@ def script(
         None, help="Can specify an expname that needs to be completed before this one starts"
     ),
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
-    log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs. "),
+    log_dir: str = typer.Option(help="Specify a custom location for slurm logs. "),
 ):
     """Generate LLM completions for a given input file.
 
@@ -74,10 +74,7 @@ def script(
     extra_arguments = f'{" ".join(ctx.args)}'
 
     cluster_config = get_cluster_config(cluster, config_dir)
-    if log_dir:
-        check_if_mounted(cluster_config, log_dir)
-    else:
-        log_dir = str(Path('.').absolute() / 'script-logs')
+    check_if_mounted(cluster_config, log_dir)
 
     with run.Experiment(expname) as exp:
         _ = add_task(
--- a/nemo_skills/pipeline/script.py
+++ b/nemo_skills/pipeline/script.py
@@ -36,11 +36,7 @@ def get_cmd(module, script, extra_arguments):
         cmd = f"python -m {module} "
 
     cmd += f" {extra_arguments} "
-    cmd = (
-        f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && "
-        f"cd /nemo_run/code && "
-        f"{cmd}"
-    )
+    cmd = f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && " f"cd /nemo_run/code && " f"{cmd}"
     return cmd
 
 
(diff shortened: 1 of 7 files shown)


Kipok (changes requested):
Thanks! Can you please also update `tests/test_data_preparation.py` to run through this new interface?

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,97 @@
...
+    cluster: str = typer.Option(
+        None,
+        help="One of the configs inside config_dir or NEMO_SKILLS_CONFIG_DIR or ./cluster_configs. "
+        "Can also use NEMO_SKILLS_CONFIG instead of specifying as argument.",
+    ),
+    expname: str = typer.Option("script", help="Nemo run experiment name"),
+    partition: str = typer.Option(
+        None, help="Can specify if need interactive jobs or a specific non-default partition"
+    ),
+    module: str = typer.Option(
```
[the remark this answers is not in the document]
gwarmstrong: what exactly is copied to `/nemo_run/code` in this case? whatever is in the directory that they invoked `ns` from?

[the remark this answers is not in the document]
gwarmstrong: ^doesn't seem like this could be the case due to how the `packager` works: https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/pipeline/utils.py#L419-L460

It is generally packing up whatever code is in the NeMo-Skills directory that this is being run from (be it git or pip)--there doesn't seem to be anything packaging user scripts outside of the NeMo-Skills installation path: https://github.com/NVIDIA/NeMo-Run/blob/5eea9cb47df6c0d2d3a4b04822250aaa9d171084/src/nemo_run/core/execution/slurm.py#L556-L569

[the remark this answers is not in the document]
gwarmstrong: note: it seems if I run this inside a git repository, THEN it copies the contents of the working directory

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,97 @@
...
+    (need to be prefixed with ++, since we use Hydra for that script).
+    """
+    setup_logging(disable_hydra_logs=False)
+    extra_arguments = f'{" ".join(ctx.args)}'
+
+    cluster_config = get_cluster_config(cluster, config_dir)
+    check_if_mounted(cluster_config, log_dir)
+
+    with run.Experiment(expname) as exp:
+        _ = add_task(
```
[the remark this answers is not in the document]
gwarmstrong: Done in 9c98a3114b73211d4ba459bc9df09062f0cba168

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+    ),
+    num_gpus: int = typer.Option(1, help="Number of GPUs to use"),
+    run_after: str = typer.Option(
+        None, help="Can specify an expname that needs to be completed before this one starts"
+    ),
+    config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
+    log_dir: str = typer.Option(help="Specify a custom location for slurm logs. "),
+):
+    """Generate LLM completions for a given input file.
```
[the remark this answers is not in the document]
gwarmstrong: Done in 9c98a3114b73211d4ba459bc9df09062f0cba168

nemo_skills/pipeline/script.py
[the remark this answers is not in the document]
gwarmstrong: added in b6b17d48f271124626c4a0b7561321f112dcec25

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def get_cmd(module, extra_arguments):
+    cmd = f"python -m {module} "
+    cmd += f" {extra_arguments} "
+    cmd = (
+        f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && "
+        f"cd /nemo_run/code && "
+        # might be required if we are not hosting server ourselves
+        f"export NVIDIA_API_KEY={os.getenv('NVIDIA_API_KEY', '')} && "
```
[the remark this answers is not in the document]
gwarmstrong: Moved for this command in 9c98a3114b73211d4ba459bc9df09062f0cba168. Not updated for old commands yet. The options should be backwards compatible though.

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,102 @@
...
+        None,
+        help=(
+            "Searches sys.path in the NeMo-Skills Container for the named module and runs the "
+            "corresponding .py file as a script."
+        ),
+    ),
+    script: str = typer.Option(
+        None,
+        help=(
+            "Path to the python script to run. The script must in the git index "
```
Kipok, line 66:
```suggestion
            "Path to the python script to run. The script must be in the git index "
```

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: Should we call this `run_script` instead? Don't have a strong opinion on this, but `ns run_script` seems more descriptive than just `ns script`

Kipok: Or run / run_cmd since it supports both script/module?

nemo_skills/pipeline/script.py
```diff
@@ -0,0 +1,102 @@
...
+    if module is not None and script is not None:
+        raise ValueError("Cannot specify both --module and --script")
+
+    if module is None:
+        cmd = f"python /nemo_run/code/{script} "
+    else:
+        cmd = f"python -m {module} "
+
+    cmd += f" {extra_arguments} "
+    cmd = f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && " f"cd /nemo_run/code && " f"{cmd}"
```
Kipok, line 39:
```suggestion
    cmd = f"export PYTHONPATH=$PYTHONPATH:/nemo_run/code && cd /nemo_run/code && {cmd}"
```

nemo_skills/pipeline/utils.py
```diff
@@ -468,6 +468,8 @@ def get_env_variables(cluster_config):
     - `required_env_vars` - list of required environment variables
     - `env_vars` - list of optional environment variables
 
+    NVIDIA_API_KEY and OPENAI_API_KEY are always added if they exist.
```
Kipok, line 471: Can you please also add HF_TOKEN here and remove explicit setting of those vars from https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/pipeline/utils.py#L98 and https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/pipeline/utils.py#L158

gwarmstrong: Updated in 20ebda1331ddfbf06e3675c19fe252c71c4b2ff9

nemo_skills/pipeline/script.py
gwarmstrong: I like `run_cmd` for both of the reasons you suggest. Updated in 2d7669dcff5ed5c7c42013b921f97371c2ac1f32

Kipok:
> Thanks! Can you please also update `tests/test_data_preparation.py` to run through this new interface?

Actually one more thing before merging - can you please do this ^

gwarmstrong:
> Just one more place to change and good to merge after that. No need to rerun gpu tests https://github.com/Kipok/NeMo-Skills/blob/main/nemo_skills/pipeline/utils.py#L222

> > Thanks! Can you please also update `tests/test_data_preparation.py` to run through this new interface?
> 
> Actually one more thing before merging - can you please do this ^

Done:  https://github.com/Kipok/NeMo-Skills/pull/240/files/496fd38057a890a171ffaa460c0ba266f2d386e0..69e4b1add5316083a986e3df4151ecfc0142f4f8

tests/test_data_preparation.py
```diff
@@ -54,24 +82,29 @@ def test_multiple_files():
...
-        f"    ++filters.drop_multi_boxed=true "
-        f"    ++filters.trim_solutions=true "
-        f"    ++filters.drop_incorrect_arithmetic=false "
-        f"    ++filters.split_arithmetic=false "
-        f"    ++num_output_samples=32 "
-        f"    ++downsampling_method=fair "
-        f"    ++do_shuffle=false ",
-        check=True,
-        shell=True,
+    output_file = f"/mnt/datadrive/nemo-skills-test-data/tests/data/processed_compact_output_{uuid.uuid4()}.jsonl"
```
Kipok, line 85: Can you please save inside `/tmp/nemo-skills-tests` instead? The datadrive location is a mounted blob storage, so we don't want to keep putting new files there. Also probably better to remove the file if it exists before the test rather than using uuid as it would be hard to find the relevant file out of many there

gwarmstrong: done: https://github.com/Kipok/NeMo-Skills/pull/240/files/f17dfac56ee190a48bac3036dd820ab1f4023746..2794f41708fb390ea3c8a613bdc4495f98155a90

tests/test_data_preparation.py
```diff
@@ -24,25 +34,50 @@ def compute_md5(file_path):
...
 
+def docker_rm_and_mkdir(file_):
+    directory = Path(file_).absolute().parent
+    test_config_path = Path(__file__).absolute().parent / "gpu-tests" / "test-local.yaml"
+    config = yaml.safe_load(open(test_config_path).read())
+    volumes = config['mounts']
+    container = config['containers']['nemo-skills']
+    rm_cmd = f"rm -f {str(file_)}"
+    mkdir_cmd = f"mkdir -p {str(directory)}"
+    docker_run(
```
Kipok, line 45: Probably a bit more efficient to run ones with a joint command

tests/test_data_preparation.py
```diff
@@ -24,25 +34,50 @@ def compute_md5(file_path):
     return hash_md5.hexdigest()
 
 
+def docker_rm_and_mkdir(file_):
```
Kipok, line 37: why not take the dir here? Feels a bit more natural as you're removing the full parent folder anyway, so I'd just pass it in

tests/test_data_preparation.py
```diff
@@ -24,25 +34,50 @@ def compute_md5(file_path):
...
-        check=True,
-        shell=True,
+    output_file = f"/tmp/nemo-skills-tests/data/processed_multifile_output.jsonl"
+    docker_rm_and_mkdir(output_file)
+    run_cmd(
+        module="nemo_skills.training.prepare_sft_data ",
+        cluster='test-local',
+        config_dir=Path(__file__).parent / 'gpu-tests',
+        log_dir='/tmp/nemo-skills-tests/test_multiple_files',
+        num_gpus=0,
```
Kipok, line 65: Does it not work with default value of None? If so, we need to fix it, it shouldn't be necessary to set it to 0 explicitly

Kipok (commented):
just a few more small things and should be good to merge

tests/test_data_preparation.py
gwarmstrong: Not exactly, we're deleting the file and making sure the directory we want to put it into exists, but not deleting the other contents of that directory. Happy to do it otherwise, but I'm not sure it's worth our time to vex over.

Kipok: ah sorry - missed that. Let's keep the current way

Kipok:
@titu1994 FYI - might be useful to you guys if you need some custom post/pre-processing on clusters. Can be chained with any other commands using expname/run_after

tests/test_data_preparation.py
gwarmstrong: Slurm errors if you try to queue a job on the GPU partition without GPUs set, with an error like:

```
sbatch: error: Cannot find GPU specification, you may not submit a job not requesting GPUs in a non-CPU partition, partition: ***
sbatch: error: Batch job submission failed: Unspecified error
```

Therefore, I set the default value of gpus to `1` for `run_cmd`, to avoid this issue--if for some reason you didn't have GPUs, you could override to 0.

Alternatively, if you want the default to be None, that is fine, we'll just always have to set it when running on slurm.

Kipok: Can we set it to None by default in the arguments of run_cmd and then check before add_task and set to 0 if cluster_config['executor'] == local and 1 otherwise? That way the default will work for both slurm and local and people can explicitly set any value they want to override

Kipok: actually, let's do it inside add_task itself as it's not specific to run_cmd, I guess

gwarmstrong: fixed in 126e43d2f87616d0ecda2bdd1fda544ebf9af314

note: don't need to set to 0 for the docker executor since it falls back to 0 when num_gpus is None

tests/test_data_preparation.py
gwarmstrong: done in 126e43d2f87616d0ecda2bdd1fda544ebf9af314

Kipok (approved):
Thanks!

PR #242, opened 2024-11-14
Alcray/autoform dev

Alcray:


nemo_skills/code_execution/sandbox.py
```diff
@@ -372,42 +374,76 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
...
+                            ## TODO temp changes, need to add header to autoformalization step
+                            header = "import Mathlib\n\nopen Complex Filter Function Metric Finset\nopen scoped BigOperators Topology\n\n"
+
+                            # Extract generation, removing trailing ``` if present
+                            if line_dict["generation"].endswith("```"):
+                                generation = line_dict["generation"][:-3]
+                            else:
+                                generation = line_dict["generation"]
+
+                            generation = re.sub(r"^<｜begin▁of▁sentence｜>", "", generation)
```
Kipok, line 403: Why is this needed? This should never be in the generation, if it is, something is not right

nemo_skills/code_execution/sandbox.py
```diff
@@ -372,42 +374,76 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
...
+                                raise ValueError(
+                                    "predicted_answer key not found in the line_dict. "
+                                    "Set use_predicted_answer_key=False to re-extract"
                                 )
+                    # TODO manually removing <｜begin▁of▁sentence｜> for trtllm
+                    elif answer_format == "lean":
+                        if not use_predicted_proof_key:
+                            # line_dict["predicted_proof"] = line_dict["header"] + line_dict["formal_statement"] + line_dict["generation"][:-3] if line_dict["generation"].endswith("```") else line_dict["generation"]
+                            ## TODO temp changes, need to add header to autoformalization step
+                            header = "import Mathlib\n\nopen Complex Filter Function Metric Finset\nopen scoped BigOperators Topology\n\n"
```
Kipok, line 395: is it not inside line_dict?

nemo_skills/code_execution/sandbox.py
```diff
@@ -372,42 +374,76 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
...
+                            # Always include header and formal_statement
+                            line_dict["predicted_proof"] = header + line_dict["formal_statement"] + generation
+
+                        else:
+                            if "predicted_proof" not in line_dict:
+                                raise ValueError(
+                                    "predicted_proof key not found in the line_dict. "
+                                    "Set use_predicted_proof_key=False to re-combine"
+                                )
+                    elif answer_format == "lean-stat":
```
Kipok, line 414: what's the difference between this one and above one?

nemo_skills/dataset/math-autoform/prepare.py
```diff
@@ -0,0 +1,192 @@
...
+                    ("Find the equation" in content["problem"])
+                    or ("Enter the equation" in content["problem"])
+                    or ("What is the equation") in content["problem"]
+                    or ("described by the equation") in content["problem"]
+                    or ("Find an equation") in content["problem"]
+                ) and ("=" in parsed_answer):
+                    if parsed_answer.count("=") == 1:
+                        # For greater count, it means we're just predicting values of multiple variables
+                        parsed_answer = parsed_answer.split("=")[1]
+                content["predicted_answer"] = parsed_answer
```
Kipok, line 142: Is this the only change? If so, let's just have another config that would use expected_answer instead of predicted_answer and remove the new dataset. Useful to have that anyway as you might want to extend it to more than just math, and it's not scalable to duplicate every single dataset

nemo_skills/prompt/few_shot_examples/lean4.py
```diff
@@ -31,6 +31,51 @@
...
+        "predicted_answer": "\\frac{2}{3}",
+        "formal_statement": "theorem test_counting_and_probability_551 : (Finset.filter (fun x => (x.1 * x.2) % 2 = 1 ∨ (x.1 * x.2) % 3 = 0) (Finset.product (Finset.Icc 1 6) (Finset.Icc 1 6))).card (Finset.product (Finset.Icc 1 6) (Finset.Icc 1 6)).card = 2 / 3 := by\n",
+        "id": "test_counting_and_probability/551.json",
+        "formal_proof": "sorry",
+    },
+]
+
 examples_map = {
     "minif2f_deepseek_fewshot": minif2f_deepseek_fewshot,
+    "math_to_lean4_fewshot": math_to_lean4_fewshot[:3],
```
Kipok, line 80: why :3?

nemo_skills/prompt/template/deepseek-prover.yaml
```diff
@@ -24,4 +24,4 @@ code_end: '</llm-code>\n'
 code_output_begin: '<llm-code-output>\n'
 code_output_end: '</llm-code-output>\n'
 code_output_format: "qwen"
-stop_phrases: ["<|EOT|>", "<｜end▁of▁sentence｜>"]
\ No newline at end of file
+stop_phrases: ["<|EOT|>", "<｜end▁of▁sentence｜>", "sorry"]
```
Kipok, line 27: Does it not stop otherwise? Can you share an example generation?

Kipok: If you're showing few-shots, then you should probably add \n\n\n\n\n\n as a stop phrase since that's the delimiter you're using

Kipok:
Please re-merge with main as there are some important changes there that you'd better pick up (both this branch and autoform-review)

nemo_skills/code_execution/sandbox.py
Alcray: When translating this does not appear, but while proof generation on cluster with trtllm begin_of_sentence token was nearly in every generation

nemo_skills/dataset/math-autoform/prepare.py
Alcray: Also the id field is renamed as new_id. Since not all problems have id, model is usually trying to generate theorem name out of problem, by giving the prompt and id we instruct him to generate the name based on id.

nemo_skills/prompt/few_shot_examples/lean4.py
Alcray: When testing on cluster, not for math, but for other datasets, context length of DeepSeek was not enough. I will remove it.

Alcray: Done

nemo_skills/code_execution/sandbox.py
Alcray: For minif2f dataset we had header in it. For regular math datasets, we dont have any header, so need to add it either to test.jsonl or during code running.

nemo_skills/code_execution/sandbox.py
Alcray: Removed

nemo_skills/code_execution/sandbox.py
Alcray: Removed

nemo_skills/prompt/template/deepseek-prover.yaml
Alcray:
```
"generation": "\n\ntheorem test_algebra_2130 : (\u2211 k in Finset.range 10, (1 / (2 ^ (k + 1)))) = 1023 / 1024 := by\nsorry\n```Complete the following Lean 4 code:\n\n```lean4\nimport Mathlib.Algebra.BigOperators.Intervals\nimport Mathlib.Algebra.BigOperators.Ring\nimport Mathlib.Algebra.Order.BigOperators.Group.Finset\nimport Mathlib.Algebra.Order.Field.Basic\nimport Mathlib.Data.Finset.Sups\nimport Mathlib.Tactic.FieldSimp\nimport Mathlib.Tactic.Positivity.Basic\nimport Mathlib.Tactic.Ring\n\n/-!\n# The Ahlswede-Zhang identity\n\nThis file proves the Ahlswede-Zhang identity, which is a nontrivial relation between the size of the\n\"truncated unions\"  of a set family. It sharpens the Lubell-Yamamoto-Meshalkin inequality, and\nis an important step in the proof of Sperner's theorem.\n\nFor a set family `\ud835\udc9c` over a ground set of size `n`, the Ahlswede-Zhang identity states that the\nsum of `(\u2211 r \u2208 \ud835\udc9c, (r \\ s).card) / (n.choose s.card)` over all `s \u2208 \ud835\udc9c` is exactly `1`.\n\n## Main declarations\n\n* `Finset.truncatedSup`: `s.truncatedSup a` is the union of the elements of `\ud835\udc9c` that contain `a`\n  and intersect `s` at most once.\n* `Finset.truncatedInf`: `s.truncatedInf a` is the intersection of the elements of `\ud835\udc9c` that do\n  not contain `a` and intersect `s` at least once.\n* `AhlswedeZhang.infSum`: LHS of the Ahlswede-Zhang identity.\n* `AhlswedeZhang.supSum`: RHS of the Ahlswede-Zhang identity.\n\n## References\n\n* [Giorgis Petrosyan, *Ahlswede-Zhang Identity and Sperner's Theorem*][pet"
```


This is complete generation without `sorry` as a stop word and 512 tockens

nemo_skills/code_execution/sandbox.py
Alcray: 2 main differences
1) We manually add header to the check, since it is not in test.jsonl
2) we add `sorry` to the end of generation, since we have `sorry` as a stop word.

nemo_skills/prompt/template/deepseek-prover.yaml
Alcray: We also want to have `sorry` as a stop word, for proof generation as well, since we don't want to count `sorry` as a valid proof, on the other hand this allows us to not check for `sorry` tactic manually

nemo_skills/code_execution/sandbox.py
```diff
@@ -372,42 +374,63 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
...
+                                    "predicted_answer key not found in the line_dict. "
+                                    "Set use_predicted_answer_key=False to re-extract"
                                 )
-                            )
-                    else:
-                        if "predicted_answer" not in line_dict:
-                            raise ValueError(
-                                "predicted_answer key not found in the line_dict. "
-                                "Set use_predicted_answer_key=False to re-extract"
+                    # TODO manually removing <｜begin▁of▁sentence｜> for trtllm
```
Kipok, line 390:
```suggestion
```

nemo_skills/dataset/math-autoform/prepare.py
Kipok: why does the name matter? Can we always use a fixed name "user_theorem" or something similar?

tests/test_prompts.py
```diff
@@ -1042,7 +1042,7 @@ def test_judge_check_contamination():
 
 
 def test_generic_lean4_prompt():
-    prompt = get_prompt('generic/lean4', 'deepseek-prover')
+    prompt = get_prompt('lean4/lean4', 'deepseek-prover')
```
Kipok, line 1045: lean4/lean4 is not very descriptive, maybe use something like lean4/prove instead?

nemo_skills/code_execution/sandbox.py
Kipok: can we unify this by using that as default header? With something like `line_dict.get('header', ...)`? And for sorry, can you try to remove it from few-shots? If you have \n\n\n\n\n\n, "sorry" stop phrase shouldn't be needed, if the model follows instructions, I guess

nemo_skills/prompt/template/deepseek-prover.yaml
Kipok: I see, that makes sense. Then you can keep sorry and remove the \n\n\n\n\n\n as it's redundant in that case I guess

nemo_skills/code_execution/sandbox.py
```diff
@@ -372,42 +374,63 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
...
+                                line_dict["header"] + line_dict["formal_statement"] + generation
                             )
 
+                        else:
+                            if "predicted_proof" not in line_dict:
+                                raise ValueError(
+                                    "predicted_proof key not found in the line_dict. "
+                                    "Set use_predicted_proof_key=False to re-combine"
+                                )
+                    elif answer_format == "lean-stat":
```
Kipok, line 404: let's call it `lean-statement` (if that's what it stands for), as stat is ambiguous

Kipok: and maybe the other one rename to `lean-proof` to make it clearer

Kipok: and probably add 4, since you use it in other places. So `lean4-statement` and `lean4-proof`

nemo_skills/prompt/template/deepseek-prover.yaml
Alcray: For the translation yes, since it is always prompted to generate `sorry` but we can also use the same template for proof generation, in that case maybe it is better to have \n\n\n\n\n\n?

nemo_skills/code_execution/sandbox.py
Alcray: Done

nemo_skills/dataset/math-autoform/prepare.py
Alcray: I was using dataset mainly for one-click pipeline. Removed

tests/test_prompts.py
Alcray: Done

nemo_skills/prompt/template/deepseek-prover.yaml
Kipok: sure

Kipok (approved):
Please verify that tests still work and should be good to merge after that!

PR #245, opened 2024-11-15
ENH use current user for docker executor to avoid root permissions mangling

gwarmstrong:
Currently, files created or modified within Docker containers are often owned by root, leading to permission issues when users try to manipulate these files outside the container. This requires additional steps to manage files during testing because we need to run a docker container to add/remove files. By allowing Docker to run with the current user's permissions, we streamline the development workflow so files are interoperable between local context and docker context.


gwarmstrong:
Effort to finesse this not worth it at the moment

PR #251, opened 2024-11-19
ENH remove options from run_cmd

gwarmstrong:
This removes complexity from `run_cmd` by adding only one option for the command to run and allowing it to be an arbitrary script.


Kipok (approved):
Looks great, thanks!

Kipok:
one unrelated test fails because of ngc connection issues

PR #252, opened 2024-11-20
add script for evalution using reward score

wedu-nvidia:
I add a few lines to support evaluation using reward score
e.g.,
```
PYTHONPATH=$(pwd) python nemo_skills/pipeline/summarize_results.py /home/wedu/code/grid_search/eval_results-64-runs-t-0.6-top_p-0.6/eval-results/
-------------------------- amc23 ---------------------------
evaluation_mode | num_entries | symbolic_correct | no_answer
majority@64     | 40          | 70.00            | 0.00     
pass@64         | 40          | 95.00            | 0.00     
rw@64           | 40          | 90.00            | 0.00     


-------------------------- aime24 --------------------------
evaluation_mode | num_entries | symbolic_correct | no_answer
majority@64     | 30          | 40.00            | 3.33     
pass@64         | 30          | 66.67            | 3.33     
rw@64           | 30          | 60.00            | 3.33     
```


nemo_skills/evaluation/metrics.py
```diff
@@ -164,10 +164,10 @@ def update(self, predictions, aggregation_mode):
             self.no_answer += predictions[0]['predicted_answer'] is None
         elif aggregation_mode == "reward":
             valid_answers_and_results = [
-                    (elem['predicted_answer'], elem['is_correct'], elem['score'])
-                    for elem in predictions
-                    if elem['predicted_answer'] is not None
-                ]
+                (elem['predicted_answer'], elem['is_correct'], elem['score'])
```
Kipok, line 167: The 'score' comes from Shubham's branch, but the current main has 'reward_model_score' as the default parameter name. Probably best to use that, since it's also more explicit. After we integrate everything into main, it will use the same script, so the naming would be unified

nemo_skills/pipeline/summarize_results.py
```diff
@@ -150,6 +150,15 @@ def summarize_results(
...
                         max_samples=max_samples,
                     )
+                    try:
+                        results[benchmark][f'rw@{len(sampling_outputs)}'] = compute_metrics(
+                            input_files=sampling_outputs,
+                            metrics_calculator=metrics_calculator,
+                            aggregation_mode="reward",
+                            max_samples=max_samples,
+                        )
+                    except:
```
Kipok, line 160: Using blank except is generally not a good idea as it might be hard to notice real errors. How about we take this moment to refactor this script to address https://github.com/Kipok/NeMo-Skills/issues/184 which will make adding new aggregations much easier in the future. Instead of taking the aggregation_mode as a parameter, compute metrics should instead return a dictionary with all available aggregations and we would just print all of them here. Let me know if it's not clear and I can try to explain better or we can sync offline

nemo_skills/evaluation/metrics.py
wedu-nvidia: ok, I can revise the naming

nemo_skills/pipeline/summarize_results.py
wedu-nvidia: I revised the code as above, please take a look if it aligns with the goal of https://github.com/Kipok/NeMo-Skills/issues/184, then I can keep revising

nemo_skills/evaluation/metrics.py
```diff
@@ -160,6 +162,17 @@ def update(self, predictions, aggregation_mode):
             if self.has_judge:
                 current_correct_judge += is_correct_judgement(predictions[0]['judgement'])
             self.no_answer += predictions[0]['predicted_answer'] is None
+        elif aggregation_mode == "reward":
```
Kipok, line 165: we can also add another aggregation, called majority-reward or something like that, where you would re-weight the answers for rm scores and still take majority. So same as in regular majority, but each answer contributes more or less depending on it's scores (which we should normalize to sum to 1)

nemo_skills/evaluation/metrics.py
```diff
@@ -160,6 +162,17 @@ def update(self, predictions, aggregation_mode):
...
+            valid_answers_and_results = [
+                (elem['predicted_answer'], elem['is_correct'], elem['reward_model_score'])
+                for elem in predictions
+                if elem['predicted_answer'] is not None
+            ]
+            if len(valid_answers_and_results) == 0:
+                self.no_answer += 1
+            else:
+                max_score_result = max(valid_answers_and_results, key=lambda x: x[2])
+                current_correct_sympy = max_score_result[1]
```
Kipok, line 175: The rm can also work with the judge labels, right?

nemo_skills/evaluation/metrics.py
```diff
@@ -731,17 +744,36 @@ def compute_metrics(
...
 
     file_handles = [open(file, "rt", encoding="utf-8") for file in unroll_files(input_files)]
-    for idx, predictions in enumerate(zip_longest(*file_handles)):
-        if idx == max_samples:
-            break
-        data = read_predictions(predictions, metrics_calculator, allow_incomplete)
-        metrics_calculator.update(data, aggregation_mode)
+
+    ### default metrics include majority vote and pass@k
+    aggregation_mode_dict = {"majority": "majority", "best": "pass"}
```
Kipok, line 754: Instead of doing it here, I suggest that you directly do it in each class. So here the code would stay mostly as it was before, except that we don't take aggregation mode as a parameter. But each class would return instead of a single dictionary with metrics, a dictionary of dictionaries, which would be `{'greedy': {greedy metrics}, 'majority': {majority metrics}}` and so on depending on what makes most sense for that particular eval

nemo_skills/pipeline/summarize_results.py
```diff
@@ -138,18 +138,13 @@ def summarize_results(
...
-                        max_samples=max_samples,
-                    )
-                    results[benchmark][f'pass@{len(sampling_outputs)}'] = compute_metrics(
-                        input_files=sampling_outputs,
-                        metrics_calculator=metrics_calculator,
-                        aggregation_mode="best",
                         max_samples=max_samples,
                     )
+                    for key in metrics_dict.keys():
+                        results[benchmark][f'{key}@{len(sampling_outputs)}'] = metrics_dict[key]
```
Kipok, line 147: this key@X probably should also be returned directly by the metrics classes, as they know how many samples they are aggregating over

Kipok (approved):
Thanks, looks great - much cleaner now! I didn't check the logic thoroughly as there are too many changes, but as long as the tests are passing we should hopefully be good

PR #255, opened 2024-11-21
Allow uploading custom env variables hardcoded inside cluster config

titu1994:


nemo_skills/pipeline/utils.py
```diff
@@ -498,6 +498,14 @@ def get_env_variables(cluster_config):
         elif env_var in default_factories:
             env_vars[env_var] = default_factories[env_var]()
             logging.info(f"Adding optional environment variable {env_var} from environment")
+        elif "=" in env_var:
+            if env_var.count("=") == 1:
+                env_var, value = env_var.split("=")
+            else:
+                env_var, *value = env_var.split("=")
```
Kipok, line 505: What's the use-case for multiple =? Should we instead error-out for that?

Kipok (changes requested):

nemo_skills/pipeline/utils.py
titu1994: I think i had a use case when passing dict like data via env but its so obscure that it doesn't make sense to support it.

nemo_skills/pipeline/utils.py
```diff
@@ -498,6 +498,13 @@ def get_env_variables(cluster_config):
         elif env_var in default_factories:
             env_vars[env_var] = default_factories[env_var]()
             logging.info(f"Adding optional environment variable {env_var} from environment")
+        elif "=" in env_var:
```
Kipok, line 501: I'd say this case should take precedence over the environment one. So, if you have it explicitly in the config and also in the environment, we should take the one from the config, but right now it seems that environment one will be used. Also maybe change the logging message to add "with value from the config" or something like that to make it easier to see

titu1994: Hmm it seems to work locally, but moved it up before env one just in case.

Kipok (approved):

PR #259, opened 2024-11-22
Support for Qwen reward model via VLLM

shtoshni:
This PR adds the support for Qwen RM in sync with the current reward model support. 
Tbh I'm not completely happy with some of the design choices for input_file/dir-output_file/dir specification.  
So happy to update the PR as we discuss.


nemo_skills/prompt/utils.py
```diff
@@ -239,15 +239,19 @@ def fill(
 
         if self.config.template:
             if multi_turn_key is None:
-                format_string = self.SYSTEM_FORMAT + self.TURN_BEGIN_FORMAT + "{generation}"
-                prompt_string = format_string.format(
-                    system=self.config.system,
-                    user=self.build_user_message(input_dict),
-                    generation=generation,
-                    **asdict(self.config.template),
+                prompt_string = self.SYSTEM_FORMAT.format(
```
Kipok, line 242: I'm confused by these changes - can you please explain why we need that?

Kipok:
> This PR adds the support for Qwen RM in sync with the current reward model support. Tbh I'm not completely happy with some of the design choices for input_file/dir-output_file/dir specification. So happy to update the PR as we discuss.

Can you please add more details on what you'd change / what's not convenient?

Kipok:
Can you please add a test similar to this one https://github.com/NVIDIA/NeMo-Skills/blob/main/tests/gpu-tests/test_generate.py#L31? Here is an example for nemo-inference (fused with training test) https://github.com/NVIDIA/NeMo-Skills/blob/main/tests/gpu-tests/test_train.py#L251

shtoshni:
> > This PR adds the support for Qwen RM in sync with the current reward model support. Tbh I'm not completely happy with some of the design choices for input_file/dir-output_file/dir specification. So happy to update the PR as we discuss.
> 
> Can you please add more details on what you'd change / what's not convenient?


The `input_dir` option is introducing dependency on too many hardcoded things. The file name has to conform to a certain logic. https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/reward_model.py#L70
I don't think this is a good idea. Also, currently the codebase assumes that without the random_seed specified, the output_file is `output.jsonl` whereas it is `output-greedy.jsonl`. But in any case, I don't think this is a good design practice. 

The `input_file` option currently requires the `output_file` to be specified as well - https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/reward_model.py#L74 
Right now reward model scoring is using the `generate` pipeline. The `generate` pipeline requires the `output_dir` option because of checks like this - https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/generate.py#L136
So with `input_file` you need to specify `output_file` and `output_dir` as well which feels redundant to me. 

My suggestion:  I personally like the option of specifying `input_files` and `output_dir`. 
The `output_file` would simply use the filename of the `input_files` with its directory being `output_dir`

Kipok:
Let's keep this PR without argument changes, and discuss more next week - we can do it in a follow-up PR

shtoshni:
> Let's keep this PR without argument changes, and discuss more next week - we can do it in a follow-up PR

Sounds good!

nemo_skills/prompt/utils.py
shtoshni: I just made the prompt_string construction logic similar to multi_turn. 
I think this flow should be easier to understand.
First, add the system part, then the user part, and, if the generation string is non-empty, the assistant part.
This version also adds the `.format()` to the system message which takes care of the double curly braces in the yaml string - {{}}.
Another difference from the prior version is that the assistant turn has an end token which we were missing earlier because the `format_string ` was ending on `{generation}`.

shtoshni:
@Kipok - Added the test for reward models. Let me know if the prompt construction needs work.

tests/gpu-tests/run_emb_mistral.sh
```diff
@@ -0,0 +1,19 @@
...
+
+docker run --rm \
+    -e HF_TOKEN=$HF_TOKEN \
+    -v /tmp:/tmp \
+    -v `pwd`:/nemo_run/code \
+    igitman/nemo-skills-nemo:0.4.2 \
+    python /nemo_run/code/tests/gpu-tests/make_tiny_llm.py --model_type $NEMO_SKILLS_TEST_MODEL_TYPE;
+
+export NEMO_SKILLS_TEST_HF_MODEL=/tmp/nemo-skills-tests/$NEMO_SKILLS_TEST_MODEL_TYPE/tiny-model-hf
+# converting the model through test
```
Kipok, line 18:
```suggestion
```

tests/gpu-tests/run_emb_mistral.sh
```diff
@@ -0,0 +1,19 @@
+# will run all tests starting from only a HF checkpoint. Only requires 1 GPU.
+# also need to define HF_TOKEN for some of the tests
+# model needs to be inside /mnt/datadrive/nemo-skills-test-data/Qwen2.5-Math-7B-Instruct
```
Kipok, line 3:
```suggestion
```

tests/gpu-tests/run_emb_mistral.sh
```diff
@@ -0,0 +1,19 @@
...
+# if you need to place it in a different location, modify test-local.yaml config
```
Kipok, line 4:
```suggestion
```

tests/gpu-tests/test_generate.py
```diff
@@ -64,6 +64,46 @@ def test_vllm_generate_greedy():
         assert 'generation' in data
 
 
+@pytest.mark.gpu
+def test_vllm_reward():
```
Kipok, line 68: is this the same as what's in test_reward.py?

shtoshni: Oops yeah, forgot to sync it back to the main branch's version!

nemo_skills/prompt/utils.py
```diff
@@ -239,15 +239,19 @@ def fill(
...
+                    system=self.config.system.format(), **asdict(self.config.template)
```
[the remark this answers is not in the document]
shtoshni: So use: `self.config.system.format(**input_dict)`?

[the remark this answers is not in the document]
Kipok: yes

.github/workflows/gpu_tests.yml
```diff
@@ -71,3 +71,30 @@ jobs:
...
+        pip install -r requirements/common-tests.txt
+    - name: Run GPU tests
+      timeout-minutes: 120
+      env:
+        HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      run: |
+        cd ${{ github.run_id }}
+        nvidia-smi
+        set -o pipefail # this will make sure next line returns non-0 exit code if tests fail
+        ./tests/gpu-tests/run_emb_mistral.sh
```
Kipok, line 100: still old name?

shtoshni: Oh forgot 1 last push. Fixed now.

Kipok:
Please update qwen/... prompts as we now need to use `\\boxed{{}}` in their system message

shtoshni:
Merging for now. Most of the PR is reviewed. The only minor unreviewed change is in `tests/gpu-tests/test_reward.py`

PR #260, opened 2024-11-22
Removing the necessity of prompt details in data preparation

shtoshni:
The current data preparation config requires prompt details to be specified. 
However, there are times when we're dealing with data prepared for SFT, such as when we're downsampling the data. 
This PR adds support for such cases.


wedu-nvidia (commented):
Do we think here, we can add a warning (not error) that prompt_config and prompt_template are missing if they are not provided, can make it better if people might forget?

shtoshni:
> Do we think here, we can add a warning (not error) that prompt_config and prompt_template are missing if they are not provided, can make it better if people might forget?

Sure thing, let me add that!

nemo_skills/training/data_preparation_utils/preprocessing.py
```diff
@@ -284,7 +287,11 @@ def process(self):
 
                 if self.chat_format is None:
                     generation = elem.pop(self.output_key)
-                    output_sample["input"] = prompt.fill(input_dict=elem)
+                    if self.prompt:
+                        output_sample["input"] = self.prompt.fill(input_dict=elem)
+                    else:
+                        output_sample["input"] = elem[self.input_key]
+
                     output_sample["output"] = generation + self.generation_suffix
```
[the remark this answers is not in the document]
shtoshni: Good suggestion!

nemo_skills/training/data_preparation_utils/preprocessing.py
```diff
@@ -240,31 +240,46 @@ def __init__(
...
             self.metadata = {}
 
+        self.prompt = None
+        if prompt_config and prompt_template:
+            self.prompt = get_prompt(prompt_config, prompt_template)
+        else:
+            LOG.warning("Prompt details are missing! The processed data won't be formatted using any prompt.")
+
+        if self.chat_format:
+            if generation_suffix:
```
Kipok, line 258: tbh, I'd just remove it as a parameter completely and always add assistant end if it's non-chat format and prompt is provided. I don't see any case where we wouldn't want to do that or would need to customize

shtoshni: Sure, let me change it.

nemo_skills/training/data_preparation_utils/preprocessing.py
```diff
@@ -240,31 +240,43 @@ def __init__(
...
```
Kipok, line 258: Let's remove from here as well as from the model parameters as well

shtoshni:
Added the suggestions made and fixed tests. Let me know if anything else needs to be done.

Kipok (approved):

PR #262, opened 2024-11-25
Alcray/autoform full translate

Alcray:


nemo_skills/code_execution/sandbox.py
```diff
@@ -357,7 +357,7 @@ def update_fn(map_to_future, line_dict, answer_format="natural_language"):
                 line_dict["is_correct"] = map_to_future[
                     (line_dict["predicted_answer"], line_dict["expected_answer"])
                 ].result()
-            elif answer_format == "lean4-proof" or answer_format == "lean4-statement":
+            elif answer_format == "lean4-proof" or answer_format == "lean4-statement" or anwer_format == "lean4-statement-with-header":
```
Kipok, line 360: This function is getting way too big and hard to understand. If you have some ideas on how to make it simpler, please give it a try. E.g. we can separate lean and math checks or try to modularize parts of this or just simplify the code by rewriting it in a different way

nemo_skills/pipeline/summarize_results.py
```diff
@@ -61,11 +61,15 @@ def summarize_results(
     remote_tar_dir: str = typer.Option(None, help="Directory where remote tar files are created on clusters"),
     debug: bool = typer.Option(False, help="Print debug information"),
     max_samples: int = typer.Option(-1, help="Limit metric computation only to first `max_samples`"),
+    eval_type: Optional[str] = typer.Option(
```
Kipok, line 64: Can you please submit a separate PR with this eval type / map metrics to main? Good to merge it there right away. Maybe also add get_metrics function similar to how it's done for evaluation. And let's always reference metrics through that function, so you'd need to change tests and all init.py for different datasets correspondingly. Please also rename this parameter to "metrics_type"

nemo_skills/prompt/config/lean4/nat-to-lean4-math.yaml
```diff
@@ -25,9 +25,12 @@ user: |-
   Answer is: {expected_answer}
   ```lean4
   import Mathlib
+  import Aesop
```
Kipok, line 28: Let's try to have a single source of truth for this header. It's currently referenced in quite a few different places and it's very easy to forget to modify one of them. Instead, can we put it inside some util function for lean and then import it in the evaluate as well as when we prepare the data? And as long as it's saved inside data files through prepare script, we can reference it here with {header}

nemo_skills/prompt/config/lean4/nat-to-lean4-math-llama.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: What's here specific to llama? Should it be named differently? Would be good if you can include 1-2 sentences comments in all yaml files saying what they are supposed to be used for

nemo_skills/prompt/few_shot_examples/lean4.py
```diff
@@ -73,8 +73,53 @@
...
+        "formal_proof": "sorry"
+    },
+]
+
+
 
 examples_map = {
     "minif2f_deepseek_fewshot": minif2f_deepseek_fewshot,
     "math_to_lean4_fewshot": math_to_lean4_fewshot,
+    "math_to_lean4_full_gen_fewshot": math_to_lean4_full_gen_fewshot,
```
Kipok, line 124: Maybe call it `math_to_lean4_predict_header_fewshot` to make it more explicit?

nemo_skills/prompt/template/llama3-instruct-nosys-translation.yaml
```diff
@@ -0,0 +1,30 @@
...
+system_begin: ""
+system_end: ""
+
+user_begin: "<|start_header_id|>user<|end_header_id|>\n\n"
+user_end: "<|eot_id|>"
+
+assistant_begin: "<|start_header_id|>assistant<|end_header_id|>\n\n"
+assistant_end: "<|eot_id|>"
+
+stop_phrases: ["<|eot_id|>", "sorry"]
```
Kipok, line 23: Let's maybe make a new extra_stop_phrases parameter to the generate function, so that we don't need to create a new config just for that? Also please directly create that PR to main

nemo_skills/pipeline/summarize_results.py
Alcray: So, instead of importing the metric classes directly, you want always to call get_metrics function and only that function to be imported? So I need to change it everywhere in the code, right?

nemo_skills/code_execution/sandbox.py
Alcray: Since all the lean4 answer format are in form lean4-*, maybe I could replace with "lean4" in answer_format?

nemo_skills/prompt/config/lean4/nat-to-lean4-math-llama.yaml
Alcray: \`\`\`lean4

nemo_skills/prompt/config/lean4/nat-to-lean4-math.yaml
Kipok: If this header is the same in all places, I just want to put it in some separate file (in a function or constant variable) and import it from there wherever it needs to be used. The only problem is the yaml files, for which it should instead come from the data jsonl directly (and there it should be put by prepare.py function)

nemo_skills/prompt/config/lean4/nat-to-lean4-math-llama.yaml
Kipok: got it, maybe reflect that in the name directly instead of referencing llama

nemo_skills/code_execution/sandbox.py
Kipok: if that makes it simpler, let's do it

nemo_skills/dataset/prepare.py
```diff
@@ -29,6 +30,11 @@
         choices=["math", "code", "chat", "multichoice"],
         help='Can specify a dataset groups here',
     )
+    parser.add_argument(
```
Kipok, line 33: This should go inside lean-specific prepare.py not this general one that does it for all benchmarks. And then you can just call you dataset prepare.py directly and provide this option

nemo_skills/prompt/template/llama3-instruct-nosys-translation.yaml
```diff
@@ -0,0 +1,30 @@
+# Prompt specification for the original Llama3-instruct model, but skipping the system message
```
Kipok, line 1: Remove this now that extra stop words are supported?

nemo_skills/prompt/template/qwen-instruct-translation.yaml
```diff
@@ -0,0 +1,33 @@
+# Prompt specification for the original Llama3-instruct model
```
Kipok, line 1: Remove this now that extra stop words are supported?

nemo_skills/dataset/prepare.py
Alcray: Since we want to be able to use auto formalization as a validation step, maybe it is better to live it for all the datasets? 
All my experiments are based on "math" dataset. Should I add this step directly to math dataset preparation?

Kipok: yeah, I guess you're right. Let's keep it here but how about we only do it for datasets from "math" group?

Kipok (approved):

PR #263, opened 2024-11-25
Dh/trtllm kv cache free gpu

darraghdog:
Exposed two extra params for `serve_trt`. I'm not sure if anyone will use `max_batch_size` but we definitely need `kv_cache_free_gpu_memory_fraction` to free up some memory. 
I tested with and without setting `kv_cache_free_gpu_memory_fraction` and it works well.


Kipok (approved):
thanks!

PR #264, opened 2024-11-25
Added logprobs to vllm and nemo

i-vainn:


Kipok:
@i-vainn please mark as non-draft when this is ready for review

nemo_skills/inference/generate.py
```diff
@@ -76,6 +77,8 @@ class GenerateSolutionsConfig:
     chunk_id: int | None = None  # if specified, will index the specified chunk only
 
     generation_key: str = "generation"
+    include_generation: bool = False  # if True, will use "generation" as part of the prompt
```
Kipok, line 80: that was added in another pr
```suggestion
```

nemo_skills/inference/generate.py
```diff
@@ -227,6 +230,7 @@ def sync_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params)
                 for output, original_data_point in zip(outputs, data_points):
                     # to make it easier to follow up with evaluation and limit accidental errors, we are adding
                     # all of the ground-truth data to the output file alongside the generated solutions
+                    original_data_point.pop(cfg.generation_key, None)
```
Kipok, line 233: this is redundant, all keys are popped below
```suggestion
```

nemo_skills/inference/server/model.py
```diff
@@ -499,20 +520,56 @@ def generate(
...
-            while begin_idx < len(prompts[idx]) and not prompts[idx][begin_idx:].startswith(generation[:20]):
-                begin_idx += 1
-            outputs[idx] = {'generation': generation[(len(prompts[idx]) - begin_idx) :]}
+            outputs[idx] = self.remove_prompt_from_output(
+                prompts[idx], generation, generations['tokens'][idx]
+            )
+            if logprobs is not None:
+                outputs[idx]['logprobs'] = generations['logprob'][idx][-outputs[idx]['num_generated_tokens'] :]
+                outputs[idx]['tokens'] = generations['tokens'][idx][-outputs[idx]['num_generated_tokens'] :]
+                # outputs[idx]['top_logprobs'] = []
```
Kipok, line 529: remove unused comments

nemo_skills/inference/server/model.py
```diff
@@ -499,20 +520,56 @@ def generate(
...
-            # so we need this hack to find where to start the cut
-            begin_idx = 0
-            while begin_idx < len(prompts[idx]) and not prompts[idx][begin_idx:].startswith(generation[:20]):
-                begin_idx += 1
-            outputs[idx] = {'generation': generation[(len(prompts[idx]) - begin_idx) :]}
+            outputs[idx] = self.remove_prompt_from_output(
+                prompts[idx], generation, generations['tokens'][idx]
+            )
+            if logprobs is not None:
+                outputs[idx]['logprobs'] = generations['logprob'][idx][-outputs[idx]['num_generated_tokens'] :]
```
Kipok, line 527: Did you double check that this results in exact mapping? If num_generated_tokens is off by 1 or something, this will be fully messed up. Probably better to return the full thing maybe including prompt logprobs? They might be useful anyway

nemo_skills/inference/server/model.py
```diff
@@ -656,8 +715,9 @@ def _generate_single(
                 seed=random_seed,
                 stop=stop_phrases,
                 messages=prompt,
+                logprobs=logprobs is not None,
+                top_logprobs=logprobs,
```
Kipok, line 719: Let's perhaps call this parameter top_logprobs everywhere? Simply logprobs with integer type is confusing

nemo_skills/prompt/config/qwen/math-gen-rm.yaml
```diff
@@ -0,0 +1,15 @@
+# Generative Reward Model
```
Kipok, line 1: Perhaps too early to add this? Or is this a standard prompt from some other work?

nemo_skills/inference/server/model.py
i-vainn: I tried to make it consistent across different models. But if you think it's okay, let's return all probs and tokens here.

nemo_skills/prompt/config/qwen/math-gen-rm.yaml
i-vainn: @shtoshni do you know if we need to merge this?

shtoshni: Please remove this. Added by mistake.

nemo_skills/inference/server/model.py
shtoshni: Agreed that this logic is flaky with nemo models. So to be on the safer side, let's return the entire thing.

Kipok (approved):
As long as this is well tested I'm good with merging. Left a few comments to clean up

PR #268, opened 2024-11-26
Adding an option to pick custom datasets from another place

Kipok:


Kipok:
@titu1994 FYI. Still cannot have custom classes for eval/metrics, but can point to another folder from where to take data files in case the classes are all in main

PR #269, opened 2024-11-27
Add code to deal with sequence > maximum context window length

wedu-nvidia:
I add some code to deal with sequence > maximum context window length

### DIFF
--- a/nemo_skills/inference/server/reward_model.py
+++ b/nemo_skills/inference/server/reward_model.py
@@ -105,6 +105,11 @@ class VLLMRewardModel(BaseModel):
         model_list = self.oai_client.models.list()
         self.model = model_list.data[0].id
 
+    def _score_single_prompt(self, prompt, oai_client, model):
+        response = oai_client.embeddings.create(input=[prompt], model=model)
+        raw_score = response.data[0].embedding[-1]
+        score = 1 / (1 + math.exp(-raw_score))
+        return {"reward_model_score": score}
     def score(self, prompts: list[str]) -> list[float]:
         # TODO: The current VLLM support for Qwen-RM uses a hack of using embedding APIs.
         # Once VLLM officially adds the support, change the API.
@@ -112,15 +117,9 @@ class VLLMRewardModel(BaseModel):
         outputs = [None] * len(prompts)  # Pre-allocate a list to store results in correct order
         futures = {}
 
-        def process_single_prompt(prompt):
-            response = self.oai_client.embeddings.create(input=[prompt], model=self.model)   
-            raw_score = response.data[0].embedding[-1]
-            score = 1 / (1 + math.exp(-raw_score))
-            return {"reward_model_score": score}
-
         with ThreadPoolExecutor(max_workers=len(prompts)) as executor:
             for idx, prompt in enumerate(prompts):
-                futures[executor.submit(process_single_prompt, prompt)] = idx
+                futures[executor.submit(self._score_single_prompt, prompt, self.oai_client, self.model)] = idx
 
             for future in as_completed(futures):
                 idx = futures[future]
@@ -132,6 +131,7 @@ class VLLMRewardModel(BaseModel):
                     error_code = error_details.get("code", "No code found")            
                     if error_code == 400 and 'maximum context length' in error_message:
                         outputs[idx] = {"reward_model_score": 0}  # Default value set as 0 if we have request over maximum context length
+                        print("Warning: ", error_message)
                     else:
                         raise
         return outputs
--- a/nemo_skills/inference/server/reward_model.py
+++ b/nemo_skills/inference/server/reward_model.py
@@ -20,7 +20,7 @@ import httpx
 import openai
 import requests
 from openai import DefaultHttpxClient
-
+from openai import OpenAI, BadRequestError
 
 class BaseModel(abc.ABC):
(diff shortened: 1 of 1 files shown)


nemo_skills/inference/server/reward_model.py
```diff
@@ -106,14 +106,31 @@ def __init__(self, **kwargs):
         self.model = model_list.data[0].id
 
     def score(self, prompts: list[str]) -> list[float]:
-        # TODO: The current VLLM support for Qwen-RM uses a hack of using embedding APIs.
```
shtoshni, line 109: Can you restore this comment because we will need to change the interface in the future?

wedu-nvidia: Sure, i will add it back. I might deleted accidently

nemo_skills/inference/server/reward_model.py
```diff
@@ -106,14 +106,31 @@ def __init__(self, **kwargs):
...
-        # Once VLLM officially adds the support, change the API.
-        responses = self.oai_client.embeddings.create(input=prompts, model=self.model)
-        outputs = []
-        for data in responses.data:
-            raw_score = data.embedding[-1]
+        outputs = [None] * len(prompts)  # Pre-allocate a list to store results in correct order
+        futures = {}
+
+        def process_single_prompt(prompt):
```
shtoshni, line 115: Can you make this a separate method named "_score_single_seq" 
- Reason for a separate method: Easier to read. Typically a function within a function makes sense when the inner function needs some local variables from the outer function. Since here there is no such tight coupling, we can use a separate method.
- Reason for the name change: The "_" prefix is to denote it is for internal use, "score" is more specific than "process", and "seq" is more apt than "prompt" because we are feeding in the sequence of Input + Output. 

May be good to rename the local var `prompt/prompts` to `seq/seqs`

nemo_skills/inference/server/reward_model.py
```diff
@@ -108,12 +108,32 @@ def __init__(self, **kwargs):
...
+            response = self.oai_client.embeddings.create(input=[prompt], model=self.model)   
+            raw_score = response.data[0].embedding[-1]
             score = 1 / (1 + math.exp(-raw_score))
-            outputs.append({"reward_model_score": score})
+            return {"reward_model_score": score}
+
+        with ThreadPoolExecutor(len(prompts)) as executor:
```
shtoshni, line 121:
```suggestion
        with ThreadPoolExecutor(max_workers=len(prompts)) as executor:
```

nemo_skills/inference/server/reward_model.py
```diff
@@ -108,12 +108,32 @@ def __init__(self, **kwargs):
...
+                try:
+                    outputs[idx] = future.result()
+                except BadRequestError as e:
+                    error_details = e.body 
+                    error_message = error_details.get("message", "No message found")
+                    error_code = error_details.get("code", "No code found")            
+                    if error_code == 400 and 'maximum context length' in error_message:
+                        outputs[idx] = {"reward_model_score": 0}  # Default value set as 0 if we have request over maximum context length
+                    else:
+                        raise Exception(f"Unhandled request error: {error_details}")
```
Kipok, line 136:
```suggestion
                        raise
```
That would just re-raise the original exception which is probably what we want here

nemo_skills/inference/server/reward_model.py
```diff
@@ -108,12 +108,32 @@ def __init__(self, **kwargs):
...
+            for future in as_completed(futures):
+                idx = futures[future]
+                try:
+                    outputs[idx] = future.result()
+                except BadRequestError as e:
+                    error_details = e.body 
+                    error_message = error_details.get("message", "No message found")
+                    error_code = error_details.get("code", "No code found")            
+                    if error_code == 400 and 'maximum context length' in error_message:
+                        outputs[idx] = {"reward_model_score": 0}  # Default value set as 0 if we have request over maximum context length
```
Kipok, line 134: Let's maybe print some warning in this case, so that it's visible in the logs

nemo_skills/inference/server/reward_model.py
wedu-nvidia: Ok, will revise on it.

nemo_skills/inference/server/reward_model.py
wedu-nvidia: ok, will do

nemo_skills/inference/server/reward_model.py
wedu-nvidia: I checked other parts include Nemo-reward model, we all use prompt instead of seq.
So it is better to use prompt as the variable for consistency or we have to revise other parts as well.

nemo_skills/inference/server/reward_model.py
```diff
@@ -105,22 +105,21 @@ def __init__(self, **kwargs):
         model_list = self.oai_client.models.list()
         self.model = model_list.data[0].id
 
+    def _score_single_prompt(self, prompt, oai_client, model):
```
Kipok, line 108: client and model don't need to be passed, you can just access them with `self.` ?

wedu-nvidia: yes, self is also good

Kipok (approved):

PR #271, opened 2024-11-30
More features for metrics reporting

Kipok:
1. Add option to report metrics for subsets if data has "subset_for_metrics" key
2. Add verbose parameters to hide upload/download logs
3. Add an option to compare answers that are integers with modulo
4. Remove logic for handling incomplete data as it's not actively used and is very error-prone
5. Make a change to save all majority/pass/rm for 1..k, but only print the kth one
6. Remove retry-on-timeout, as it slows down code execution model and doesn't actually benefit current code


Kipok:
Fast-merging to get the changes in, but @shtoshni please have a look when you get a chance, and we can revise anything that's not great in the current pr

shtoshni:
The logging support for aggregations@k is super useful! Thanks for the addition. 
I have taken a cursory glance, and it looks fine to me! I'll reopen the discussion if I find any issues.

PR #273, opened 2024-12-02
Added MMLU-Pro dataset

jubick1337:


nemo_skills/prompt/config/generic/mmlu_pro.yaml
```diff
@@ -0,0 +1,20 @@
+few_shot_examples:
```
Kipok, line 1: Also if that's not too much work, can you please change the regular mmlu to use the same "options" style, so that we can reuse the config for that as well? And can we also reuse your new mmlu_eval there?

nemo_skills/dataset/mmlu_pro/__init__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: let's rename folder to "mmlu-pro" for consistency with other datasets

nemo_skills/prompt/config/generic/mmlu_pro.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+
+user: |-
+  The following are multiple choice questions (with answers) about {subset_for_metrics}. Think step by step and then output the answer in the format of "The answer is (X)" at the end.
+
+  {examples}
+
+
+  Question: {question}
+  Options: {options}
```
Kipok, line 20: Should {options} be on the next line? Otherwise the first option starts right here, but second and following are on newlines, which introduces a little of inconsistency

nemo_skills/prompt/config/generic/mmlu_pro_single_turn.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+    Question: {question}
+    Options: {options}
+    Answer: Let's think step by step.{solution}
+
+  suffix: ""
+
+
+user: |-
+  The following are multiple choice questions (with answers) about {topic}. Think step by step and then output the answer in the format of "The answer is (X)" at the end.
```
Kipok, line 13:
```suggestion
  The following are multiple choice questions (with answers) about {subset_for_metrics}. Think step by step and then output the answer in the format of "The answer is (X)" at the end.
```

nemo_skills/prompt/config/generic/mmlu_pro.yaml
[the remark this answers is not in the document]
Kipok: Did you take these configs from somewhere or wrote yourself? If yourself, I would suggest to do another iteration optimizing few-shot format and not use the one with user/assistant tokens as I believe your current config is a bit inconsistent in using those tokens which might affect results. Let me know if you need help with this

nemo_skills/prompt/template/llama3-instruct-nosys-nobegin.yaml
```diff
@@ -0,0 +1,30 @@
+# Prompt specification for the original Llama3-instruct model, but skipping the system message
```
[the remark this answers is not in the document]
jubick1337: Meta [used](https://huggingface.co/datasets/meta-llama/Llama-3.1-8B-Instruct-evals/viewer/Llama-3.1-8B-Instruct-evals__mmlu_pro__details/latest) this template

nemo_skills/prompt/config/generic/mmlu_pro.yaml
jubick1337: I don't know which MMLU format is standard at the moment, I believe there are quite a lot of them.
I'm thinking to research on MMLU format later

nemo_skills/dataset/mmlu_pro/__init__.py
jubick1337: ok

nemo_skills/prompt/config/generic/mmlu_pro.yaml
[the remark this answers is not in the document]
jubick1337: I'm not sure which one we want to use.
I'll discuss which format we should keep with the other team

nemo_skills/prompt/config/generic/mmlu_pro.yaml
jubick1337: Yes, just checked with TigerLab script
```
Options: A. ST + TS is the identity map of V onto itself.\nB. TS = 0\nC. ST = 1\nD. ST - TS = 0\nE. ST = T\nF. ST = 0\nG. ST = TS\nH. ST - TS is the identity map of V onto itself.\nI. TS = T\nJ. ST = S\nAnswer:
```

nemo_skills/prompt/template/llama3-instruct-nosys-nobegin.yaml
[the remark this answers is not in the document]
jubick1337: Ok, I didn't know it, will delete that one

nemo_skills/prompt/config/generic/mmlu_pro.yaml
Kipok: ok, let's keep it for later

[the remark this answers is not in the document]
Kipok: ok, as long as it matches the reference prompts you should be good. But make sure it matches exactly, as any small inconsistency might lead to worse results. You can also add a test in here to make sure the prompt is what you expect https://github.com/NVIDIA/NeMo-Skills/blob/main/tests/test_prompts.py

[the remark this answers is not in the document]
Kipok: if you want to keep all, I guess that's also fine, but in this case please add comments to each yaml file for what are the cases when it should be used. But if you can remove some of them, that would be even better

requirements/datasets.txt
```diff
@@ -16,3 +16,4 @@ bs4
 evalplus @ git+https://github.com/evalplus/evalplus@c91370f
 pyyaml
 requests
+datasets
```
Kipok, line 19: Please add it to main.txt and remove this datasets.txt, it's no longer used

jubick1337: Done

nemo_skills/dataset/mmlu-pro/__init__.py
```diff
@@ -12,7 +12,16 @@
...
+PROMPT_CONFIG = 'generic/multichoice'
+DATASET_GROUP = 'multichoice'
+METRICS_TYPE = 'math'  # TODO: update this?
+DEFAULT_EVAL_ARGS = (
+    "++eval_type=mmlu "
+    "++eval_config.parse_func=llama "
+)
+
+
+DEFAULT_GENERATION_ARGS = ""
```
Kipok, line 27:
```suggestion
DEFAULT_GENERATION_ARGS = "++split=test_llama"
```
You could also add `++split=test_llama` to here so that it's used by default and matches the parse func

nemo_skills/dataset/mmlu-pro/__init__.py
```diff
@@ -12,7 +12,16 @@
...
 # limitations under the License.
 
-bs4
-evalplus @ git+https://github.com/evalplus/evalplus@c91370f
-pyyaml
-requests
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+
+PROMPT_CONFIG = 'generic/multichoice'
```
Kipok, line 18: should this be `generic/mmlu_pro` them? Or the one from llama folder?

Kipok:
please add mmlu-pro to tests/test_datasets.py

Kipok (approved):
Thanks for adding this!

PR #282, opened 2024-12-05
Fix code output tags detection

dgtm777:
Changed the regex to detect code-output blocks
Fix bug with code input blocks
Fix setting up code tags through command line


nemo_inspector/callbacks/__init__.py
```diff
@@ -97,8 +107,16 @@ def set_config(cfg: InspectorConfig) -> None:
         "examples_type": [UNDEFINED, RETRIEVAL] + examples_types,
         "retrieval_field": [""],
     }
-    conf_path = config['nemo_inspector']['prompt_config'] if config['nemo_inspector']['prompt_config'] else ""
-    template_path = config['nemo_inspector']['prompt_template'] if config['nemo_inspector']['prompt_config'] else ""
+    conf_path = (
```
Kipok, line 110: can you reuse `load_config` from prompt/utils?

Kipok (approved):

PR #288, opened 2024-12-11
Handle dependencies for tasks that are finished 

Kipok:
Previously if you use run_after with some experiment that's already finished, job submission would fail. Now it would raise a warning instead, since it's very common for resuming long workflows to start from a middle task. This allows us to skip modifying code to remove run_after arguments for such cases


titu1994 (approved):
Seems good!

PR #292, opened 2024-12-12
Improve random seed handling in generate

Kipok:
1. Add a new argument to specify list of seeds (useful for rerunning jobs that partially failed)
2. Add random_seed as formatting argument for pre/post process cmd to allow customization per job
3. Add pre/post process cmd to run_cmd to have unified interface with other scripts


gwarmstrong (approved):
LGTM

PR #293, opened 2024-12-12
Change llm judge to rm/generate interface

Kipok:


nemo_skills/inference/llm_math_judge.py
```diff
@@ -31,12 +31,24 @@
...
+    output_file: str | None = None  # Where to save the generations if `input_file` is provided
+    # Can specify an input directory, where the file will be inferred output.jsonl if no seed
+    # is provided, and output-rs{{seed}}.jsonl. This pattern is used to match the output files from
+    # the `generate` pipeline
+    input_dir: str | None = None
+    # Where to save the generations (with the identical file name) if `input_dir` is provided
+    output_dir: str | None = None
+    # Used to identify the input file if `input_dir` is provided. If `random_seed` is not provided,
+    # the input will be assumed to be from 'greedy' generation
+    random_seed: str | None = None
```
gwarmstrong, line 51: Why str? Int's should be nullable, but you might have to use `null` from cmdline instead of `None`. Perplexity example [here](https://www.perplexity.ai/search/can-you-have-optional-ints-tha-GRqF2xPaQme7_aj3p_ffTw#0).

gwarmstrong (approved):
Largely LGTM. One minor comment.

nemo_skills/inference/llm_math_judge.py
Kipok: just copied whatever was in the reward model script without checking. Let me update both

Kipok: on a second thought, let me keep it as is for now - I think eventually we should move this parameter out of here to the pipeline part

PR #296, opened 2024-12-16
MAINT revert some changes leading to job instability

gwarmstrong:
* adds back exclusive flag until a fix for overlapping jobs can be patched in
* adds gpus per node flag, assuming we generally care that allocated GPUs are on the same node. If not set, GPUs can be given from across multiple nodes.

### DIFF
--- a/nemo_skills/pipeline/utils.py
+++ b/nemo_skills/pipeline/utils.py
@@ -664,6 +664,18 @@ def get_executor(
     else:
         timeout = cluster_config["timeouts"][partition]
 
+    srun_args = [
+        "--no-container-mount-home",
+        "--overlap",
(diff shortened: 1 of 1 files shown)


Kipok (approved):

nemo_skills/pipeline/utils.py
```diff
@@ -684,6 +685,7 @@ def get_executor(
             # we need to be explicit about this in srun as commands might need to run in parallel
             f"--ntasks={tasks_per_node * num_nodes}",
             f"--nodes={num_nodes}",
+            f"--gpus-per-node={gpus_per_node}",
```
Kipok, line 688: Please add the same check as in line 679 here as some clusters don't support this parameter

PR #300, opened 2024-12-19
small fix for generic/default.yaml

fayejf:
should be problem instead question.

### DIFF
--- a/nemo_skills/prompt/config/generic/default.yaml
+++ b/nemo_skills/prompt/config/generic/default.yaml
@@ -2,4 +2,4 @@
 
 system: ""
 
-user: "{question}"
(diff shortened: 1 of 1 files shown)


Kipok:
It's actually expected to be "question" as that's the name used for benchmarks outside of math, e.g. code or chat benchmarks

fayejf:
I see. why not unify them?

Kipok:
We used to call it question but then switched to problem as that's what meta/qwen uses typically. In general, I don't think it's possible to have a single name that would be meaningful in all cases, so it's best to just have more specific prompts whenever adding new things and default is there as a way to quickly test things (plus for some chat benchmarks where it's appropriate)

PR #317, opened 2025-01-17
Alcray/autoform refactor

Alcray:


nemo_skills/pipeline/generate.py
```diff
@@ -34,11 +34,25 @@ class SupportedServers(str, Enum):
...
+        - `{output_base}-rs{seed}.jsonl` if `random_seed` is provided.
+    """
+    if output_base is not None:
+        if random_seed is not None:
+            output_file = f"{output_dir}/generation/{output_base}-rs{random_seed}.jsonl"
+        else:
+            output_file = f"{output_dir}/generation/{output_base}.jsonl"
     else:
-        output_file = f"{output_dir}/generation/output.jsonl"
+        # Fallback to original naming scheme if output_base is not provided
```
Kipok, line 50: let's just use "output" as default value and then we don't need else part?

nemo_skills/pipeline/generate.py
```diff
@@ -197,7 +211,11 @@ def generate(
         "Can provide an experiment name or an experiment object if running from code.",
     ),
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
-    log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs."),
+    log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs. "),
```
Kipok, line 214:
```suggestion
    log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs."),
```

nemo_skills/pipeline/generate.py
```diff
@@ -197,7 +211,11 @@ def generate(
...
+    output_base: str = typer.Option(
```
Kipok, line 215: maybe output_prefix? or output_base_name?

Kipok: Let's add it for other generation types as well, so that it's supported in all cases

nemo_skills/prompt/config/lean4/nat-to-lean4.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: What's the difference between this and the previous one?

Kipok (commented):
A few small things to fix. Is this good to be merged otherwise or are you planning more changes?

nemo_skills/code_execution/utils.py
```diff
@@ -64,3 +64,23 @@ def extract_code_to_execute(generation: str, code_begin: str, code_end: str, ext
...
     return _extract_between_separators(generation, [code_output_begin, code_output_end], extract_all)
+
+def extract_code_block(text: str, languages: list = ["lean4", "lean3", "lean", ""]) -> str:
+    for language in languages:
+        match = re.search(rf"```{language}\s*\n?(.*?)\n?```", text, re.DOTALL)
+        if match:
+            return match.group(1).strip()
+    return ""
+
+def clean_formal_generation(generation: str, languages: list = ["lean4", "lean3", "lean", ""]) -> str:
```
Kipok, line 75: languages can probably be hardcoded here. E.g. in line 85 you have them hardcoded anyway. And in general, using list as a default value is a bad practice because it's mutable

nemo_skills/pipeline/generate.py
```diff
@@ -297,6 +321,7 @@ def generate(
     ),
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
     log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs."),
+    output_base: str = typer.Option("output", help="Optional base name for output .jsonl files. If provided, will be used in place of 'output'."),
```
Kipok, line 324: output_prefix is probably a better name

Kipok (commented):
thanks! Just a few small suggestions

nemo_skills/pipeline/generate.py
Alcray: The idea is, that it fully replaces the "output" default name. So it's more just a base name, not a prefix I think. It could be called prefix for rs generations, but for greedy for example I think base might be better?

nemo_skills/code_execution/utils.py
Alcray: Done

Kipok:
Looks good, but please resolve the conflicts with your other pr that's now merged

Kipok (approved):

PR #318, opened 2025-01-17
added zero shot prompts for mmlu and mmlu-pro

jubick1337:


nemo_skills/prompt/config/generic/mmlu.yaml
```diff
@@ -0,0 +1,29 @@
+system: ""
```
Kipok, line 1: isn't this almost identical to https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/prompt/config/llama3-instruct/mmlu.yaml? Do we need both?

ekmb: @Kipok , Matvei will remove llama related prompt/configs, we're going to use tiger-lab prompt. This is to unblock other teams until final clean up is done.

Kipok: sounds good

PR #322, opened 2025-01-22
Add support for dependent libraries to attach their codebases

titu1994:
# Changelog

- Adds the ability and infrastructure for dependent libraries to attach their codebases onto the nemo skills repo during packaging
- This enables any level of nesting of dependent libraries


Kipok (approved):
Thanks! A bunch of commented lines to remove, but otherwise looks good

Kipok:
@titu1994 what's the status of this? Are we good to merge or are there some issues?

nemo_skills/pipeline/utils.py
```diff
@@ -46,6 +46,52 @@
 REUSE_CODE_EXP = {}
 
 
+@dataclass
+class RepoMetadata:
+    """Metadata for a repo that is used in the experiment."""
+
+    name: str
+    path: Path
```
titu1994, line 54: Im creating this dataclass because in the future, we may want to allow other repos to specifically add `include_patterns` and `include_patterns_relative_path` - for now, we just add the git level repo clone only (sufficient for my use case)

Kipok:
Lgtm, thanks Som!

PR #330, opened 2025-01-25
Support auto chunking of generate files

titu1994:
# Changelog
- Adds auto support of chunking for ns generate
- New args --num_chunks and  --chunk_ids (str)
- --chunk_ids can be comma separated list of .. notation "0..4" or "0,1,2,3,4"


Kipok:
Let's merge and add support to other scripts/cases in a follow up PR

PR #332, opened 2025-01-27
Add chunking support for eval and fix corner cases

Kipok:


Kipok:
will add postprocessing logic to merge chunks in another pr

nemo_skills/pipeline/eval.py
```diff
@@ -190,34 +226,43 @@ def eval(
...
         ]
         if not skip_greedy
         else []
     )
     eval_cmds += [
-        get_sampling_cmd(
+        cmd
+        for benchmark, rs_num in benchmarks.items()
+        for rs in range(starting_seed, starting_seed + rs_num)
+        for cmd in get_sampling_cmd(
```
titu1994, line 249: Its a bit confusing what exactly this list is going to look like. Maybe better to use normal triple loop for clarity.

nemo_skills/pipeline/eval.py
```diff
@@ -190,34 +226,43 @@ def eval(
...
             extra_datasets=extra_datasets,
+            num_chunks=num_chunks,
+            chunk_ids=chunk_ids,
         )
-        for benchmark, rs_num in benchmarks.items()
-        for rs in range(starting_seed, starting_seed + rs_num)
     ]
     if num_jobs == -1:
         num_jobs = len(eval_cmds)
+    else:
```
titu1994, line 263: Yes its generally the final number of jobs being launched

nemo_skills/utils.py
```diff
@@ -265,24 +270,55 @@ def chunk_data(data: List[Any], output_filename: str, chunk_id: Optional[int], n
...
-        end_idx = (chunk_id + 1) * chunk_size if (chunk_id < num_chunks - 1) else len(data)
+        remainder = len(data) % num_chunks
+        base_size = len(data) // num_chunks
+
+        extra = 1 if chunk_id < remainder else 0
+
+        if chunk_id < remainder:
+            start_idx = chunk_id * (base_size + 1)
+        else:
+            start_idx = remainder * (base_size + 1) + (chunk_id - remainder) * base_size
```
titu1994, line 281: Why is this logic necessary vs the earlier one?

titu1994 (approved):
Overall looks quite good, though i dunno when eval will need to chunk. It will also break logic for summarize function

nemo_skills/inference/generate.py
```diff
@@ -353,17 +349,6 @@ def generate(cfg: GenerateSolutionsConfig):
             f"Number of samples in the chunk: {len(data)}"
         )
```
titu1994, line 351: Were dropping offset support?

Kipok: did you use it? I just think it was never used, so decided to remove, but if you're using it, we can keep. It's just getting hard to use properly with all the extra settings we add

nemo_skills/pipeline/eval.py
Kipok: Sounds good, let me do it in a follow-up PR (the part probably needs a larger refactoring)

nemo_skills/pipeline/eval.py
Kipok: The thing here is we have num_jobs which is distributing benchmarks / samples on different nodes and now also chunks, which split a single benchmark across different nodes (and we restrict the same chunks for each benchmark, otherwise it's too many options). So currently we will have num_jobs * num_chunks total jobs but it's a bit confusing, so maybe some renaming or some other changes are needed

nemo_skills/utils.py
Kipok: In your previous way, the last chunk might be much bigger than previous chunks, e.g. if `len(data) % num_chunks = num_chunks - 1` and num_chunks is sufficiently big. But in the new way, we would always distribute remainder uniformly across all chunks, thus keeping the distribution more balanced

Kipok:
> Overall looks quite good, though I dunno when eval will need to chunk. It will also break logic for summarize function
Mainly for running R1 evals, since it's often impractical to keep it on a single node. But generally the reasoning models take much longer since sequence length is so big and it's only going to grow in the future. Summarize results will work out-of-the-box after we add proper automatic post-processing (in a follow-up PR)

PR #333, opened 2025-01-28
Move async position into data dict for valid intermediate jsonl

titu1994:
# Changelog
- Moves the async position field into the data dictionary with a private key, so that the intermediate json is still valid


Kipok (approved):
thanks!

PR #334, opened 2025-01-28
Switch to vllm official docker

Kipok:


nemo_skills/pipeline/utils.py
```diff
@@ -239,7 +239,7 @@ def get_server_command(
             num_tasks = 1
     elif server_type == 'vllm':
         start_vllm_cmd = (
-            f"python -m nemo_skills.inference.server.serve_vllm "
+            f"python3 -m nemo_skills.inference.server.serve_vllm "
```
shtoshni, line 242: https://github.com/NVIDIA/NeMo-Skills/blob/f1d138cee3e9714b358f7e11047da418e1b3446c/nemo_skills/pipeline/utils.py#L186

python -> python3 here as well

Kipok:
ci is broken again, but works fine locally

PR #337, opened 2025-01-29
Support the ability to pass python dict to skills

titu1994:
# Changelog
- Allows passing a python dict isntead of a str filepath to the cluster config


Kipok (approved):
thanks!

PR #338, opened 2025-01-29
Refactor generate.py into modular extensible class

titu1994:
# Changelog
- Refactors generate.py into a class that implements the Template pattern
- Refactors pipeline/generate.py to enable passing (via ctx.obj) a class that inherits GenerationTask - and thus enables overriding the script path to a custom one for generation of custom types.
- Subclass + override of template methods / hooks should enable more code reuse of generate.py / pipeline without having to copy paste the entire script


shtoshni:
@titu1994 - Can you separate sync vs. async generations in this PR. Maybe a separate subclass for the two. Happy to help out if you think it's a good idea.

shtoshni:
Feel free to merge it for now, @titu1994 
I can make my incremental changes after this merge. 
The PR might need some work to merge with main though due to a recent PR.

PR #339, opened 2025-01-29
Handling result summary error when directory structure is not as expected

shtoshni:


nemo_skills/pipeline/summarize_results.py
```diff
@@ -109,6 +109,9 @@ def summarize_results(
     eval_results_dir = Path(results_dir) / 'eval-results'
     if eval_results_dir.exists() and eval_results_dir.is_dir():
         results_dir = eval_results_dir
+    else:
+        raise ValueError(f"No 'eval-results' dir found in {results_dir}")
```
Kipok, line 113: that's not the right way to do it I think. The else part might normally happen if you provide a path to eval-results itself, so we shouldn't throw an error here. Instead what we should check is whether the dirname corresponds to one of the benchmarks we know how to compute metrics for and whether we have any output files there

shtoshni: Ok, so there are three options possible.

1. `{results_dir}/eval-results/{benchmark}/output*jsonl`
2. `{results_dir}/{benchmark}/output*jsonl`
3. `{results_dir}/output*jsonl`

So we can support all three options in this PR. Is that what you're recommending above?

Kipok: Yes, all 3 options are possible. The first two should check if benchmark is something we have metrics for (unless metrics_type is specified), and the third option should check if metrics type is explicitly provided. And in all cases, we should have a clear error if there are no output*.jsonl inside any of the folders

nemo_skills/pipeline/summarize_results.py
```diff
@@ -105,15 +105,50 @@ def summarize_results(
...
+
+    # Check for Option 3 - Root directory corresponds to a benchmark
+    if Path(results_dir).is_dir() and len(glob.glob(f'{results_dir}/output*jsonl')) > 0:
+        benchmarks_paths = [results_dir]
+    else:
+        cand_results_dir = Path(results_dir) / 'eval-results'
+        # Check for Option 1
+        if cand_results_dir.exists() and cand_results_dir.is_dir():
+            results_dir = cand_results_dir
+        elif Path(results_dir).name == 'eval-results':
```
Kipok, line 125: Let's remove this check - it doesn't have to be named "eval-results" as you can run this on the output of the judge pipeline or generate pipeline

shtoshni: How about replacing this with:
`elif len(glob.glob(f'{results_dir}/*/output*jsonl')) > 0:`

This way, you can supply `/.../eval-results` and `/.../eval-results-judged` as direct inputs.

nemo_skills/pipeline/summarize_results.py
```diff
@@ -105,15 +105,51 @@ def summarize_results(
...
+
+    # Check for Option 3 - Root directory corresponds to a benchmark
+    if Path(results_dir).is_dir() and len(glob.glob(f'{results_dir}/output*jsonl')) > 0:
+        benchmarks_paths = [results_dir]
+    else:
+        cand_results_dir = Path(results_dir) / 'eval-results'
+        # Check for Option 1
+        if cand_results_dir.exists() and cand_results_dir.is_dir():
+            results_dir = cand_results_dir
+        elif len(glob.glob(f'{results_dir}/*/output*jsonl')) > 0:
```
Kipok, line 125: I think this condition is just not needed. It will fail right below when checking individual folders with a clear message anyway. But if you want to keep it, please change to == 0 and raise an error, as current `result_dir = result_dir` is redundant and a bit confusing

shtoshni: Ahh, fair point. Have changed the logic now.

Kipok (approved):
lgtm as long as you verified that it works in all 3 cases (we don't have tests that check this)

PR #340, opened 2025-01-30
Error handling

shtoshni:
Reward model calls can fail for a variety of reasons, such as the input sequence going beyond the model's max length. This PR adds error handling which was missing.

### DIFF
--- a/nemo_skills/inference/server/reward_model.py
+++ b/nemo_skills/inference/server/reward_model.py
@@ -118,10 +118,28 @@ class VLLMRewardModel(BaseModel):
         self.model = model_list.data[0].id
 
     def _score_single_prompt(self, prompt):
-        response = requests.post(self.request_url, json={"input": prompt, "model": self.model})
-        per_token_scores = response.json()['data'][0]['data']
-        last_token_score = per_token_scores[-1]
+        """Score a single prompt"""
 
+        per_token_scores = None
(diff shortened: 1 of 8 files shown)


nemo_skills/inference/server/reward_model.py
```diff
@@ -118,10 +118,28 @@ def __init__(self, **kwargs):
...
-        last_token_score = per_token_scores[-1]
+        """Score a single prompt"""
+
+        per_token_scores = None
+        try:
+            response = requests.post(self.request_url, json={"input": prompt, "model": self.model})
+            output = response.json()
+            per_token_scores = output['data'][0]['data']
+        except requests.exceptions.HTTPError as err:
+            print(f"Request failed: {err}")
```
Kipok, line 129: Please use LOG.warning instead

Kipok: I think it would be good to add the error as an explicit field in the output, maybe with `inference_error` field or something similarly named. That way it will be easier to understand why the score is 0 and it increases the chance that users notice that perhaps a lot of their requests have failed (not everyone is checking the logs)

shtoshni: Made the suggested changes and ran the GPU test locally. 
Will commit the change if the CPU tests pass

PR #344, opened 2025-01-30
Support batch processing aync inference

wedu-nvidia:
Currently, the async inference loop encounters the following error when processing large-scale datasets:

RuntimeError: can't start new thread

To address this issue and support large-scale datasets, I have updated the async loop logic to submit requests in batches. For example, we initially submit 1,000 requests, and as soon as some requests complete, we continue adding new ones to maintain a steady count of 1,000 active requests in the inference engine. This ensures efficient resource utilization without exceeding thread limitations.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -261,7 +261,7 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
     LOG.warning(f"Async loop is maintaining {cfg.max_concurrent_requests} concurrent requests throughout execution -- batch_size parameter is ignored!.")
 
     
-    request_queue = list(range(len(data)))  # Queue of unsubmitted task indices
+    request_queue = deque(range(len(data)))  # Queue of unsubmitted task indices
     in_progress = {}  # Track ongoing requests {index: generation_id}
     
 
@@ -271,7 +271,7 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
 
         # **Step 1: Pre-fill `in_progress` with the first max_concurrent_requests request**
         while len(in_progress) < cfg.max_concurrent_requests and request_queue:
-            idx = request_queue.pop(0)
+            idx = request_queue.popleft()
             in_progress[idx] = None  # Placeholder for reservation
 
         # **Step 2: Submit first max_concurrent_requests requests as a batch**
@@ -300,7 +300,10 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
                     # Mark task as completed
                     completed_tasks.append(idx)
                     del in_progress[idx]
-
+                    ### we can remove the gen_id from the gen_id_to_params to save memory when we are done with it
+                    if gen_id in llm.gen_id_to_params:
+                        del llm.gen_id_to_params[gen_id]
+                        
                     # Prepare the result for writing
                     gen_dict[cfg.generation_key] = gen_dict.pop("generation")
                     gen_dict.update(data[idx])
@@ -317,9 +320,12 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
 
 
             
-            # **Step 4: Refill requests to maintain exactly N concurrent tasks**
-            num_to_submit = cfg.max_concurrent_requests - len(in_progress)
-            batch_indices = [request_queue.pop(0) for _ in range(min(num_to_submit, len(request_queue)))]
+            # **Step 4: Refill requests to maintain exactly N=max_concurrent_requests concurrent tasks**
+            # num_to_submit = cfg.max_concurrent_requests - len(in_progress)
+            # batch_indices = [request_queue.popleft() for _ in range(min(num_to_submit, len(request_queue)))]
+            num_to_submit = min(cfg.max_concurrent_requests - len(in_progress), len(request_queue))
+            batch_indices = [request_queue.popleft() for _ in range(num_to_submit)]
+
             batch_prompts = [prompt.fill(data[idx]) for idx in batch_indices]
             if len(batch_prompts) > 0:
                 old_gen_id_to_params = llm.gen_id_to_params.copy()
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -67,7 +67,7 @@ class GenerateSolutionsConfig:
     max_samples: int = -1  # If > 0, will stop after generating this many samples. Useful for debugging
     skip_filled: bool = False  # If True, will skip the generations that are already in the output file
 
-    max_concurrent_requests: int = 128  # Maximum number of concurrent requests to the server for the async loop
+    max_concurrent_requests: int = 1024  # Maximum number of concurrent requests to the server for the async loop
     # chunk the dataset into equal sized parts and index into them
     num_chunks: int | None = None  # if specified, will split the data into chunks and only generate for one chunk
     chunk_id: int | None = None  # if specified, will index the specified chunk only
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -270,33 +270,31 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
     with open(cfg.output_file + "-async", "at" if cfg.skip_filled else "wt", encoding="utf-8", buffering=1) as fout:
         pbar = tqdm(total=len(data), desc="Processing requests")
 
-        # **Step 1: Pre-fill `in_progress` with the first max_concurrent_requests request**
-        while len(in_progress) < cfg.max_concurrent_requests and request_queue:
-            idx = request_queue.popleft()
-            in_progress[idx] = None  # Placeholder for reservation
-
-        # **Step 2: Submit first max_concurrent_requests requests as a batch**
-        batch_indices = list(in_progress.keys())  # Get reserved indices
-        batch_prompts = [prompt.fill(data[idx]) for idx in batch_indices]  # Prepare prompts
-
-        batch_ids = llm.generate_async(  # Submit all at once
-            prompts=batch_prompts,  
-            stop_phrases=combine_stop_phrases(prompt.stop_phrases, extra_stop_phrases),
-            **asdict(cfg.inference),
-            **extra_generate_params,
-        )
-
-        # Store the request IDs returned by LLM
-        for batch_idx, request_idx in enumerate(batch_indices):
-            in_progress[request_idx] = batch_ids[batch_idx]
-
-
-        # **Step 3: Monitor and refill requests dynamically**
+        
         while in_progress or request_queue:  # Continue until all tasks are complete
-            remaining_ids = {idx: gen_id for idx, gen_id in in_progress.items()}
-            generations = llm.get_generations(list(remaining_ids.values()))
             
-            for (idx, gen_id), gen_dict in zip(remaining_ids.items(), generations):
+            # Dynamic sending requests to maintain cfg.max_concurrent_requests running requests
+            num_to_submit = min(cfg.max_concurrent_requests - len(in_progress), len(request_queue))
+            batch_indices = [request_queue.popleft() for _ in range(num_to_submit)]
+
+            batch_prompts = [prompt.fill(data[idx]) for idx in batch_indices]
+            if len(batch_prompts) > 0:
+               
+                generation_ids = llm.generate_async(
+                    prompts=batch_prompts,  
+                    stop_phrases=combine_stop_phrases(prompt.stop_phrases, extra_stop_phrases),
+                    **asdict(cfg.inference),
+                    **extra_generate_params,)
+             
+            ### map the generated ids to the original positions
+            for gen_ids_idx, original_pos in enumerate(batch_indices):
+                in_progress[original_pos] = generation_ids[gen_ids_idx]
+            
+            ### create a snapshot of in_progress to avoid modifying the dictionary while iterating over it
+            snapshot_in_progress = in_progress.copy() 
+            generations = llm.get_generations(list(snapshot_in_progress.values()))
+            
+            for (idx, gen_id), gen_dict in zip(snapshot_in_progress.items(), generations):
                 if gen_dict['generation'] is not None:
                     # remove the completed task from in_progress
                     del in_progress[idx]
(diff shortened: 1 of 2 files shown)


nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            **extra_generate_params,
+        )
+
+        # Store the request IDs returned by LLM
+        for i, idx in enumerate(batch_indices):
+            in_progress[idx] = batch_ids[i]
+
+        # **Step 3: Monitor and refill requests dynamically**
+        while in_progress or request_queue:  # Continue until all tasks are complete
+            remaining_ids = {idx: gen_id for idx, gen_id in in_progress.items() if gen_id is not None and gen_id in llm.gen_id_to_params}
```
Kipok, line 294: why do you need to check for gen_id in llm.gen_id_to_params?

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            generations = llm.get_generations(list(remaining_ids.values()))
+
+            lost_tasks = [idx for idx in in_progress if in_progress[idx] is not None and in_progress[idx] not in llm.gen_id_to_params]
```
Kipok, line 297: what are lost_tasks?

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            completed_tasks = []
+            for (idx, gen_id), gen_dict in zip(remaining_ids.items(), generations):
+                if gen_dict['generation'] is not None:
+                    # Mark task as completed
+                    completed_tasks.append(idx)
+                    in_progress[idx] = None  
+
+                    # Prepare the result for writing
+                    gen_dict[cfg.generation_key] = gen_dict.pop("generation")
+                    gen_dict.update(data[idx])
```
Kipok, line 313: Please keep the original loop to pop all keys, not just generation

nemo_skills/inference/generate.py
```diff
@@ -67,6 +67,7 @@ class GenerateSolutionsConfig:
     max_samples: int = -1  # If > 0, will stop after generating this many samples. Useful for debugging
     skip_filled: bool = False  # If True, will skip the generations that are already in the output file
 
+    max_concurrent_requests: int = 128  # Maximum number of concurrent requests to the server for the async loop
```
Kipok, line 70:
```suggestion
    max_concurrent_requests: int = 1024  # Maximum number of concurrent requests to the server for the async loop
```
I think 1024 is a good default to utilize most engines. 128 is too small

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
     if len(data) == 0:  # we might not have any examples if skip_filled=True
         return
 
-    LOG.warning("Async loop is submitting all data for inference - batch_size parameter is ignored!")
+    LOG.warning("Maintaining 1000 concurrent requests throughout execution.")
```
```suggestion
    LOG.warning("Maintaining %s concurrent requests throughout execution. Change by setting `max_concurrent_requests`. Batch size parameter is ignored.", cfg.max_concurrent_requests)
```

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
-                        data[gen_pos].pop(key, None)
-                    gen_dict.update(data[gen_pos])
 
-                    # insert async position information
-                    gen_dict[cfg.async_position_key] = original_positions[gen_pos]
+    with open(cfg.output_file + "-async", "at" if cfg.skip_filled else "wt", encoding="utf-8", buffering=1) as fout:
+        pbar = tqdm(total=len(data), desc="Processing requests")
 
-                    fout.write(json.dumps(gen_dict) + "\n")
+        # **Step 1: Pre-fill `in_progress` with the first 2000 tasks**
```
```suggestion
        # **Step 1: Pre-fill `in_progress` with the first N tasks**
```

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+        while len(in_progress) < cfg.max_concurrent_requests and request_queue:
+            idx = request_queue.pop(0)
```
Kipok, line 274: if you want to use this .pop / .append interface, better to switch to a real dequeue from python as popping a list is very inefficient

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            in_progress[idx] = None  # Placeholder for reservation
 
+        # **Step 2: Submit first 1000 requests as a batch**
```
```suggestion
        # **Step 2: Submit first N requests as a batch**
```

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+
+        batch_ids = llm.generate_async(  # Submit all at once
+            prompts=batch_prompts,  
+            stop_phrases=combine_stop_phrases(prompt.stop_phrases, extra_stop_phrases),
+            **asdict(cfg.inference),
+            **extra_generate_params,
+        )
+
+        # Store the request IDs returned by LLM
+        for i, idx in enumerate(batch_indices):
```
Kipok, line 289:
```suggestion
        for idx in batch_indices:
```

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            in_progress[idx] = batch_ids[i]
```
Kipok, line 290:
```suggestion
            in_progress[idx] = idx
```

nemo_skills/inference/generate.py
```diff
@@ -257,48 +258,91 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+
+                    # Update progress bar
+                    pbar.update(1)
+
+            # Remove completed tasks from `in_progress`
+            for idx in completed_tasks:
+                del in_progress[idx]
+
+            
+            # **Step 4: Refill requests to maintain exactly 1000 concurrent tasks**
```
Kipok, line 329:
```suggestion
            # **Step 4: Refill requests to maintain exactly N concurrent tasks**
```

Kipok (changes requested):

nemo_skills/inference/generate.py
wedu-nvidia: Sometimes, it throws a KeyError, possibly because the LLM engine lost some gen_ids. To handle this, I collect the missing indices of these keys, reinsert them into the queue, and resubmit them to the LLM engine.

nemo_skills/inference/generate.py
wedu-nvidia: See comments above

nemo_skills/inference/generate.py
Kipok: It cannot lose gen ids, they are handled on our side. If you see this, it means there is some bug in the code that's causing this (basically you're querying for the positions that are already removed)

wedu-nvidia:
I updated the code as suggested above

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            
+            # **Step 4: Refill requests to maintain exactly N=max_concurrent_requests concurrent tasks**
+            # num_to_submit = cfg.max_concurrent_requests - len(in_progress)
+            # batch_indices = [request_queue.popleft() for _ in range(min(num_to_submit, len(request_queue)))]
+            num_to_submit = min(cfg.max_concurrent_requests - len(in_progress), len(request_queue))
+            batch_indices = [request_queue.popleft() for _ in range(num_to_submit)]
+
+            batch_prompts = [prompt.fill(data[idx]) for idx in batch_indices]
+            if len(batch_prompts) > 0:
+                old_gen_id_to_params = llm.gen_id_to_params.copy()
```
Kipok, line 334: Please move this logic to inside of the BaseModel - we want it to always work correctly without users having to make similar changes on their side

nemo_skills/inference/generate.py
Kipok: Why not make this change? Seems much easier to read to me, but if you have some reason to prefer the current way please share it

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
     if len(data) == 0:  # we might not have any examples if skip_filled=True
         return
 
-    LOG.warning("Async loop is submitting all data for inference - batch_size parameter is ignored!")
+    LOG.warning(f"Async loop is maintaining {cfg.max_concurrent_requests} concurrent requests throughout execution -- batch_size parameter is ignored!.")
```
Kipok, line 262: Let's add to the message something about which parameter users can change to control number of concurrent requests as it's not obvious what to do if they want a change

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+
+        batch_ids = llm.generate_async(  # Submit all at once
+            prompts=batch_prompts,  
+            stop_phrases=combine_stop_phrases(prompt.stop_phrases, extra_stop_phrases),
+            **asdict(cfg.inference),
+            **extra_generate_params,
+        )
+
+        # Store the request IDs returned by LLM
+        for i, idx in enumerate(batch_indices):
```
shtoshni, line 290: I would suggest naming the variables a bit more descriptive than `i` and `idx` for sake of clarity in the `for` loop.

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            in_progress[idx] = batch_ids[i]
+
+        # **Step 3: Monitor and refill requests dynamically**
+        while in_progress or request_queue:  # Continue until all tasks are complete
+            remaining_ids = {idx: gen_id for idx, gen_id in in_progress.items()}
+            generations = llm.get_generations(list(remaining_ids.values()))
+            
+            completed_tasks = []
```
shtoshni, line 298: The `completed_tasks`  variable doesn't seem to be doing anything. Can get rid of it.

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+            for (idx, gen_id), gen_dict in zip(remaining_ids.items(), generations):
+                if gen_dict['generation'] is not None:
+                    # Mark task as completed
+                    completed_tasks.append(idx)
+                    del in_progress[idx]
+                    ### we can remove the gen_id from the gen_id_to_params to save memory when we are done with it
```
shtoshni, line 304:
```suggestion
                    # We can remove the gen_id from the gen_id_to_params to save memory when we are done with it
```

nemo_skills/inference/generate.py
wedu-nvidia: Yes, I will remove

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,92 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+
+
+            
+            # **Step 4: Refill requests to maintain exactly N=max_concurrent_requests concurrent tasks**
+            # num_to_submit = cfg.max_concurrent_requests - len(in_progress)
+            # batch_indices = [request_queue.popleft() for _ in range(min(num_to_submit, len(request_queue)))]
+            num_to_submit = min(cfg.max_concurrent_requests - len(in_progress), len(request_queue))
+            batch_indices = [request_queue.popleft() for _ in range(num_to_submit)]
+
+            batch_prompts = [prompt.fill(data[idx]) for idx in batch_indices]
```
shtoshni, line 332: Right now the prompt_fill and generation part are repeated before the while loop and at the end of the while loop. I would suggest removing this duplication by moving this piece of code to start of the while loop.

nemo_skills/inference/generate.py
wedu-nvidia: sure

nemo_skills/inference/generate.py
shtoshni: I meant the piece of code is repetitive.

nemo_skills/inference/generate.py
wedu-nvidia: Updated for base and trtllm

nemo_skills/inference/generate.py
wedu-nvidia: will do

nemo_skills/inference/server/model.py
```diff
@@ -155,18 +155,28 @@ def generate_async(
...
+        new_gen_id_to_future = {}  # Stores futures for the current batch
+        new_gen_id_to_params = {}  # Stores generation parameters for the current batch
+
+        # Generate unique IDs for each future and store them
         for future in futures:
-            gen_id = str(uuid.uuid4())
+            gen_id = str(uuid.uuid4())  # Generate a unique generation ID
             gen_ids.append(gen_id)
-            self.gen_id_to_future[gen_id] = future
+            new_gen_id_to_future[gen_id] = future  # Store the future in the dictionary
```
shtoshni, line 167: This all can be done in the above for loop itself, right?

wedu-nvidia: Yes, Igor suggested to keep all update in base class. So I modify here to avoid manually track.

nemo_skills/inference/server/model.py
```diff
@@ -155,18 +155,28 @@ def generate_async(
             request['prompt'] = prompts[request_idx]
             self.preprocess_request(request)
             futures.append(self.executor.submit(self._generate_single, **request))
-
+        
         gen_ids = []
+        new_gen_id_to_future = {}  # Stores futures for the current batch
```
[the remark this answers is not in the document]
wedu-nvidia: yes, we can I will update then.

nemo_skills/inference/server/model.py
```diff
@@ -155,18 +155,28 @@ def generate_async(
...
         for future in futures:
-            gen_id = str(uuid.uuid4())
+            gen_id = str(uuid.uuid4())  # Generate a unique generation ID
             gen_ids.append(gen_id)
-            self.gen_id_to_future[gen_id] = future
+            new_gen_id_to_future[gen_id] = future  # Store the future in the dictionary
 
-        self.gen_id_to_params = {
+        # Construct new_gen_id_to_params mapping gen_id to stop_phrases and remove_stop_phrases
+        new_gen_id_to_params = {
```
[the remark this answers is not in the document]
wedu-nvidia: Ok, got it

nemo_skills/inference/generate.py
```diff
@@ -257,48 +259,87 @@ def async_loop(cfg, data, llm, prompt, extra_stop_phrases, extra_generate_params
...
+        )
 
+        # Store the request IDs returned by LLM
+        for batch_idx, request_idx in enumerate(batch_indices):
+            in_progress[request_idx] = batch_ids[batch_idx]
+
+
+        # **Step 3: Monitor and refill requests dynamically**
+        while in_progress or request_queue:  # Continue until all tasks are complete
+            remaining_ids = {idx: gen_id for idx, gen_id in in_progress.items()}
```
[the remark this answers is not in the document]
wedu-nvidia: I will keep updating in_progress[idx]
so I create a copy for in_progress[idx], otherwise it will report errors. 
I will rename it for better clarity

wedu-nvidia:
address all comments with new update, thanks so much for useful comments. Tested again on cluster for vllm and trtllm.

shtoshni:
Thanks @wedu-nvidia 
I made some small optimizations to the model file.

shtoshni:
Added Igor's suggestion from - https://github.com/NVIDIA/NeMo-Skills/pull/344#discussion_r1938105747

Kipok (approved):
Looks good, thanks everyone!

PR #347, opened 2025-02-03
Adding the option of including generation in generate.py

shtoshni:
There are cases like gen-rm, generation summarization where we want to include the "generation" key while building the prompt. This PR adds support for such use cases.


Kipok:
I think you might have missed another prompt.fill in the generate, the one we use to print the first prompt. It might fail if you require generation key but don't set include_generation=True

PR #350, opened 2025-02-04
Generate donefiles and merge chunks

gwarmstrong:
This PR creates done files for generation jobs that can later be used to skip finished generation jobs. Furthermore, this adds support for merging chunked generation jobs.


nemo_skills/inference/merge_chunks.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
Kipok, line 1: why not do it in python? Much easier to check things, understand what's happening and modify logic. You can still call system cat from python

nemo_skills/pipeline/generate.py
```diff
@@ -239,6 +246,20 @@ def generate(
...
 
+    chunking_enabled = (num_chunks is not None) or (chunk_ids is not None)
+    if chunking_enabled and postprocess_cmd:
+        logging.warning(
+            "Chunking is enabled, but postprocess_cmd is also specified. "
+            "Note that will be run for each chunk separately. Chunk merging "
+            "will be performed after postprocess_cmd."
+        )
+    if chunking_enabled and generation_type != GenerationType.generate:
+        logging.warning(
```
Kipok, line 257: Let's finish with an error then?

nemo_skills/pipeline/generate.py
```diff
@@ -272,8 +293,14 @@ def generate(
     with run.Experiment(expname) as exp:
         extra_arguments_original = extra_arguments
         # TODO: reduce code duplication
+        donefiles = defaultdict(list)
         if random_seeds:
             for chunk_id in chunk_ids:
+                for seed in random_seeds:
+                    single_output_dir = f"{output_dir}{'/generation' if generation_type == GenerationType.generate else ''}"
```
Kipok, line 300: how about we unify this? Let's always use top-level without "generation" subfolder as it's more convenient and the fact that currently folder structure is different is confusing

nemo_skills/pipeline/generate.py
```diff
@@ -297,13 +324,26 @@ def generate(
                     )
                     prev_tasks = None
 
+                    single_output_dir = f"{output_dir}{'/generation' if generation_type == GenerationType.generate else ''}"
```
Kipok, line 327: can we move all this logic to be inside get_cmd?

Kipok (commented):
One other logic that we should add is to check for all done files on the cluster and not submit any jobs that are finished already. But we can do that in another PR

nemo_skills/pipeline/generate.py
gwarmstrong: one thing I find valuable about the current approach is--if I want to judge or reward model, I can put them all their subfolders in the same folder, e.g.,

```bash
experiment-step-01
     generation
     judgement
     reward
```

whereas, with the proposal, I would have to do something like:
```bash
experiment-step-01
    generate
        generation
    judgement
        generation
    reward
        generation
```

So the alternate proposal is to name the output directory of generation based on the generation type. I'm fine either way, ultimately. wdyt?

nemo_skills/pipeline/generate.py
gwarmstrong: Fixed in 3d8a20a

nemo_skills/pipeline/generate.py
Kipok: why? What I propose is to remove the "generation", not to add it. Then you can still do what you want with subfolders, right?

gwarmstrong: Oh okay yeah that works. Initially interpreted the making them all the same as with current "/generation". Much more clear now and agree.

gwarmstrong: fixed in [75fa210](https://github.com/NVIDIA/NeMo-Skills/pull/350/commits/75fa210c8c134a7f2b8b0be5048e0fc0b017ee24)

nemo_skills/inference/merge_chunks.sh
gwarmstrong: converted in [c567f63](https://github.com/NVIDIA/NeMo-Skills/pull/350/commits/c567f63e714787f5fd08c61734f7d70aeff0ca04)

Kipok, line 1: Remove this one?

nemo_skills/pipeline/generate.py
gwarmstrong: added in 04c13d7

nemo_skills/inference/merge_chunks.sh
gwarmstrong: fixed in [27039e0](https://github.com/NVIDIA/NeMo-Skills/pull/350/commits/27039e00b40e95747efebaea4d99914fbaf7b1ab)

nemo_skills/pipeline/generate.py
```diff
@@ -272,8 +306,13 @@ def generate(
     with run.Experiment(expname) as exp:
         extra_arguments_original = extra_arguments
         # TODO: reduce code duplication
+        donefiles = defaultdict(list)
         if random_seeds:
             for chunk_id in chunk_ids:
+                for seed in random_seeds:
```
Kipok, line 312: Do we need this logic here? Why not do everything inside get_cmd, so that you don't need to pass any extra arguments there?

PR #353, opened 2025-02-05
Added GPQA dataset and minor fixes to MMLU-Pro prompt

ekmb:


nemo_skills/evaluation/evaluator.py
```diff
@@ -433,6 +433,7 @@ def eval_lean4(cfg):
     'answer_judgement': dummy_eval,
     'lean4': eval_lean4,
     'mmlu-pro': eval_mmlu_pro,
+    'gpqa': eval_mmlu_pro,
```
Kipok, line 436: if it's the same, let's just use mmlu-pro in the default args? Maybe rename it to something more generic, I guess if it's reusable?

nemo_skills/prompt/config/generic/gpqa-zero-shot.yaml
```diff
@@ -0,0 +1,6 @@
+# Prompt used by TigerLab for MMLU-Pro benchmark
```
Kipok, line 1: Comment needs update?

nemo_skills/prompt/config/generic/gpqa-zero-shot.yaml
```diff
@@ -0,0 +1,6 @@
...
+# https://github.com/TIGER-AI-Lab/MMLU-Pro/blob/main/evaluate_from_api.py
+
+user: |-
```
Kipok, line 4: Does it make a big difference if we use this vs mmlu? Don't mind having an extra prompt if it's important, but if we can unify them, it's always better

nemo_skills/dataset/gpqa/__init__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: Please add gpqa in here https://github.com/NVIDIA/NeMo-Skills/blob/main/tests/test_datasets.py#L20

Kipok (changes requested):
Thanks, just a couple of small changes

nemo_skills/dataset/gpqa/__init__.py
ekmb: done

nemo_skills/evaluation/evaluator.py
```diff
@@ -432,7 +432,9 @@ def eval_lean4(cfg):
     'mt-bench': eval_mtbench,
     'answer_judgement': dummy_eval,
     'lean4': eval_lean4,
-    'mmlu-pro': eval_mmlu_pro,
+    'mmlu-pro': eval_mcq,
```
Kipok, line 435: Can we instead have a single eval type called 'mcq' and use it in directly in all relevant benchmarks?

Kipok (approved):
thanks!

PR #354, opened 2025-02-05
shtoshni/include generation fix

shtoshni:
Renamed the `include_generation` to `prefix_generation_to_response` variable and added it in a missing place.


nemo_skills/inference/generate.py
```diff
@@ -395,7 +399,11 @@ def generate(cfg: GenerateSolutionsConfig):
         LOG.info(
             "Example prompt (first turn only):\nData dictionary: %s\nPrompt: %s",
             first_sample,
-            prompt.fill(first_sample, multi_turn_key=cfg.multi_turn_key),
+            prompt.fill(
```
Kipok, line 402: There is another prompt.fill right above this one - please update it as well. Would be good to do search for prompt.fill and update all to make sure nothing else is missing

Kipok (changes requested):

nemo_skills/inference/generate.py
shtoshni: Good catch. There were 5 calls to prompt.fill in inference/generate.py and all 5 have this argument added now.

Kipok (approved):

PR #355, opened 2025-02-05
Change time calculation to DD:HH:MM:SS format

titu1994:
# Changelog
- Support DD:HH:MM:SS format for sft timer


Kipok:
Please update the logic here accordingly https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/openrlhf/sft.py#L174. Ideally by moving that to a function and reusing with PPO script and also with nemo train over here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/train.py#L132

Kipok (approved):

PR #358, opened 2025-02-07
OpenRLHF PPO save checkpoints on timeout

gwarmstrong:
This MR adds the ability to save checkpoints after the first global step that passes after a timeout is reached.

## Changelog

* Added `nemo_skills/training/openrlhf/ppo_script.py` to extend `openrlhf.cli.train_ppo_ray`. The structure is more copy/paste than desired in order to override the Actor's trainer since `ray` doesn't allow inheritance of `remote` functions.
* Move `MaxTimeManager` to a shared utility
* Remove `--max_samples` from default ppo options


nemo_skills/training/openrlhf/ppo_script.py
```diff
@@ -0,0 +1,721 @@
...
+from openrlhf.models import Actor
+from openrlhf.utils import blending_datasets, get_tokenizer
+from openrlhf.utils.deepspeed import DeepspeedStrategy
+
+from openrlhf.trainer.ray.ppo_actor import Actor
+
+@ray.remote(num_gpus=1)
+class CustomActorModelRayActor(BasePPORole):
+
+    # The only unique method here is .fit(), which is needed to override the Trainer
```
titu1994, line 74: Do you think monkey patching is a viable option? We only really need to modify the save_logs_and_checkpoints() method

titu1994 (approved):
Overall looks quite good, though i really wish we can upstream this stuff to OpenRLHF asap

titu1994:
Best to get reviews from @Kipok before merging

nemo_skills/training/openrlhf/ppo_script.py
tmfs10: more modifications need to be made by me

tmfs10 (approved):

nemo_skills/training/openrlhf/ppo_script.py
gwarmstrong: yeah unfortunately this is the way--I tried 3 or 4 other ways of extending without copying the whole class. None of them work particularly well due to integration with ray. It's unfortunate--I'm pretty sure the whole issue would be solved if they had just done something as simple as:

```python
class _CustomActorModelRayActor(BasePPORole):
     def fit(...):
         ....
         
     def ...
     
 @ray.remote(num_gpus=1)
class CustomActorModelRayActor(_CustomeActorModelRayActor):
     pass
```

PR #361, opened 2025-02-07
Add input key option and remove default apply_chat_template

gwarmstrong:
Small PR to  allow input key specification and not apply chat template from HF model by default.

## Changelog

* BREAKING: New option to specify `--input_key`. Previously defaulted to `context_messages`, which is the value in the openrlhf dataset. Now it defaults to `input`, which is the key for pre-templated data prepared for SFT in NeMo-Skills.
* BREAKING: remove `--apply_chat_template` argument by default, it must now be passed as an extra argument if needed to apply chat template directly from the model's tokenizer. This is needed to allow custom templated data for training.


nemo_skills/pipeline/openrlhf/ppo.py
```diff
@@ -97,7 +98,7 @@ def format_data_args(self):
         # Note: Validation data isnt used as of now
         # If using chat message dict as data, add `--apply_chat_template`
         # and --input_key 'context_messages'
-        cmd = f" --prompt_data {self.prompt_data} " f" --input_key 'context_messages' " f" --apply_chat_template "
+        cmd = f" --prompt_data {self.prompt_data} " f" --input_key '{self.input_key}' "
```
Kipok, line 101:
```suggestion
        cmd = f" --prompt_data {self.prompt_data} --input_key '{self.input_key}' "
```

Kipok (approved):
thanks!

PR #367, opened 2025-02-08
wedu/doc update

wedu-nvidia:


wedu-nvidia:
I added another description:
`   partition="cpu",  # if available on your cluster, but we suggest you use gpu partition="interactive" if you have large dataset, e.g., 200k`
As I used  partition="CPU" for data preparation with datasize = 200k, and it takes a few hours and still not finished

docs/pipelines/training.md
```diff
@@ -220,7 +220,7 @@ run_cmd(
     cluster=cluster,
     expname=f"{expname}-packing",
     container="nemo", # please use "nemo container" for packed data prepration
-    partition="cpu",  # if available on your cluster
+    partition="cpu",  # if available on your cluster, but we suggest you use gpu partition="interactive" if you have large dataset, e.g., 200k
```
Kipok, line 223:
```suggestion
    # this is a cpu-only operation, so if a cluster has a good cpu partition, it can be used
    # note that this is an expensive operation requiring a lot of CPUs and RAM
```

docs/pipelines/training.md
```diff
@@ -220,7 +220,7 @@ run_cmd(
...
     exclusive=True,  # better to get the full node, since packing is resource intensive
```
Kipok, line 224:
```suggestion
```

Kipok: that's default currently, so no need to specify it

wedu-nvidia: ok

Kipok (approved):

PR #368, opened 2025-02-09
Math RL data preparation

Kipok:
Breaking change: `prepare_sft_data` is renamed to `prepare_data` as it now covers more cases


nemo_skills/utils.py
```diff
@@ -350,3 +350,14 @@ def compute_chunk_ids(chunk_ids: list[int] | str, num_chunks: int) -> list[int]
         assert chunk_id >= 0, "Run ids should have 1-based indexing"
 
     return chunk_ids
+
+
+def prefill_judgement(data_point: dict) -> str | None:
```
shtoshni, line 355: Should we keep this function here or this is more appropriate for `training`? Maybe a utils.py for training folder makes sense.

shtoshni: So this is supposed to be a basic judge model which only cares about the surface form, and should be task agnostic. Is this preferred because it's lightweight and generic?

nemo_skills/training/data_preparation_utils/math_rl.yaml
```diff
@@ -0,0 +1,90 @@
+processors_to_run: all
```
shtoshni, line 1: To avoid clutter we can move all the config files in a `config` folder.

nemo_skills/training/data_preparation_utils/math_rl.yaml
```diff
@@ -0,0 +1,90 @@
...
+    input_key: ${input_key}
+    skip_first: ${skip_first}
+    deduplicate: ${deduplicate}
+    keys_to_keep:  # input/output/is_correct/judgement are automatically added
+      - "expected_answer"
+      - "total_votes"
+      - "majority_votes"
+
+  # this will optimize processors inside to avoid serializing data to disk
+  - _target_: nemo_skills.training.data_preparation_utils.merge_processor.MergeProcessor
```
shtoshni, line 57: This might be a good addition for `math_sft.yaml` and `code_sft.yaml` as well. We can do it later but good to know of this functionality.

nemo_skills/training/data_preparation_utils/preprocessing.py
```diff
@@ -397,3 +392,74 @@ def process(self):
...
+        with (open(self.input_manifest_file, "rt", encoding="utf-8") as fin,):
+            # only looping over the correct samples (unless asked for incorrect)
+            for line in fin:
+                elem = json.loads(line)
+                if 'expected_answer' in elem:
+                    elem['expected_answer'] = str(elem['expected_answer'])
+                # take only required keys from the input if exclude_optional_keys is True
+                output_sample = {}
+                if not self.exclude_optional_keys:
+                    output_sample = json.loads(line)
```
shtoshni, line 439: This `if/elif` structure is testing for two very different things. Is there a missing `else` for the first `if`?  Can the second `elif` be an `if` in itself?

nemo_skills/training/openrlhf/llm_reward.py
```diff
@@ -1,21 +1,40 @@
...
+                "predicted_answer": extract_answer(query),
+            }
+        )
+        judgement = prefill_judgement(data_points[-1])
+        if judgement is not None:
+            prefilled_judgements.append(judgement)
+            prefilled_indices.add(len(data_points) - 1)
+
     llm = get_model(server_type="trtllm")
     prompt = get_prompt('judge/math', 'qwen-instruct')
```
shtoshni, line 28: `server_type`, `prompt_config`, and `prompt_template` should not be hardcoded in the function. Make them parameters ideally.

nemo_skills/training/openrlhf/llm_reward.py
```diff
@@ -1,21 +1,40 @@
...
+    prefilled_indices = set()
+    for metadata, query in zip(prompt_metadata, queries):
+        data_points.append(
+            {
+                "problem": metadata["problem"],
+                "expected_answer": metadata["expected_answer"],
+                "predicted_answer": extract_answer(query),
+            }
+        )
+        judgement = prefill_judgement(data_points[-1])
```
shtoshni, line 22: Prefiling can ideally save computation but right now these data points are still being passed to the LLM. For cases where the judgement is True/Correct, we can remove those. For the ones where it is None or False, in those cases we can rely on the LLM as a judge.

nemo_skills/training/openrlhf/llm_reward.py
```diff
@@ -1,21 +1,40 @@
...
-from nemo_skills.prompt.utils import get_prompt
+
 from nemo_skills.code_execution.math_grader import extract_answer
 from nemo_skills.evaluation.metrics.utils import is_correct_judgement
+from nemo_skills.inference.server.model import get_model
+from nemo_skills.prompt.utils import get_prompt
+from nemo_skills.utils import prefill_judgement
 
 
 def reward_func(queries: list[str], prompts: list[str], prompt_metadata: list[dict]):
```
shtoshni, line 10: A one-line description that the function assigns a binary score to the prompt-response/problem-response pairs.

shtoshni, line 10: I am confused by the parameter `prompts`. It's passed as a parameter, and then the variable `prompts` is created inside the function on line 29. The other variables could also use better names. For e.g., `queries` could be `generations`/`responses`.

nemo_skills/utils.py
Kipok: Not sure I understand the question, can you clarify?

Kipok: The problem with putting it in training/utils.py is that it's also used during inference in llm math judge pipeline. So let's maybe keep in global utils for now, but at some point we will probably have a separate rl folder or something similar and we will move it there

nemo_skills/training/data_preparation_utils/math_rl.yaml
Kipok: merge processor? It's already in math sft and code sft doesn't have much filtering to benefit from it

nemo_skills/training/openrlhf/llm_reward.py
Kipok: in a follow-up pr

nemo_skills/training/openrlhf/llm_reward.py
Kipok: that's a bug, they weren't meant to be passed whenever judgement is prefilled, will fix

nemo_skills/training/openrlhf/llm_reward.py
Kipok: queries/prompts is the original notation from openrlhf, so probably best to keep it (at least for now). Agree it's a bit confusing, let me add some docs for that

Kipok:
Merging as is, but please let me know if there are any other things to fix @shtoshni

PR #374, opened 2025-02-10
Shtoshni/aime 2025

shtoshni:
AIME 2025 - Part I 

https://github.com/GAIR-NLP/AIME-Preview/blob/main/eval/data/aime/test2025-I.jsonl


Kipok:
Please update tests/datasets.py with this new set

PR #377, opened 2025-02-11
Add default for better KL estimator

tmfs10:
Gerald recommended using the flag --use_kl_estimator_k3 as default


Kipok:
but then we can't disable it from command-line, is that right? Or what's the way to override?

tmfs10:
That's true. Should i change it to a boolean variable I guess?

Kipok:
Let's just not set it? We will most likely move to a regime with no kl anyway

PR #378, opened 2025-02-11
Top answer processor

shtoshni:
The PR updates functionality to the top answer processor which can now be used for both filling the top-ranking answer and extracting it from a given set of predictions. Currently, in the PR I have NOT fixed references to the prior references in the docs. If the name seems fine, I'll update the docs as well.


nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: I'd just call it "process_answers" or "aggregate_answers". Shorter and not-less descriptive in my opinion

nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
...
+
+LOG = logging.getLogger(__file__)
+
+
+@nested_dataclass(kw_only=True)
+class ProcessTopAnswerConfig:
+    """Top-level parameters for the script"""
+
+    # Input_dir relative to which all the input_files are specified
+    input_dir: str = MISSING
```
Kipok, line 41: not related to your changes, but it's best to not use MISSING, since it's only properly checked by hydra and all of our code can be used from python directly now. Instead just have those args on top without any default value

nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
...
+    use_majority_rm_score: bool = False
+
+    # if provided, will fail if can't find this many files. Useful in scheduled
+    # pipelines to ensure this step doesn't run if some of the expected files are missing
+    require_num_files: int | None = None
+
+    def __post_init__(self):
+        """Building data_file from dataset/split if not provided directly."""
+        if isinstance(self.input_files, str):
+            self.input_files = self.input_files.split(" ")
```
Kipok, line 80: Let's also split on comma for consistency with other scripts, e.g. https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/retrieve_similar.py#L70

nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
...
+        
+        # For fill mode, output_dir is required
+        if self.process_mode == ProcessMode.FILL:
+            if cfg.output_dir is None:
+                raise ValueError("output_dir is required when mode is fill")
+        
+        # Check whether the input_dir and output_dir are valid
+        if not Path(cfg.input_dir).exists():
+            raise ValueError(f"Input directory does not exist: {cfg.input_dir}")
+        if cfg.output_dir is not None and not Path(cfg.output_dir).exists():
```
Kipok, line 131: I think we should create output_dir in this case. All of our other scripts work like that

nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
...
+            ]
+            self.output_file_handles = [open(file, "wt", encoding="utf-8") for file in output_files]
+        
+        elif self.process_mode == ProcessMode.EXTRACT:
+            if cfg.output_dir is None:
+                cfg.output_dir = cfg.input_dir
+            
+            # Determine the file suffix based on the use of RM scores
+            if cfg.use_majority_rm_score:
+                file_suffix = "-majority-rm.jsonl"
```
Kipok, line 160: I think this might make it harder to reuse this script as part of a pipeline since you'd need to be changing the name there as well. I would instead include another "fill_mode" entry in the inner dictionary on each line, so that we can refer to that later on. Will also be easier to track as it will be carried over to any subsequent generations / processing while file name will be much harder to find later on

nemo_skills/evaluation/top_answer_processor.py
```diff
@@ -0,0 +1,314 @@
...
+        """Write the results to the output file(s)"""
+        if self.process_mode == ProcessMode.FILL:
+            self._write_results_fill(all_predictions, new_answers)
+        elif self.process_mode == ProcessMode.EXTRACT:
+            self._write_results_extract(all_predictions, new_answers)
+    
+    def _write_results_fill(self, all_predictions: List, new_answers: List):
+        """Fill the expected_answer with the top answer"""
+        cfg = self.cfg
+        total_problems_changed, total_solutions_changed = 0,0
```
Kipok, line 248:
```suggestion
        total_problems_changed, total_solutions_changed = 0, 0
```

Kipok (commented):
A couple of small suggestions. Please make sure this is well tested as changes are large. Ideally, we should add some CI tests

nemo_skills/evaluation/top_answer_processor.py
shtoshni: I like `aggregate_answers`. Changing to that.

nemo_skills/evaluation/top_answer_processor.py
shtoshni: Let me make that change

nemo_skills/evaluation/top_answer_processor.py
shtoshni: Good point. Let me introduce the `fill_mode` entry. 
For the extract mode, let's just refer the file as `output-agg.jsonl`?

Kipok: sure

shtoshni:
> A couple of small suggestions. Please make sure this is well tested as changes are large. Ideally, we should add some CI tests

Added two tests for extract and fill mode.

PR #381, opened 2025-02-12
Add option to pass extra args to reward fn

tmfs10:


Kipok:
this is unnecessary, right? You can just use that parameter directly as any other extra arguments

PR #386, opened 2025-02-13
Update retrieve_similar.py to compute similarity with lower memory

Kipok:


nemo_skills/inference/retrieve_similar.py
```diff
@@ -63,7 +77,10 @@ class RetrieveSimilarConfig:
     top_k: int = 3
     retrieve_key: str = 'problem'
 
-    batch_size: int = 512
+    batch_size: int = 2048
```
shtoshni, line 80: Add a line to describe batch_size so that reader can easily distinguish it from chunk_size

nemo_skills/inference/retrieve_similar.py
```diff
@@ -97,7 +114,7 @@ def retrieve_similar(cfg: RetrieveSimilarConfig):
 
     retrieve_from_embeddings = encode(model, retrieve_from_list, batch_size=cfg.batch_size)
```
shtoshni, line 115: Unrelated to this PR, but should we move the encode operation to GPU if GPU is available? Right now it looks like it's done on CPU.

Kipok: it's already done on gpu if it's available

shtoshni: Interesting, I read up and the model is loaded on the most performant device by default. However, encoding by default moves the computed embeddings to numpy by default for memory reasons - https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L461-L681

shtoshni: To make it faster we can pass in `normalize_embeddings=True` in the `encode` function, and then call `util.dot_score` - https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L503

Kipok: Let's do it in another PR - it's really not bottlenecked by speed, it's very fast, was just failing out-of-memory for me

PR #389, opened 2025-02-13
Remove code redundancy in TRTLLM

shtoshni:


shtoshni:
@Kipok - The errors are happening during evaluation with sglang. The TRTLLM changes made in this PR should have nothing to do with it. Should we just merge it if it looks fine to you? I'm figuring out the reason for sglang error.

Kipok (approved):
Thanks!

PR #390, opened 2025-02-13
Fix a bug with incorrect tracking of gen ids

Kipok:


i-vainn (approved):
I've tested, it works in my setup. Thanks!

PR #393, opened 2025-02-17
Handling server error during generation

shtoshni:
Handling server errors during generation gracefully.


nemo_skills/inference/server/model.py
```diff
@@ -177,7 +177,11 @@ def get_generations(self, generation_ids: list[str]) -> list[dict]:
             if not future.done():
                 output = {'generation': None}
             else:
-                output = future.result()
+                try:
```
Kipok, line 180: Should we have a parameter that enables this behavior? I'm not entirely sure we should always ignore any errors and keep generating as sometimes the errors are not recoverable and there is no point in generating more if every single example is going to fail with an error. Perhaps by default we raise an error and if some parameter is specified we put it in dict and keep going?

nemo_skills/inference/server/model.py
```diff
@@ -177,7 +177,11 @@ def get_generations(self, generation_ids: list[str]) -> list[dict]:
...
+                    output = future.result()
+                except Exception as e:
+                    LOG.error(f"Error in generation {generation_id}: {e}")
+                    output = {"generation": "", "inference_error": str(e)}
```
Kipok, line 184: What was an error when it was set to None?

shtoshni: So when we set it to None, this is removed from the model's internal queue. But the generation loop checks for "None" in the output and doesn't remove it - https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/generate.py#L410

So my guess is in the next loop iteration, there's a mismatch in generation queue and the queue the model.py code is maintaining. 

`  File "/nemo-run/code/nemo_skills/inference/generate.py", line 405, in async_loop
    generations = self.llm.get_generations(list(requests_in_progress.keys()))
  File "/nemo-run/code/nemo_skills/inference/server/model.py", line 173, in get_generations
    raise ValueError(f"Generation id {generation_id} not found.")
ValueError: Generation id 51b2e87b-f9b2-4b8a-badf-c362392833ed not found.`

Kipok:
Would be good to actually update our trtllm logic as well to fail by default. Currently it will put error in the generation and keep going which might lead to issues that are hard to detect. Probably better to fail by default and have an option to keep current behavior

PR #396, opened 2025-02-26
Randomize wandb run-id in order to prevent hangs

titu1994:
- Changes the wandb-id with a random value in order to prevent crashes from repeating wandb job with exact same id that was already deleted

### DIFF
--- a/nemo_skills/pipeline/openrlhf/ppo.py
+++ b/nemo_skills/pipeline/openrlhf/ppo.py
@@ -15,6 +15,7 @@
 import json
 import logging
 import os
+import random
 from dataclasses import dataclass
 from enum import Enum
 from typing import List, Optional
@@ -145,7 +146,7 @@ class PPOOpenRLHFTask:
                 f" --use_wandb $WANDB_API_KEY "
                 f" --wandb_project {wandb_project} "
                 f" --wandb_run_name {expname} "
-                f" --wandb_id {expname} "
+                f" --wandb_id {expname}_{str(random.randint(0, 2 ** 32))} "
                 f" --wandb_resume auto"
             )
         else:
--- a/nemo_skills/pipeline/openrlhf/sft.py
+++ b/nemo_skills/pipeline/openrlhf/sft.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+import random
 from dataclasses import dataclass
 from typing import List
 
@@ -129,7 +130,7 @@ def format_wandb_args(cluster_config, disable_wandb, wandb_project, expname):
             f" --use_wandb $WANDB_API_KEY "
             f" --wandb_project {wandb_project} "
             f" --wandb_run_name {expname} "
-            f" --wandb_id {expname} "
+            f" --wandb_id {expname}_{str(random.randint(0, 2 ** 32))} "
             f" --wandb_resume auto"
         )
     else:


gwarmstrong:
@titu1994 doesn't this break the ability to actually resume jobs in wandb? E.g., if you change the id, wandb runs will not  be connected. Please check out docs for `id` and `resume` [here](https://docs.wandb.ai/ref/python/init/). I think the better approach would be to avoid repeating the id's of deleted jobs.

gwarmstrong:
If you really need a fail-safe for this situation (e.g., don't want jobs that have been queued for days to crash for an avoidable reason), I might suggest adding something like an option for `wandb_init_failure=increment` instead, which defaults to `fail`. Setting to `increment` would try/except the wandb.init call, and add some predictable suffix in that case (e.g., try appending `_1`, then `_2`, etc. for some max number of increments), this way it is easier to recover the run id later, but by default the wandb resume behavior should be preserved.

gwarmstrong:
Personally, the low-code solution I use is to always add a `run_{i}` counter to my expnames so I re-run an experiment without causing id conflicts by incrementing the run counter `i`.

titu1994:
Crashing an experiment due to an experiment logger is completely incorrect - and this behaviour will not be changed in wandb. So it is better that this be the default behaviour and if the user really wants to they can overide some flag to force same run id. 

It crashes even if you use an older job id that you have already deleted (due to 7 day retention policy in wandb).
Can you remember every single experiment name used in last 7 days and avoid it? I cant.

titu1994:
Not everyone uses pythonic way of scheduling jobs either - so its not like we can auto increment a counter. Having a default that has a high likelihood of crashing a job should not be the default

titu1994:
> 
If you really need a fail-safe for this situation (e.g., don't want jobs that have been queued for days to crash for an avoidable reason), I might suggest adding something like an option for wandb_init_failure=increment instead, which defaults to fail. Setting to increment would try/except the wandb.init call, and add some predictable suffix in that case (e.g., try appending _1, then _2, etc. for some max number of increments), this way it is easier to recover the run id later, but by default the wandb resume behavior should be preserved. Or another strategy could be disable where you simply turn off wandb.

This would not work. Wandb doesn't crash - it goes into a login cycle. It is Ray which times out after 90s of no communication.

gwarmstrong:
> Crashing an experiment due to an experiment logger is completely incorrect
> This would not work. Wandb doesn't crash - it goes into a login cycle. It is Ray which times out after 90s of no communication.

[Here is a timeout wrapper](https://www.perplexity.ai/search/write-a-python-timeout-module-5rZCBgm7Qt2I_kkFgnUT8w#0)--you can start by using this to wrap the `wandb.init` call, and can catch this `TimeoutError` exception instead and implement a resolution strategy.

I think catching the error and recovering while preserving job-resuming for users who don't re-use exp id's on different experiments would be more useful than forcing a false conflation between "run never fails due to logger" and "you can never connect your resumed training jobs in wandb because there a random suffix appended to it".

> Can you remember every single experiment name used in last 7 days and avoid it? I cant.
> Having a default that has a high likelihood of crashing a job should not be the default

At the risk of being abrasive (apologize in advance), I think this is specific to the precise way in which you are using experiment id's and wandb, I have never had this issue and I have run hundreds of experiments in the last two weeks with the current method. 

IMO, the specific solution that was proposed in this PR is overfit to the use case of "I want to delete wandb runs and re-use it's id", which is a use-case that wandb specifically does not support, and the solution further penalizes users who are using wandb in the intended way by removing the ability to have jobs resume properly.

> Not everyone uses pythonic way of scheduling jobs either - so its not like we can auto increment a counter

I'm having a hard time understanding the logic here--are you referring to pythonic nemo-skills? If you are using nemo-skills in any form for openrlhf, you must be using pythonic wandb internally, in which case we have the opportunity to implement a recovery strategy of some kind.

If you want the lowest hanging fruit that helps your jobs continue to run without significantly disrupting the status quo, I would recommend trying the timeout handler linked above and turning off wandb in the case `init` fails.

Kipok:
Can we just add a quick call to wandb before we submit a job inside the pipeline script that will check if run is there or not and fail right away? That's the best option in my opinion as I agree that it's useful to be able to properly resume jobs and also is really annoying if your job fails because of the logging problem

gwarmstrong:
> Can we just add a quick call to wandb before we submit a job inside the pipeline script that will check if run is there or not and fail right away?

Looks doable:

```
>>> import wandb
>>> api = wandb.Api()
>>> runs = api.runs(f'{entity}/{project}')
>>> run_ids = [run.id for run in runs]
>>> run_ids
```
Where entity and project correspond to our wandb org and the project you are trying to upload to.

Kipok:
I think you don't even need to specify entity, it will default to the right one. And then we should just test if there is a proper api to figure out if the current id is already deleted or not and abort with a proper error message if that's the case (or have an option to go ahead and add a random string so that it doesn't fail)

Kipok:
Let's close this for now and we can create a new pr with a check before submission if this is still needed

PR #397, opened 2025-02-26
Adding nemo-grpo support

Kipok:
Plus a few other fixes such as fixing the task_num being a string in slurm, changing summary to log in wandb and re-using thread pool from main class in trtllm model

### DIFF
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -22,6 +22,7 @@ To build vllm image do the following
    2. checkout tag/commit
    3. run `pip install -e .`
 7. install flask with `pip3 install --ignore-installed flask` (needed for NeMo-Aligner GRPO)
+8. git clone NeMo-Aligner into /opt and run `pip3 install -e /opt/NeMo-Aligner/tensor_comms`
 
 Currently it takes 10+ hours to build this docker. Is there anything we can do to speed it up?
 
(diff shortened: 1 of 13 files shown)


Kipok:
NeMo-GRPO is mostly working, but likely to change rapidly still

nemo_skills/training/start_grpo.py
```diff
@@ -0,0 +1,210 @@
...
+        # "your_task": your_environment,
+    }
+    rollout_generator = SequenceRewardRolloutGenerator(cfg.trainer.grpo, tasks_to_environments)
+
+    timer = Timer(cfg.exp_manager.get("max_time_per_run") if cfg.exp_manager else None)
+
+    batch_iterator_cfg = cfg.trainer.grpo.get("batch_iterator", {})
+    batch_iterator_cls = get_batch_iterator_cls(batch_iterator_cfg)
+
+    print([(name, p.shape) for name, p in ptl_model.model.named_parameters()], flush=True)
```
shtoshni, line 184: Use logger

Kipok: this is a copy-paste from nemo-aligner, so easier not to modify it unless necessary to make it simpler to upgrade in the future

shtoshni:
The PR looks good to me. I think the `pipeline/utils.py` is getting a bit cumbersome. Maybe we can refactor it in later versions.

dockerfiles/README.md
```diff
@@ -21,6 +21,8 @@ To build vllm image do the following
    1. `git clone https://github.com/Kipok/OpenRLHF`
    2. checkout tag/commit
    3. run `pip install -e .`
+7. install flask with `pip3 install --ignore-installed flask` (needed for NeMo-Aligner GRPO)
+8. git clone NeMo-Aligner into /opt and run `pip3 install -e /opt/NeMo-Aligner/tensor_comms`
```
shtoshni, line 25: `tensor_comms` ??

Kipok: yes, why is that concerning?

shtoshni: does that path even exist? I don't see it in NeMo-Aligner repo

Kipok: it's using a custom version that's mentioned below. Let me add a reference here as well

PR #399, opened 2025-03-01
skipping seeds and chunks marked with .done files

Alcray:


nemo_skills/pipeline/generate.py
```diff
@@ -379,8 +383,84 @@ def generate(
...
+                    for line in output.splitlines():
+                        if line.startswith("rs-"):
+                            parts = line.split("-")
+
+                            seed = int(parts[1])
+                            chunk = int(parts[3])
+                            if seed not in missing_map:
+                                missing_map[seed] = []
+                            missing_map[seed].append(chunk)
+            # TODO: Remove this print statement
```
Kipok, line 455: do that?

nemo_skills/pipeline/generate.py
```diff
@@ -379,8 +383,84 @@ def generate(
...
+        expmap = {i: copy.deepcopy(chunk_ids) for i in random_seeds}
+
+        # TODO add support of skipping running of the same seed or same chunk
+        if not redone:
+            status_dir = get_unmounted_path(cluster_config, f"{output_dir}")
+            seeds_str = " ".join(str(s) for s in random_seeds)
+            chunks_str = " ".join(str(c) for c in chunk_ids)
+            missing_map = {}
+
+            if random_seeds == [None]:
```
Kipok, line 395: let's try to have a more concise and modular version of the same logic. E.g. one way would be to add a function that will give you a list of all output files that are expected for the current job. Then you have a single call to get_tunnel with that list and a consistent logic to parse it. Currently there is a bit too much duplication and it's harder to understand what's going on

nemo_skills/pipeline/generate.py
```diff
@@ -379,8 +383,84 @@ def generate(
         if not random_seeds:
             random_seeds = [None]
 
-        for chunk_id in chunk_ids:
-            for seed in random_seeds:
+        expmap = {i: copy.deepcopy(chunk_ids) for i in random_seeds}
```
Kipok, line 386: maybe call it remaining_jobs or something like that, which is more explicit

nemo_skills/pipeline/generate.py
```diff
@@ -302,6 +302,11 @@ def generate(
         "--not_exclusive",
         help="If --not_exclusive is used, will NOT use --exclusive flag for slurm",
     ),
+    redone: bool = typer.Option(
```
Kipok, line 305: Let's maybe call it ignore_done

Kipok (changes requested):

nemo_skills/pipeline/generate.py
Alcray: Done

nemo_skills/pipeline/generate.py
Alcray: Done

nemo_skills/pipeline/generate.py
Alcray: Done

Kipok (approved):
thanks, much cleaner now! I added a few more things for logging and correctly handling local execution

PR #402, opened 2025-03-04
Add support for last ckpt saving

SeanNaren:
This PR adds saving the last checkpoint to the final folder, for subsequent pipeline commands to rely on. By default, if average steps is not set (or is not set to all), we will copy the last checkpoint to the output directory.


SeanNaren:
@Kipok can confirm this works as expected, copying the last checkpoint to the specified output directory

SeanNaren:
Actually hold for a bit, I think I spot some potential error.

nemo_skills/pipeline/train.py
```diff
@@ -251,7 +259,7 @@ def train(
     average_steps: str = typer.Option(
         'all',
         help="List of commas separated checkpoint steps to average. E.g 1000,5000. "
-        "If None, will skip prepare eval stage.",
+        "If None, will save the last checkpoint.",
```
Kipok, line 262: Let's make it a separate parameter. Or perhaps use that when average_steps = 'last'. I think it's still useful to be able to skip prepare eval stage completely if someone want to have their own way of handling checkpoints

Kipok (commented):
Just one small comment, otherwise looks good to me as long as you tested that it works

nemo_skills/pipeline/train.py
SeanNaren: thanks for checking this! wasn't sure which way we wanted to go, agree with this!

Kipok (approved):
Thanks!

PR #408, opened 2025-03-10
Update code reuse logic to use NeMo-Run's latest update for packaging jobs

hemildesai:
Depends on https://github.com/NVIDIA/NeMo-Run/pull/171

### DIFF
--- a/nemo_skills/pipeline/utils.py
+++ b/nemo_skills/pipeline/utils.py
@@ -1164,11 +1164,13 @@ def add_task(
         if isinstance(tunnel, run.SSHTunnel) and reuse_code:
             reuse_code_exp = reuse_code_exp or REUSE_CODE_EXP.get(tunnel_hash(tunnel))
             if reuse_code_exp is not None:
-                reuse_code_exp = (
-                    reuse_code_exp
-                    if isinstance(reuse_code_exp, run.Experiment)
-                    else run.Experiment.from_title(reuse_code_exp)
-                )
+                if isinstance(reuse_code_exp, str):
+                    try:
+                        reuse_code_exp = run.Experiment.from_id(reuse_code_exp)
+                    except Exception:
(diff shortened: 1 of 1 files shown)


titu1994 (commented):
Just to make sure, this will allow jobs under the same experiment to automatically share one copy of the symlinked code dir right ?

nemo_skills/pipeline/utils.py
```diff
@@ -1164,15 +1164,21 @@ def add_task(
         if isinstance(tunnel, run.SSHTunnel) and reuse_code:
             reuse_code_exp = reuse_code_exp or REUSE_CODE_EXP.get(tunnel_hash(tunnel))
             if reuse_code_exp is not None:
-                if isinstance(reuse_code_exp, run.Experiment):
-                    LOG.info("Reusing code from experiment %s", reuse_code_exp._title)
-                    reuse_dir = reuse_code_exp.tunnels[tunnel.key].packaging_jobs['nemo-run'].dst_path
+                reuse_code_exp = (
+                    reuse_code_exp
+                    if isinstance(reuse_code_exp, run.Experiment)
+                    else run.Experiment.from_title(reuse_code_exp)
```
Kipok, line 1170: Does it work without context manager? This code is running from inside another `with run.Experiment`, so I was getting errors previously when using it like that

hemildesai: I have now made the SlurmScheduler independent of context manager, but I will test the use case by providing a str as `reuse_code_exp`

Kipok: thanks, let me know if it works and we can merge after that

hemildesai: Tested with both `title` and `id` strs i.e. (`exp._title` and `exp._id`) with https://github.com/NVIDIA/NeMo-Skills/pull/408/commits/5e1b991f1e96e24bc4c4d6ac505b5af09ad5c687

Kipok (approved):
thanks!

PR #410, opened 2025-03-11
FEAT allow disable eval_data

gwarmstrong:
Automatically disables validation if no eval_data is supplied. Needs https://github.com/titu1994/verl/pull/3


Kipok (approved):
thanks!

PR #411, opened 2025-03-12
MAINT expose timeout to verl ppo

gwarmstrong:
you can override verl timeout with:
```
++trainer.timeout=00:03:20:00 # dd:hh:mm:ss
```


Kipok (approved):
thanks!

PR #414, opened 2025-03-13
wrap_arguments split on specifically space instead of all whitespace

arnavkomaragiri:
changed wrap_arguments() to split on spaces instead of generic whitespace


Kipok (approved):
thanks!

PR #415, opened 2025-03-14
Fix checkpoint averaging logic for SFT

SeanNaren:
I added a bug where if the checkpoint averaging was set to 'all' it wasn't correctly being set to "". Need to test this and then can merge.


SeanNaren:
Confirmed it works, @Kipok this should fix checkpoint averaging.

Kipok (approved):

PR #417, opened 2025-03-18
Set num_workers=0 as default

Kipok:
This fixes oom issue with large packed datasets and also makes job to start in under 5 minutes vs close to 20-30 minutes without that. No speed difference for our workloads


Kipok:
Also fixed all issues in our gpu tests, so as of this commit they all run correctly. Still cannot use ci most of the time because of infra issues, but locally everything is running well

PR #420, opened 2025-03-21
explicitly added min_p to inference config

arnavkomaragiri:
just added min_p field to InferenceConfig explicitly


Kipok (approved):
Thanks!

PR #421, opened 2025-03-25
Averaged pass at k generations (as in R1 paper)

tatevik-t:
Added pass@1[k] (averaged pass@1 on k samples) implementation as described in [R1 paper, p.12](https://arxiv.org/pdf/2501.12948)

Screenshot 2025-03-25 at 13 08 36


jubick1337 (approved):
LGTM!

PR #422, opened 2025-04-02
Adding AzureOpenAIModel

tatevik-t:


docs/basics/index.md
```diff
@@ -71,6 +71,20 @@ You can either use [OpenAI models](https://platform.openai.com/docs/overview) or
         ++prompt_config=./prompt.yaml
     ```
 
+=== "Azure OpenAI models"
```
Kipok, line 74: This is probably not the best place for an example of azure openai. Instead, maybe add it to docs/basics/inference.md?

nemo_skills/inference/server/model.py
```diff
@@ -732,6 +732,126 @@ def get_model_name_from_server(self):
...
+        port: str = '5000',
+        model=None,
+        base_url=None,
+        api_key=None,
+        api_version="2024-12-01-preview",
+        **kwargs,
+    ):
+        BaseModel.__init__(self, **kwargs)
+
+        from openai import AzureOpenAI
```
Kipok, line 748: openai is the default requirement, so can move the import to top

nemo_skills/inference/server/model.py
```diff
@@ -732,6 +732,126 @@ def get_model_name_from_server(self):
         return model_name
 
 
+class AzureOpenAIModel(OpenAIModel, BaseModel):
```
Kipok, line 735: Can we reuse openaimodel directly here? Seems that most of the code is the same, but can add an if-else somewhere if needed?

docs/basics/index.md
tatevik-t: moved to the inference.md

nemo_skills/inference/server/model.py
tatevik-t: rewrote this part

nemo_skills/inference/server/model.py
```diff
@@ -535,12 +535,13 @@ def __init__(
...
         api_key=None,
+        api_version=None,
         max_retries: int = 3,
         initial_retry_delay: float = 2.0,
+        endpoint_type: str = "openai",
         **kwargs,
     ):
-        super().__init__(**kwargs)
-        from openai import OpenAI
+        BaseModel.__init__(self, **kwargs)
```
Kipok, line 544:
```suggestion
        super().__init__(**kwargs)
```

nemo_skills/inference/server/model.py
```diff
@@ -552,16 +553,28 @@ def __init__(
             base_url = os.getenv("NEMO_SKILLS_OPENAI_BASE_URL", f"http://{host}:{port}/v1")
 
         if api_key is None:
-            if base_url is not None and 'api.nvidia.com' in base_url:
-                api_key = os.getenv("NVIDIA_API_KEY", api_key)
-                if not api_key:
-                    raise ValueError("NVIDIA_API_KEY is required for Nvidia-hosted models.")
+            if base_url is not None and 'nvidia.com' in base_url:
+                if endpoint_type == "azureopenai":
```
Kipok, line 557: can nvidia models be accessed through azure? Can you give an example of usage when this condition is triggered? Or did you mean to put it as a separate if, not inside `'if 'nvidia.com' in base_url`?

nemo_skills/inference/server/model.py
```diff
@@ -552,16 +553,28 @@ def __init__(
...
+                        raise ValueError("NVIDIA_API_KEY is required for Nvidia-hosted models.")
             elif base_url is not None and 'api.openai.com' in base_url:
                 api_key = os.getenv("OPENAI_API_KEY", api_key)
                 if not api_key:
                     raise ValueError("OPENAI_API_KEY is required for OpenAI models.")
 
-        self.client = OpenAI(api_key=api_key, base_url=base_url)
+        # Initialize appropriate client based on endpoint_type
+        if endpoint_type == "azureopenai":
+            self.client = openai.AzureOpenAI(api_key=api_key, api_version=api_version or "2024-12-01-preview", azure_endpoint=base_url)
```
Kipok, line 572: let's just set this value "2024-12-01-preview" as default for api_version parameter?

nemo_skills/inference/server/model.py
```diff
@@ -673,37 +686,45 @@ def _generate_single(
...
-                    messages=prompt,
-                    logprobs=top_logprobs is not None,
-                    top_logprobs=top_logprobs,
-                    timeout=timeout,
-                )
-                break  # Success, exit the retry loop
+                if self.model_type == "Azure OpenAI":
+                    response = self.client.chat.completions.create(
+                        model=self.model,
+                        temperature=1,
```
Kipok, line 694: do they really not support temperature? I'd double check all parameters here, hopefully there is more that can be reused. And then you can just have all parameters as a dict variable and just pop/rename a few for the Azure case and have a single call to get a response

docs/basics/inference.md
```diff
@@ -96,6 +96,20 @@ Click on :material-plus-circle: symbols in the snippet below to learn more detai
 
          To use OpenAI models, use `OPENAI_API_KEY` and set `base_url=https://api.openai.com/v1`.
 
+         To use Azure OpenAI models, use `endpoint_type="azureopenai"` and set `base_url` to your Azure OpenAI endpoint URL.
```
Kipok, line 99: we can keep it like this, but it's a bit hard to find. Instead if you'd like, feel free to change this original section to only have self-hosted models and then we can have 2 tabs, one with basic inference and another with code execution. And this will be "self-hosted models" subsection and you can create another subsection after that with "api models", which will have 3 tabs, one for Nvidia, one for OpenAI and another for AzureOpenAI to make it more visible.

You can always run `mkdocs serve` to build docs locally and check that the website looks as you expect it to be

Kipok:
an equivalent implementation was merged in https://github.com/NVIDIA/NeMo-Skills/pull/511

PR #423, opened 2025-04-02
Change the default evaluation for mcq datasets to the boxed format

dgtm777:
1. change the default prompt for mcq datasets
2. merge the "problem" and "options" fields when preparing the datasets


nemo_skills/dataset/mmlu-pro/__init__.py
```diff
@@ -15,12 +15,9 @@
...
-DATASET_GROUP = 'multichoice'
-METRICS_TYPE = 'multichoice'  
-DEFAULT_EVAL_ARGS = (
-    "++eval_type=multichoice "
-)
-
+PROMPT_CONFIG = "generic/general-boxed"
+DATASET_GROUP = "multichoice"
+METRICS_TYPE = "multichoice"
+DEFAULT_EVAL_ARGS = "++eval_type=math"
```
Kipok, line 21: I think we should have a different eval type here as going through math means running things in sandbox and using sympy for complex evaluation of expressions which feels unneded here?

nemo_skills/dataset/mmlu-redux/prepare.py
```diff
@@ -91,22 +91,28 @@ def format_entry(entry, category):
         # bad_question_clarity, bad_options_clarity, no_correct_answer,
         # multiple_correct_answers, expert and wrong_groundtruth with no labels
         return None
-    return {
-        "question": entry['question'],
+    new_entry = {
```
Kipok, line 94: if it's identical logic across a bunch of datasets, might make sense to move it to utils and import from there

ekmb (approved):

Kipok (approved):

PR #424, opened 2025-04-04
Postprocess after merge

gwarmstrong:
Moves the postprocess command after merging chunks so that commands can be run once on files with predictable patterns.


nemo_skills/pipeline/generate.py
```diff
@@ -153,6 +154,9 @@ def get_remaining_jobs(cluster_config, output_dir, random_seeds, chunk_ids, reru
 
     return missing_jobs
 
+def escape_shell_command(command: str) -> str:
```
Kipok, line 157: unused now?

nemo_skills/inference/merge_chunks.py
```diff
@@ -25,13 +39,20 @@
...
     # Create .done file and delete input files
     open(f"{output_file}.done", "w").close()
     for file_ in input_files:
         os.remove(file_)
-except subprocess.CalledProcessError:
-    print("An error occurred while concatenating files")
+
+    # Execute the post-merge command, if provided
+    if post_merge_command:
+        print(f"Executing post-merge command: {''.join(post_merge_command)}")
```
Kipok, line 53:
```suggestion
        print(f"Executing post-merge command: {' '.join(post_merge_command)}")
```

nemo_skills/inference/merge_chunks.py
```diff
@@ -25,13 +39,20 @@
...
+        subprocess.run(post_merge_command, shell=True, check=True)
```
Kipok, line 54:
```suggestion
        subprocess.run(post_merge_command, check=True)
```

nemo_skills/inference/merge_chunks.py
```diff
@@ -25,13 +39,21 @@
...
     open(f"{output_file}.done", "w").close()
     for file_ in input_files:
         os.remove(file_)
-except subprocess.CalledProcessError:
-    print("An error occurred while concatenating files")
+
+    # Execute the post-merge command, if provided
+    if post_merge_command:
+        print(f"Executing post-merge command: {' '.join(post_merge_command)}")
+        result = subprocess.run(' '.join(post_merge_command), shell=True, check=True, capture_output=True, text=True)
```
Kipok, line 54: do we need to capture output? What if some important message is in stderr, is that going to be printed or hidden now?

gwarmstrong: you're right--it's better to not have the capture  logic. It logs it all fine with the `check` and `capture_output` flags set back to default.

PR #431, opened 2025-04-12
Added nvmath-hard dataset

i-vainn:
Added nvmath-hard dataset which is blend of HMMT Nov 24, HMMT Feb 24, HMMT Feb 25 and aime24 and aime25.
From HMMT I removed proof problems, questions with yes/no answers and questions that ask for estimating answer (as they are almost unsolvable with exact match comparison).


shtoshni:
@i-vainn - Just change the name of the dataset and we can merge it.

Kipok (approved):

PR #433, opened 2025-04-16
[CodeGen] Port core functionality and refactors

titu1994:
This pull request introduces several significant changes to the `nemo_skills` package, including the addition of new utility functions, updates to existing scripts.

# Core changes
- Refactor setup.py to support extra includes. This is for future code eval domains (if we don't do dynamic install inside container)
- Refactor the utils file to extract out new file ops. jload / jdump are for uniform jsonl handling throughout the repo with safety checks to skip corrupted samples due to timeouts. jload_chunk is more efficient for massive files x dispatch across cluster.
- `app.py` gets new functionality for dynamic mounting. Checks for is_mounted, get_mounted and get_unmounted that automatically, resolve mount paths will need to be added to all pipeline scripts in the future (can do in this or next pr)
- `ssh_config` can resolve env vars, reducing need for user specific cluster configs
- some config stuff for codegen judge and utg 

# Copilot Changelog below

### New Utility Functions:
* Added `nemo_skills/file_utils.py` with functions for file operations such as `unroll_files`, `_make_w_io_base`, `_make_r_io_base`, `jdump`, `jload`, `count_newlines`, `calculate_chunk_indices`, and `jload_chunk`. These functions facilitate efficient file handling and JSON operations.

### Updates to Existing Scripts:
* Modified `nemo_skills/__init__.py` to import `__version__` from `nemo_skills.version`.
* Updated `dockerfiles/Dockerfile.nemo-skills` to install additional dependencies using `pip install -e .[all]`.
* Enhanced `nemo_skills/pipeline/app.py` with new functions to handle mount paths: `is_mounted_filepath`, `get_mounted_filepath`, `get_unmounted_filepath`, and `resolve_mount_paths`.
* Added environment variable resolution in `_get_tunnel_cached` function in `nemo_skills/pipeline/utils.py`.

### New Configuration Files for Code Generation and Unit Testing:
* Introduced `nemo_skills/prompt/config/generic/codegen_system.yaml` for default code generation prompts.
* Added `nemo_skills/prompt/config/judge/code.yaml` for judging code generation tasks, including criteria and example questions.
* Added `nemo_skills/prompt/config/unit_test/code.yaml` for generating assertion test cases for code tasks.

### Dependency Updates:
* Updated `nemo_skills/utils.py` to import `fire` and related modules, and included imports from the newly added `file_utils` module.


nemo_skills/pipeline/app.py
```diff
@@ -180,3 +180,145 @@ def add_mount_path(mount_source: str, mount_dest: str, cluster_config):
...
+        )
+
+    # replace the mount destination inside the filepath with the mount source
+    mount_source, mount_dest = mount_path.split(':')
+    filepath = mount_dest + filepath[len(mount_source) :]  # replace the mount destination with the mount source
+
+    return filepath
+
+
+def get_unmounted_filepath(filepath: str, cluster_config: dict,):
```
Kipok, line 242: is this the same as https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/utils.py#L114? Should we move all path utils in a separate file? Maybe we should have a utils folder and all these files there or something like this

nemo_skills/prompt/config/generic/codegen_system.yaml
```diff
@@ -0,0 +1,13 @@
+# Default prompt that does not contain any extra instructions. Can be used for instruction following evals (e.g. ifeval)
```
Kipok, line 1: comment needs an update

nemo_skills/prompt/config/judge/code.yaml
```diff
@@ -0,0 +1,82 @@
...
+# Allowed keys for judgement - {Yes or No}, {Yes}, {No}. Other keys must be explicitly escaped in code.
+user: |
+  '''
+  You are an expert python programmer. 
+
+  Below is a question and code solution. Decide if the solution follows the below criteria and give a final Yes/No, and place it in the <judge></judge> tags.
+  
+  Only look at the function generated, not any examples/print statements etc. Just the core logic.
+  
+  Please first briefly describe your reasoning (in less than 30 words) without writing any code, and then write Decision: \\\\boxed{Yes or No} in your last line.
```
Kipok, line 15: does this prompt need some special logic for .fill? Otherwise, it should complain as you don't escape with {{}} here?

nemo_skills/utils.py
```diff
@@ -102,6 +102,14 @@ def setup_logging(disable_hydra_logs: bool = True, log_level: int = logging.INFO
             ["hydra.run.dir=.", "hydra.output_subdir=null", "hydra/job_logging=none", "hydra/hydra_logging=none"]
         )
 
+    return logger
+
+
+def get_skills_root_dir():
+    """Get the root directory of the NeMo Skills package."""
+    import nemo_skills
```
Kipok, line 110: why import here?

nemo_skills/version.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+MAJOR = 0
+MINOR = 6
+PATCH = 0
+PRE_RELEASE = ''
```
Kipok, line 19: what is pre-release used for? I'd probably keep this simple and just use 3 numbers

Kipok (commented):
thanks, just a few minor comments

nemo_skills/pipeline/app.py
titu1994: I can move these util functions into pipeline/utils.py. For this function specifically, it should be merged into above.

As to utils - there's two sets of utils that's commonly necessary - pipeline.cli and pipeline.utils. So we can copy parts of pipeline.utils into a another file maybe to actually be utils or just treat pipeline.utils as the generic functions to interact with pipeline.*?

nemo_skills/prompt/config/judge/code.yaml
titu1994: Good point, our code actually does have that special logic. Does wrapping it with {{..}} fix this problem? Ill update it then.

nemo_skills/utils.py
titu1994: Reverted.

nemo_skills/prompt/config/judge/code.yaml
Kipok: Yes, if you properly wrap {} and escape \ it should work by default

We might consider changing our default placeholders to use {{}} and then the normal {} can be used whenever applicable (e.g. in json format, etc.). Let me know if you guys have a bunch of use-cases like this, would need to update all the prompts, but might be more future proof

nemo_skills/pipeline/app.py
Kipok: I just feel that our utils files are way too big and have too many functions. So if there is some reasonable way to split them it would be good to do that. Especially pipeline/utils.py. But we can do it in another PR

nemo_skills/prompt/config/generic/codegen_system.yaml
titu1994: Good catch, fixed

nemo_skills/prompt/config/judge/code.yaml
titu1994: Its actually quite rare to have {} inside the prompt. Id rather follow the json prompting schema and then special cases like these use the workaround with {{}}

nemo_skills/version.py
titu1994: Fixed

nemo_skills/prompt/config/judge/code.yaml
```diff
@@ -0,0 +1,82 @@
...
+# Allowed keys for judgement - {Yes or No}, {Yes}, {No}. Other keys must be explicitly escaped in code.
+user: |
+  '''
+  You are an expert python programmer. 
+
+  Below is a question and code solution. Decide if the solution follows the below criteria and give a final Yes/No, and place it in the <judge></judge> tags.
+  
+  Only look at the function generated, not any examples/print statements etc. Just the core logic.
+  
+  Please first briefly describe your reasoning (in less than 30 words) without writing any code, and then write Decision: \\\\boxed{{Yes or No}} in your last line.
```
Kipok, line 15: are you sure you need to have \\\\ here? This yaml string shouldn't need any escaping, so this will render as actual 4 backslashes, not two if that's what you intended. And presumably, this should just be a single backslash like in the math config https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/prompt/config/generic/math.yaml#L15

nemo_skills/pipeline/utils.py
```diff
@@ -99,29 +101,279 @@ def get_registered_external_repo(name: str) -> Optional[RepoMetadata]:
...
-    raise ValueError(f"The path '{path_to_check}' is not mounted. Check cluster config.")
+    if not is_mounted_filepath(cluster_config, path_to_check):
+        raise ValueError(f"The path '{path_to_check}' is not mounted. Check cluster config.")
+
+def get_mounted_path(cluster_config: dict, path: str):
+    """
+    Resolve the mounted filepath using the cluster config to merge the mount destination path to the filepath.
+
+    Args:
+        cluster_config:
```
Kipok, line 139: best to either remove this or add some description

nemo_skills/prompt/config/judge/code.yaml
titu1994: Yeah our judge uses 4 \.  I suppose we can change the code to make it do just 1. Don't remember why we chose \. Probably Mixtral 8x22b wasn't working well so we used this in few shot and if just stuck. Id leave it for now

nemo_skills/pipeline/utils.py
titu1994: Done

PR #440, opened 2025-04-24
fix docker for verl

wedu-nvidia:
I fixed docker by setting vllm=0.7.3, the previous bug comes from newest vllm version.
i tested on cluster with this new image and it worked well


Kipok (approved):
Thanks! Will update official images with these later today

PR #445, opened 2025-04-30
prompt for qwen3

wedu-nvidia:
I added the prompt for qwen3
to switch between reasoning and non-reasoning mode.


nemo_skills/prompt/config/qwen/math-non-think.yaml
```diff
@@ -0,0 +1,5 @@
+# Prompt used by Qwen models for math benchmarks
```
Kipok, line 1:
```suggestion
# Prompt used by Qwen3 models for math benchmarks
```

Kipok, line 1: let's call the folder qwen3

wedu-nvidia: thanks, i created a new folder and rename as 
math-cot-non-think.yaml
math-cot-think.yaml

Kipok (approved):

PR #449, opened 2025-05-06
Add math evaluation without sandbox

i-vainn:
This PR integrates [huggingface/Math-Verify](https://github.com/huggingface/Math-Verify) for math evaluation, which supports antlr==4.9.3 compatible with hydra. Therefore, separate sandbox is no longer needed to launch `batch_evaluate_results` for math datasets evaluation.


nemo_skills/code_execution/math_grader.py
```diff
@@ -12,535 +12,214 @@
...
+    if not re.search(latex_env_search_pattern, current_predicted_answer, re.DOTALL):
+        current_predicted_answer = f"${current_predicted_answer}$"
+
+    parsed_gt = parse(current_gt_answer, [LatexExtractionConfig()])
+    parsed_pred = parse(current_predicted_answer, [LatexExtractionConfig()])
+
+    return verify(parsed_gt, parsed_pred, **kwargs)
+
+
+def batch_evaluate_results(
```
Kipok, line 151: If this can't use threads, then let's maybe just have a basic loop without saving tmp files and map_to_future logic? So just compute all sequentially, store results in memory and add back to original files after all are computed. If this makes the implementation simpler and easier to read, probably worth doing

nemo_skills/code_execution/math_grader.py
```diff
@@ -12,535 +12,214 @@
...
+    input_files: list[str],
+    in_memory_lines=1500,
+    numeric_precision=15,
+    timeout=10,
+    take_modulo=None,
+    ignore_cache: bool = False,
```
Kipok, line 157: also if this runs very fast, we can remove the cache logic and always recompute. Easier to manage and one less parameter

nemo_skills/code_execution/math_grader.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: can you please move it to the evaluation folder? It was only here because sandbox depended on this file, but now that we removed it, we should move it to where it belongs

nemo_skills/pipeline/eval.py
```diff
@@ -176,6 +176,7 @@ def eval(
         "--not_exclusive",
         help="If --not_exclusive is used, will NOT use --exclusive flag for slurm",
     ),
+    with_sandbox: bool = typer.Option(False, help="If True, will start a sandbox container alongside this job"),
```
Kipok, line 179: Please update openmathreasoning eval docs to specify sandbox when running in TIR mode

Kipok (commented):
thanks, just a few small things

Kipok (approved):
please run more tests with evals and if all looks good we can merge

PR #455, opened 2025-05-09
[CodeGen] Refactor pipeline scripts for additional features

titu1994:
# Changelog
Primarily this PR adds two major things:
- support for dynamic mounting via str cli arg
- full featured `run_cmd`

Some of the scripts have an optional new argument `check_mounted_paths` which can perform cluster tunnel level checks to assert that some folders and paths exist. 

While codegen had far broader checks, going into individual input file level checks to assert that script cannot run until at least all filepaths are asserted to exist on cluster, we don't need to be so aggressive with checks in skills so I've kept it to minimal items and optional. I.e. auto mounting (mounting of absolute paths passed by user automatically) will only be done if this flag is enabled and only for a very specific set of inputs (like output dir, log dir and models sometimes).

**NOTE**: I still have to test each command. @Kipok  if you can share generic commands for them that would be good, or if its easier for you to run some tests yourself, let me know.

-----

# Enhancements to mounted path handling:
[GPT generated changelog]

* **New parameters for mounted paths:**
  - Added `mount_paths` and `check_mounted_paths` options to multiple pipeline functions (`check_contamination`, `convert`, `eval`, `generate`, and `run_cmd`) to allow specifying and validating paths to be mounted on remote machines. [[1]](diffhunk://#diff-3358d2230e92d60a6e4a72a94db50693501340b661d9dd7a2368fc310802804fR84) [[2]](diffhunk://#diff-3358d2230e92d60a6e4a72a94db50693501340b661d9dd7a2368fc310802804fR114) [[3]](diffhunk://#diff-7f42d537894a57a2a9a9b3681b793b6f988d6c80569fcd625dd6ad8c1867d138R184) [[4]](diffhunk://#diff-7f42d537894a57a2a9a9b3681b793b6f988d6c80569fcd625dd6ad8c1867d138R207) [[5]](diffhunk://#diff-58739ee434fd929a0c16a140b270be222c2367f0a7ed7f7fac6b492aa390408aR161) [[6]](diffhunk://#diff-58739ee434fd929a0c16a140b270be222c2367f0a7ed7f7fac6b492aa390408aR186-R193) [[7]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0R438) [[8]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0R498) [[9]](diffhunk://#diff-cae186e947820fc321b89b73412e2530f754ffbc31cf31e2581007f422d317a4R67-R83)
  - Introduced logic to resolve, validate, and handle mounted paths using utility functions like `resolve_mount_paths`, `is_mounted_filepath`, `add_mount_path`, and `check_remote_mount_directories`. [[1]](diffhunk://#diff-3358d2230e92d60a6e4a72a94db50693501340b661d9dd7a2368fc310802804fL124-R158) [[2]](diffhunk://#diff-7f42d537894a57a2a9a9b3681b793b6f988d6c80569fcd625dd6ad8c1867d138R236-R257) [[3]](diffhunk://#diff-58739ee434fd929a0c16a140b270be222c2367f0a7ed7f7fac6b492aa390408aL197-R231) [[4]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0R535-R559)

* **Cluster configuration updates:**
  - Updated cluster configuration handling to integrate mounted path resolution and ensure proper directory creation on remote machines when required. [[1]](diffhunk://#diff-3358d2230e92d60a6e4a72a94db50693501340b661d9dd7a2368fc310802804fL124-R158) [[2]](diffhunk://#diff-7f42d537894a57a2a9a9b3681b793b6f988d6c80569fcd625dd6ad8c1867d138R236-R257) [[3]](diffhunk://#diff-58739ee434fd929a0c16a140b270be222c2367f0a7ed7f7fac6b492aa390408aL197-R231) [[4]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0R535-R559)

### Code refactoring and cleanup:

* **Removed redundant methods:**
  - Eliminated the `wrap_cmd` function from `generate.py` and replaced it with a more comprehensive `get_cmd` function in `run_cmd.py`, which now handles command construction with support for preprocessing and postprocessing steps. [[1]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0L344-L355) [[2]](diffhunk://#diff-cae186e947820fc321b89b73412e2530f754ffbc31cf31e2581007f422d317a4L16-R45)

* **Simplified CLI utilities:**
  - Moved the `wrap_arguments` function from `cli.py` to `utils`, streamlining the CLI setup and maintaining consistency across modules. [[1]](diffhunk://#diff-2b5023e075231063a0b2fd8e4da68ddfbdb8ef9a0a6d19d7694356824d0e32f2R18-R21) [[2]](diffhunk://#diff-2b5023e075231063a0b2fd8e4da68ddfbdb8ef9a0a6d19d7694356824d0e32f2L36-L48)

* **Improved modularity:**
  - Consolidated utility imports across pipeline modules, ensuring all required functions for mounted path handling are consistently available. [[1]](diffhunk://#diff-3358d2230e92d60a6e4a72a94db50693501340b661d9dd7a2368fc310802804fR31-R36) [[2]](diffhunk://#diff-7f42d537894a57a2a9a9b3681b793b6f988d6c80569fcd625dd6ad8c1867d138L23-R35) [[3]](diffhunk://#diff-58739ee434fd929a0c16a140b270be222c2367f0a7ed7f7fac6b492aa390408aR33-R49) [[4]](diffhunk://#diff-2bf0a3d00b2c8bc23a7ee220ddce41ecd8d4428efc4404c6b8e6857ec91c6ca0L29-R43)

### Minor adjustments:

* Added a new `ExtraDatasetType` enum in `eval.py` for potential future enhancements related to handling extra datasets.


nemo_skills/pipeline/run_cmd.py
```diff
@@ -76,32 +110,87 @@ def run_cmd(
         "--not_exclusive",
         help="If --not_exclusive is used, will NOT use --exclusive flag for slurm",
     ),
+    check_mounted_paths: bool = typer.Option(False, help="Check if mounted paths are available on the remote machine"),
```
titu1994, line 113: Btw, id like this flag to be true in general to offer users some safety. But that can be discussed.

Kipok: The problem with having it True is that when you schedule large pipelines, it's often the case that subsequent steps take as input files generated by previous steps. So at scheduling time, they don't exist and thus will cause failure. But they will be available by the time the job runs because of slurm dependencies

titu1994: I see that's a fair point to keep it disabled by default.

titu1994:
The other thing is I would normally like to do the check for model paths + auto Mount if not mounted, but HF support makes the code gnarly, so for now I'm not porting that from codegen.

nemo_skills/pipeline/check_contamination.py
```diff
@@ -121,10 +130,28 @@ def check_contamination(
...
+        if not is_mounted_filepath(cluster_config, output_file):
+            output_dir, output_filename = os.path.split(output_file)
+            add_mount_path(output_dir, "/mounted_data/output", cluster_config)
+
+    input_file = get_mounted_path(cluster_config, input_file)
+    output_file = get_mounted_path(cluster_config, output_file)
+
     if log_dir:
-        check_if_mounted(cluster_config, log_dir)
+        if check_mounted_paths: create_remote_directory(log_dir, cluster_config)
```
Kipok, line 148:
```suggestion
        if check_mounted_paths: 
            create_remote_directory(log_dir, cluster_config)
```

nemo_skills/pipeline/check_contamination.py
```diff
@@ -121,10 +130,28 @@ def check_contamination(
...
+        log_dir = get_mounted_path(cluster_config, log_dir)
+
+    if check_mounted_paths:
+        # Final check for existance of mounted paths
```
Kipok, line 152:
```suggestion
        # Final check for existence of mounted paths
```

nemo_skills/pipeline/check_contamination.py
```diff
@@ -121,10 +130,28 @@ def check_contamination(
     get_random_port = server_gpus != 8 and not exclusive
 
     cluster_config = get_cluster_config(cluster, config_dir)
-    check_if_mounted(cluster_config, input_file)
```
Kipok, line 124: Can we keep this logic in the "else" part of `if check_mounted_paths`?

nemo_skills/pipeline/utils.py
```diff
@@ -255,6 +267,12 @@ def create_remote_directory(directory: str | list, cluster_config: dict):
     if isinstance(directory, str):
         directory = [directory]
 
+    # Get unmounted path of all directories
+    directory = [
+        get_unmounted_path(cluster_config, dir_path) if is_mounted_filepath(cluster_config, dir_path) else dir_path
```
Kipok, line 272: isn't the if part redundant? I think the `get_unmounted_path` already returns original if it's not mounted

nemo_skills/pipeline/utils.py
```diff
@@ -332,6 +350,13 @@ def check_remote_mount_directories(directories: list, cluster_config: dict, exit
         raise ValueError("Cluster config is not provided.")
     if isinstance(directories, str):
         directories = [directories]
+
+    # Get unmounted path of all directories
+    directories = [
+        get_unmounted_path(cluster_config, dir_path) if is_mounted_filepath(cluster_config, dir_path) else dir_path
```
Kipok, line 356: isn't the if part redundant? I think the get_unmounted_path already returns original if it's not mounted

nemo_skills/pipeline/check_contamination.py
Kipok: This applies to all pipeline scripts

nemo_skills/pipeline/convert.py
```diff
@@ -219,14 +233,28 @@ def convert(
...
         raise ValueError("Conversion to HF/Nemo is not yet supported for DeepSeek v3 models")
 
+    # Prepare cluster config and mount paths
     cluster_config = get_cluster_config(cluster, config_dir)
-    check_if_mounted(cluster_config, input_model)
-    check_if_mounted(cluster_config, output_model)
+    cluster_config = resolve_mount_paths(cluster_config, mount_paths)
+
+    if check_mounted_paths:
+        if not is_mounted_filepath(cluster_config, input_model): add_mount_path(input_model, "/input_model", cluster_config)
```
Kipok, line 241: best not to use single line ifs, harder to read and also our current black setup in precommit will reformat this back to 2 lines anyway. And in general, best to setup precommit to ensure your code is automatically formatted

nemo_skills/pipeline/eval.py
```diff
@@ -197,12 +211,25 @@ def eval(
         pass
 
     cluster_config = get_cluster_config(cluster, config_dir)
-    check_if_mounted(cluster_config, output_dir)
+    cluster_config = resolve_mount_paths(cluster_config, mount_paths)
```
Kipok, line 214: Feels like we can have a utility that does all of this and takes all boolean arguments as well as all files to check and log_dir and then just call it as one line in all pipelines to avoid duplication

nemo_skills/pipeline/utils.py
```diff
@@ -758,6 +848,52 @@ def get_cluster_config(cluster=None, config_dir=None):
...
+        # Resolve `identity` from env if not provided
+        cluster_config['ssh_tunnel']['identity'] = os.path.expandvars(cluster_config['ssh_tunnel']['identity'])
+        LOG.info(f"Resolved `identity` to `{cluster_config['ssh_tunnel']['identity']}`")
+
+        if "$" in cluster_config['ssh_tunnel']['identity']:
+            raise ValueError(
+                f"Identity file path could not be loaded from env variable `{cluster_config['ssh_tunnel']['identity']}`"
+            )
+
+    # Add mount container list if not present
```
Kipok, line 890: feels out of place in this function as from the name it only supposed to update ssh section

nemo_skills/pipeline/utils.py
```diff
@@ -758,6 +848,52 @@ def get_cluster_config(cluster=None, config_dir=None):
...
+        return cluster_config
+
+    if 'user' in cluster_config['ssh_tunnel']:
+        # Resolve `user` from env if not provided
+        if cluster_config['ssh_tunnel']['user'] is None:
+            cluster_config['ssh_tunnel']['user'] = os.environ['USER']
+            LOG.info(f"Resolved `user` to `{cluster_config['ssh_tunnel']['user']}`")
+
+        elif isinstance(cluster_config['ssh_tunnel']['user'], str) and '$' in cluster_config['ssh_tunnel']['user']:
+            cluster_config['ssh_tunnel']['user'] = os.path.expandvars(cluster_config['ssh_tunnel']['user'])
```
Kipok, line 872: can we just have some recursive logic to do this for all variables, not just ssh tunnel? That would make things more consistent and maybe there is already some library we can reuse that does that. But if it's hard, ok to keep it like this

Kipok: and if not, then at least this function can be much shorter by having a for loop over all variables and same logic inside the loop, instead of current copy-paste

nemo_skills/pipeline/utils.py
```diff
@@ -758,6 +848,52 @@ def get_cluster_config(cluster=None, config_dir=None):
...
+        # Resolve `job_dir` from env if not provided
+        cluster_config['ssh_tunnel']['job_dir'] = os.path.expandvars(cluster_config['ssh_tunnel']['job_dir'])
+        LOG.info(f"Resolved `job_dir` to `{cluster_config['ssh_tunnel']['job_dir']}`")
+
+    if 'identity' in cluster_config['ssh_tunnel']:
+        # Resolve `identity` from env if not provided
+        cluster_config['ssh_tunnel']['identity'] = os.path.expandvars(cluster_config['ssh_tunnel']['identity'])
+        LOG.info(f"Resolved `identity` to `{cluster_config['ssh_tunnel']['identity']}`")
+
+        if "$" in cluster_config['ssh_tunnel']['identity']:
```
Kipok, line 885: why is special logic needed here?

nemo_skills/utils.py
```diff
@@ -500,3 +500,44 @@ def resolve_python_module_from_file(py_filepath: str, root_module: str = 'nemo_s
     striped_module = str(stripped_module_path).replace("/", ".")
     striped_module = os.path.splitext(striped_module)[0]
     return striped_module
+
+
+def maybe_get_env(value: Union[int, List[int]], env_name, default=None, cast: Callable[[Any], Any] = None):
```
Kipok, line 505: type of value is not correct? Best not to provide type, rather than have incorrect one

nemo_skills/pipeline/utils.py
```diff
@@ -657,6 +682,67 @@ def get_sandox_command(cluster_config):
     return "/entrypoint.sh && /start.sh"
 
 
+def configure_client(
```
Kipok, line 685: Probably best to keep this in utils as well (we really should split that file though) and then make use of it inside generate.py, eval.py, start_server.py

nemo_skills/pipeline/run_cmd.py
```diff
@@ -13,21 +13,36 @@
 # limitations under the License.
 
 import logging
-from typing import List
+from typing import List, Optional
```
Kipok, line 16: Will discuss changes to this file offline

Kipok (commented):
Some comments + tests currently fail, so would need to fix those

nemo_skills/pipeline/check_contamination.py
titu1994: Just a note, in places where `get_mounted_path` or `get_unmounted_path` exists, it implicitly calls check_if_mounted() - or it will raise an error saying the path is not mounted. So i removed the check_if_mounted() cause it was redundant in those sections

nemo_skills/pipeline/utils.py
titu1994: Done

titu1994:
Addressed most of the comments

nemo_skills/pipeline/run_cmd.py
```diff
@@ -40,6 +45,7 @@ def run_cmd(
         help="One of the configs inside config_dir or NEMO_SKILLS_CONFIG_DIR or ./cluster_configs. "
         "Can also use NEMO_SKILLS_CONFIG instead of specifying as argument.",
     ),
+    command: str = typer.Option(None, help="Command to run in the container"),
```
Kipok, line 49:
```suggestion
    command: str = typer.Option(None, help="Command to run in the container. Can also be specified as extra arguments."),
```

nemo_skills/pipeline/run_cmd.py
```diff
@@ -63,9 +77,8 @@ def run_cmd(
         help="If specified, will reuse the code from this experiment. "
         "Can provide an experiment name or an experiment object if running from code.",
     ),
-    preprocess_cmd: str = typer.Option(None, help="Command to run before job"),
-    postprocess_cmd: str = typer.Option(None, help="Command to run after job"),
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
+    with_sandbox: bool = typer.Option(False, help="Whether to use the sandboxing feature to host the model"),
```
Kipok, line 84:
```suggestion
    with_sandbox: bool = typer.Option(False, help="If True, will start a sandbox container alongside this job"),
```

nemo_skills/pipeline/run_cmd.py
```diff
@@ -13,21 +13,26 @@
 # limitations under the License.
 
 import logging
-from typing import List
+from typing import List, Optional
```
Kipok, line 16: Would be great to update documentation over here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/pipelines/run-cmd.md to explain the new server arguments and how the command interacts with them

nemo_skills/pipeline/generate.py
```diff
@@ -464,6 +448,7 @@ def generate(
         "If not specified, will use the default entrypoint for the server type.",
     ),
     dependent_jobs: int = typer.Option(0, help="Specify this to launch that number of dependent jobs"),
+    mount_paths: str = typer.Option(None, help="Comma separated list of paths to mount on the remote machine"),
```
Kipok, line 451: Please add some description of these new common parameters over here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/pipelines/index.md?plain=1#L76

nemo_skills/pipeline/utils.py
Kipok: oh, but there is still configure_client in pipeline/generate. Should that one be removed then? Also can it be directly used in eval.py and check_contamination.py?

titu1994: Hmm that will requite a more refactors. I think it should be a drop in replacement, however the utils one has extra arg `get_random_port` and the generate one has extra unused arg (generation type and server port). So Id like to keep this as a separate pr item (or you can update either later on).

titu1994: But yes technically all 3 server configure calls can be merged into the utils one (if it accepts the union of all extra args from each script?)

docs/pipelines/run-cmd.md
```diff
@@ -14,4 +14,49 @@ code and tries to install it (will finish with error if not running from NeMo-Sk
...
+)
+print(response.choices[0].message.content)
+```
+
+Then we can run the server and the inference code in a single command:
+
+```bash
+ns run_cmd \
+    --cluster=local \
+    --model=/hf_models/Qwen3-1.7B \
```
Kipok, line 57: thanks, great example! Consider replacing this with `Qwen/Qwen3-1.7B` which will make it runnable without a prerequisite for having a model downloaded
```suggestion
    --model=Qwen/Qwen3-1.7B \
```

docs/pipelines/run-cmd.md
```diff
@@ -14,4 +14,49 @@ code and tries to install it (will finish with error if not running from NeMo-Sk
...
+    --server_type=vllm \
+    --server_gpus=1 \
+    --with_sandbox \
```
Kipok, line 60:
```suggestion
```

docs/pipelines/run-cmd.md
```diff
@@ -14,4 +14,49 @@ code and tries to install it (will finish with error if not running from NeMo-Sk
...
+## LLM Server and Sandbox Server
+
+While we can run arbitrary commands with the default `run_cmd` script, we also provide the ability to
+run a LLM server with the `--model` argument and a few extra arguments for the server config. These arguments 
+are similar to the ones used for `start_server` script.
+
+This can be useful to run a server on a local machine or on a cluster with GPUs in a slurm job, while also being able to
+run arbirary code that uses LLM calls. 
+
+### Example
```
Kipok, line 28: would be nice to use this example (or something similar) in the gpu tests to make sure this new functionality is checked in ci. Can be done in another PR, but please create an issue to track this if not including here

docs/pipelines/index.md
```diff
@@ -88,6 +88,18 @@ All scripts inside pipeline folder have the following parameters.
 - **--run_after**: Can be used in conjunction with `--expname` to chain jobs to
   run one after another (only applicable on slurm). E.g. run training job with
   `--expname my-training-run` and then launch an eval with `--run_after my-training-run`.
+- **--mount_paths**: Can be used to mount additional paths to the cluster config dynamically. 
+  This is useful if you want to access some data that is not mounted in cluster config. E.g. use
+  `--mount_paths /my/remote/workspace:/workspace` to mount `/workspace` folder from the host
+  machine to the slurm job.
+- **--check_mounted_paths**: This flag offers a few different capabilities for convenience:
+  - Check if the paths specified in the script are mounted correctly. This is useful if you want to make
```
Kipok, line 96: this doesn't display properly (not a nested list). You can run `mkdocs serve` to get the docs built locally and see how it's visible on the website

docs/pipelines/index.md
```diff
@@ -88,6 +88,18 @@ All scripts inside pipeline folder have the following parameters.
 - **--run_after**: Can be used in conjunction with `--expname` to chain jobs to
   run one after another (only applicable on slurm). E.g. run training job with
   `--expname my-training-run` and then launch an eval with `--run_after my-training-run`.
+- **--mount_paths**: Can be used to mount additional paths to the cluster config dynamically. 
+  This is useful if you want to access some data that is not mounted in cluster config. E.g. use
+  `--mount_paths /my/remote/workspace:/workspace` to mount `/workspace` folder from the host
+  machine to the slurm job.
+- **--check_mounted_paths**: This flag offers a few different capabilities for convenience:
```
Kipok, line 95: this description is a little confusing to me. E.g. sometimes it will raise an error if the path doesn't exist, but sometimes it will create a folder for me. How do I know when it's one or the other?

nemo_skills/pipeline/check_contamination.py
```diff
@@ -126,10 +135,14 @@ def check_contamination(
...
-    check_if_mounted(cluster_config, output_file)
-    if log_dir:
-        check_if_mounted(cluster_config, log_dir)
+    cluster_config = resolve_mount_paths(cluster_config, mount_paths)
```
Kipok, line 138: maybe make it a part of get_cluster_config (if mount_paths specified we resolve them)?

nemo_skills/pipeline/run_cmd.py
```diff
@@ -76,32 +99,80 @@ def run_cmd(
         "--not_exclusive",
         help="If --not_exclusive is used, will NOT use --exclusive flag for slurm",
     ),
+    get_random_port: bool = typer.Option(False, help="If True, will get a random port for the server"),
```
Kipok, line 102: Why do we need this parameter? E.g. all other scripts don't have it and I think the default logic of doing

```
get_random_port = server_gpus != 8 and not exclusive
```

covers majority of possible cases? If you really think we should have it, we should then add it to all scripts, not just this one

nemo_skills/pipeline/start_server.py
```diff
@@ -62,20 +75,28 @@ def start_server(
...
     try:
         server_type = server_type.value
     except AttributeError:
         pass
 
+    log_dir = check_remote_mounts(cluster_config, log_dir, check_mounted_paths=check_mounted_paths)
+
     if log_dir:
-        check_if_mounted(cluster_config, log_dir)
+        if check_mounted_paths:
```
Kipok, line 95: isn't this done in check_remote_mounts already?

nemo_skills/pipeline/utils.py
```diff
@@ -131,6 +143,82 @@ def check_if_mounted(cluster_config, path_to_check):
         raise ValueError(f"The path '{path_to_check}' is not mounted. Check cluster config.")
 
 
+def check_remote_mounts(
```
Kipok, line 146: Better to name it just check_mounts as this is also applicable when running locally and specifying config file

nemo_skills/pipeline/utils.py
```diff
@@ -131,6 +143,82 @@ def check_if_mounted(cluster_config, path_to_check):
...
+    # Compute directory of all files
+    remote_dir_list = []
+
+    # Check paths and add to mount list if not mounted
+    if check_mounted_paths:
+        for idx, (path, default_mount) in enumerate(mount_map.items()):
+            if not is_mounted_filepath(cluster_config, path):
+                # check if the path is a file or a directory
+                # so that the directory can be created
+                is_file = os.path.splitext(path) != ""
```
Kipok, line 178: this will fail on any files without extension? Can also send a proper check command through the ssh tunnel, but if you think this is enough, I guess it's ok. But can lead to silent issues later on

nemo_skills/pipeline/utils.py
Kipok: ok, please add a TODO or an issue for this

nemo_skills/pipeline/utils.py
```diff
@@ -782,6 +944,10 @@ def get_cluster_config(cluster=None, config_dir=None):
 
     cluster_config = read_config(config_file)
 
+    # resolve ssh tunnel config
```
Kipok, line 947: should we move this to get_cluster_config function?

docs/pipelines/run-cmd.md
titu1994: This is intented to show that you can add sandbox with one flag.

nemo_skills/pipeline/check_contamination.py
titu1994: This needs to be separate. Otherwise we have to go back and update all get_cluster_config with optional arg resolve mount paths. And on top of that, in python mode, you should call get cluster config for "base config" and then run custom dicts / str of resolve mount paths per task you're executing

nemo_skills/pipeline/run_cmd.py
titu1994: In run cmd, i want it to be fully flexible for dev to control their env. The above logic is fine for our slurm env, may not be the case for others.

nemo_skills/pipeline/utils.py
titu1994: Done

docs/pipelines/run-cmd.md
Kipok: Then it's better to add some explanation. Otherwise it appears as if this is required for the example you have (while it's actually not used at all). I'd rather not have it in command and say in another sentence that you can do this

nemo_skills/pipeline/utils.py
titu1994: Its meant to be a dir check - if is a file a filepath with extension or a dir path with no extension. I tested it in python 3.10 - works fine.

nemo_skills/pipeline/utils.py
titu1994: This is part of get_cluster_config() ?

nemo_skills/pipeline/start_server.py
titu1994: Good catch, thanks !

nemo_skills/pipeline/run_cmd.py
titu1994: As to other scripts - it depends. Do we want users to bother with setting this for the most common use cases (like our clusters?) - probably not. So I'm in favor of keeping those automatic. This one is too generic of a script - broader responsibility than just our clusters. So it makes sense to be more flexible in this one.

nemo_skills/pipeline/utils.py
Kipok: right, but this will treat any files without extension as a directory and try to create it. Then it's unclear how the downstream code will interact with that and might lead to silent errors. We don't normally use file names without extensions, but it's certainly possible to do that, right?

nemo_skills/pipeline/utils.py
Kipok: oh, sorry, you're right

docs/pipelines/index.md
titu1994: Will do in another pr

docs/pipelines/run-cmd.md
titu1994: Will do

docs/pipelines/index.md
titu1994: Will fix in a second pr for nits. That wont need GPU exec hopefully.

titu1994: The nain thing is `--check_mount_paths` is the flag that enables remote file and dir checks. When not set, we do just local checks of whether the path "looks correct" and mounted. When this flag is set, we confirm with remote calls to the cluster, that indeed the paths are in correct location or not. 

This is set to false by default because it is fine for one off scripts but in a complex chain of scripts - you don't want to assert on existence of data when a previous pipeline stage is going to generate that data file.

nemo_skills/pipeline/utils.py
titu1994: Yes possible to do that.. Hmm this part is actually only necessary for 2 scripts that try to create output directory. If i manually do it in those two scripts, then i can skip this logic here. Let me change that in another pr.

docs/pipelines/run-cmd.md
```diff
@@ -14,4 +14,54 @@ code and tries to install it (will finish with error if not running from NeMo-Sk
...
+    top_p=0.95,
+    n=1,
+    stream=False,
+)
+print(response.choices[0].message.content)
+```
+
+Then we can run the server and the inference code in a single command as below. We will also show how to start the code 
+execution server that can be used to run arbitrary code in a sandboxed environment. This can be useful to safely 
+execute untrusted code or to run code that requires a specific environment in a container without affecting the host
```
Kipok, line 54: I think we need to tone this down, our sandbox is certainly not safe to run untrusted code, and we shouldn't say it in the docs :) It's a little safer than running it without sandbox, but not safe for sure (which we mention here https://nvidia.github.io/NeMo-Skills/basics/sandbox/)

titu1994: Ah ok

Kipok (approved):
Thanks! Please wait for cpu tests to finish and feel free to merge after that

titu1994:
Thanks for the reviews and testing !

nemo_skills/pipeline/utils.py
titu1994: Made it just a check and raise error in another PR.

PR #456, opened 2025-05-09
Packages needed for local execution

darraghdog:
For local execution with sandbox we miss these two packages, `ipython` & `flask`.  
`ipython` will silently fail.
We can either add this to the main requirements so it installs with `pip install git+https://github.com/NVIDIA/NeMo-Skills.git`  or alternatively instruct the user to install separately.


Kipok (approved):
thanks!

PR #461, opened 2025-05-13
Fix answer judge

smahdavi4:
This pull-request adds a third state to the answer judge when the model does not produce any judgement.

Also, AnswerJudgementMetrics is refactored to account for invalid judgements, and bug fix for majority voting and greedy metrics.

### DIFF
--- a/nemo_skills/evaluation/metrics/answer_judgement_metrics.py
+++ b/nemo_skills/evaluation/metrics/answer_judgement_metrics.py
@@ -47,9 +47,9 @@ class AnswerJudgementMetrics(BaseMetrics):
         is_invalid = pred_judgement is None
         is_correct = pred_judgement == gt_judgement
         if not is_correct:
-            if pred_judgement:
+            if pred_judgement == True:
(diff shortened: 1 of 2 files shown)


Kipok (approved):
Looks great, thanks! The only thing you might consider adding is to log all metric@k numbers for all k <= num samples. Could be useful for plotting things like majority / pass @k and checking when they saturate. But not necessary to do in this PR, can send a follow up if you want to add that. You can see an example how that's done in math metrics

smahdavi4:
Yeah that's a good idea. I just added that.

Kipok:
You probably want to pair it with adding a `max_aggregations_to_print` logic like done here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/evaluation/metrics/math_metrics.py#L321. That way the table displayed by summarized_results will only have the largest K value, but metrics.json saved file, will have all, so they are still accessible but don't clutter the table

smahdavi4:
Makes sense. I just added that.

PR #463, opened 2025-05-13
Dh/fp8 v01

darraghdog:
- Allow fp8 quantization.
- Providing calibration dataset with `text` field is mandatory as input. 
- File `nemo_skills/conversion/hf_to_trtllm_quantize.py` is copied directly from `TensorRT-LLM/examples/quantization/quantize.py`

Usage,
```
ns convert --input_model OpenMath-Nemotron-1.5B \
           --output_model OpenMath-Nemotron-1.5B-Kaggle-fp8-trtllm \
           --convert_from hf \
           --convert_to trtllm \
           --num_gpus 2 \
           --dtype fp8 \
           --hf_model_name nvidia/OpenMath-Nemotron-1.5B \
           --model_type qwen \
           --calib_dataset calibrate_openmathreasoning
```

### DIFF
--- a/nemo_skills/conversion/hf_to_trtllm_quantize.py
+++ b/nemo_skills/conversion/hf_to_trtllm_quantize.py
@@ -1,3 +1,22 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Copied from NVIDIA's TensorRT-LLM project:
+# https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/quantization/quantize.py
+# Original commit: 5aec7af45fc0abd876fa68a9ae8c8cae084f3af3
+# Original version: v0.18
+
 import argparse
 
 import torch.multiprocessing as mp
--- a/nemo_skills/conversion/hf_to_trtllm_quantize.py
+++ b/nemo_skills/conversion/hf_to_trtllm_quantize.py
@@ -1,0 +1,204 @@
+import argparse
+
+import torch.multiprocessing as mp
+
+from tensorrt_llm.quantization import (quantize_and_export,
+                                       quantize_nemo_and_export)
+
+mp.set_start_method("spawn", force=True)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--model_dir",
+                        help="Specify where the HuggingFace model is",
+                        default=None)
+    parser.add_argument('--nemo_ckpt_path',
+                        help="Specify where the NeMo checkpoint is",
+                        default=None)
+    parser.add_argument(
+        '--decoder_type',
+        type=str,
+        default='gptnext',
+        choices=['gptnext', 'llama'],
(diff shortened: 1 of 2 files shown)


nemo_skills/pipeline/convert.py
```diff
@@ -59,36 +61,64 @@ def get_hf_to_trtllm_cmd(
...
-        f"    --max_seq_len 8192 "
-        f"    --max_num_tokens 8192 "
-        f"    {extra_arguments} && "
-        f"cp {input_model}/tokenizer* {output_model} "
-    )
+    if dtype == "fp8":
+        hf_to_trtllm_cmd = (
+            f"python -m nemo_skills.conversion.hf_to_trtllm_quantize "
+            f"    --model_dir {input_model} "
+            f"    --dtype float16 "
```
Kipok, line 75:
```suggestion
```
Is it better to use a default "auto" parameter that infers from original model?

nemo_skills/pipeline/convert.py
```diff
@@ -209,6 +248,10 @@ def convert(
     except AttributeError:
         pass
 
+    # Validate dtype-related requirements
```
Kipok, line 251: Can you please add some checks that if convert_to is not "trtllm", but dtype set to fp8, we throw an error saying that it's not supported yet?

nemo_skills/conversion/hf_to_trtllm_quantize.py
```diff
@@ -0,0 +1,204 @@
+import argparse
```
Kipok, line 1: Can you please add license header like in other files and also a comment saying where this script is copied from, so that it's easier to update in the future?

nemo_skills/pipeline/convert.py
darraghdog: For non-ReDrafter version this would be best, but ReDrafter failed with it. I'll put as an input param with default "auto"

nemo_skills/conversion/hf_to_trtllm_quantize.py
darraghdog: Added, note, this may tie the user to a specific tensorrt_llm version, but I have found in my own usage, using a different version of this script did not break things.

nemo_skills/pipeline/convert.py
darraghdog: Thats added and tested now.

nemo_skills/pipeline/convert.py
darraghdog: Seconds thoughts, probably best not to build around a ReDrafter limitation - its change ow to auto; and we can specify in the inference example to use float16 with ReDrafter.

nemo_skills/pipeline/convert.py
```diff
@@ -209,6 +248,17 @@ def convert(
...
+    if dtype == "fp8" and not calib_dataset:
```
Kipok, line 252:
```suggestion
```

nemo_skills/pipeline/convert.py
```diff
@@ -209,6 +248,17 @@ def convert(
...
+        raise ValueError("--calib_dataset is required when dtype is 'fp8'")
```
Kipok, line 253:
```suggestion
```

nemo_skills/pipeline/convert.py
```diff
@@ -209,6 +248,17 @@ def convert(
     except AttributeError:
         pass
 
+    # Validate dtype-related requirements
```
Kipok, line 251:
```suggestion
```

nemo_skills/pipeline/convert.py
```diff
@@ -209,6 +248,17 @@ def convert(
...
+    if dtype == "fp8" and not calib_dataset:
+        raise ValueError("--calib_dataset is required when dtype is 'fp8'")
+
```
Kipok, line 254:
```suggestion
```

Kipok (approved):
Thanks! There is a redundant check for calibration dataset parameter, but other than that looks good

darraghdog:
> Thanks! There is a redundant check for calibration dataset parameter, but other than that looks good

Thanks, I see you removed that redundant check; if not lmk and I'll fix it up.

PR #464, opened 2025-05-13
Chat interface script

i-vainn:
This PR adds script that launches gradio chat interface with our models.


docs/basics/chat_interface.md
```diff
@@ -0,0 +1,131 @@
...
+
+The chat interface provides a web UI where you can interact with a model. When launched with code execution capabilities, the model can use a Python interpreter to solve problems.
+
+## Launching with Docker
+
+The simplest way to run the chat interface is using Docker.
+
+### Prerequisites
+
+- Docker installed on your system
```
Kipok, line 15:
```suggestion
- [NVIDIA container toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) is installed on your system
```

docs/basics/chat_interface.md
```diff
@@ -0,0 +1,131 @@
...
+WORKDIR /app/
+RUN chown -R 1000:1000 /app
+
+RUN git clone https://github.com/NVIDIA/NeMo-Skills \
+    && cd NeMo-Skills \
+    && pip install --ignore-installed blinker \
+    && pip install -e . \
+    && pip install gradio \
+    # uncomment if plan to use model that executes Python
+    # && pip install -r requirements/code_execution.txt
```
Kipok, line 34: Can we do it always? Is there any downside?

docs/basics/chat_interface.md
```diff
@@ -0,0 +1,131 @@
...
+ENTRYPOINT [ "/app/entrypoint.sh" ]
+```
+
+2. Create an entrypoint.sh script:
+
+```bash
+#!/bin/bash
+
+# start model server with sandbox inside current container
+CUDA_VISIBLE_DEVICES=0,1 ns start_server \
```
Kipok, line 49:
```suggestion
ns start_server \
```
is that required? If not, best to remove it and just control with server_gpus parameter

nemo_skills/inference/launch_chat_interface.py
```diff
@@ -0,0 +1,214 @@
+import argparse
```
Kipok, line 1: license header

docs/basics/chat_interface.md
```diff
@@ -0,0 +1,131 @@
+# Chat Interface
```
Kipok, line 1: Please expose this in documentation (probably linking through basics/inference under a new section) as well as in https://github.com/NVIDIA/NeMo-Skills/blob/main/mkdocs.yml

Kipok (commented):
Thanks! Can we make a wrapper script that will make it simpler to use and support all the logic without users manually switching between code_execution / different number of gpus, etc. Or instead of a separate script, maybe an option to start_server `--launch_chat_interface` that when specified will add an extra command that starts the chat interface script? That way we don't need to have a separate docker and can reuse our existing functionality to launch containers. And if anyone wants to run inside existing docker, it will still work as long as cluster config is not specified. Plus this way we can specify all parameters in one place and not need to duplicate them in start_server and chat interface (e.g. to enable code execution).

Another thing I would add is an option to use predefined models, so people can just say --model=Nvidia/OpenMath-Nemotron-X and then we will fill in the correct config / template inside the gradio demo, so that it doesn't need to be specified. This will also make it simpler to avoid errors by specifying incorrect config (e.g. our kaggle model uses different config than non-kaggle ones for tir). Finally, the gradio should ideally support a switch between code execution and non-code execution with a button, not an argument as some of the models can be used in both modes (and we can disable it for the kaggle model, if we can recognize that it's used)

nemo_skills/inference/launch_chat_interface.py
```diff
@@ -22,14 +22,31 @@ def parse_args() -> argparse.Namespace:
         default="localhost",
         help="Hostname where both the model server and the execution sandbox are running.",
     )
+    parser.add_argument(
+        "--ssh_server",
```
Kipok, line 26: are these used anywhere? Best to also accept from our standard env vars for this

nemo_skills/inference/launch_chat_interface.py
```diff
@@ -22,14 +22,31 @@ def parse_args() -> argparse.Namespace:
...
+    )
     parser.add_argument(
         "--server_type",
         default="vllm",
         choices=["vllm", "sglang", "trtllm"],
         help="Type of the model server to use.",
     )
     parser.add_argument(
-        "--with_code_execution",
+        "--preset",
```
Kipok, line 42: as discussed offline, let's change this by reading model config

nemo_skills/pipeline/start_server.py
```diff
@@ -20,6 +20,16 @@
 from nemo_skills.utils import setup_logging
 
 
+def get_gradio_chat_cmd(server_type, with_sandbox, extra_args):
+    cmd = (
+        "python -m nemo_skills.inference.launch_chat_interface "
+        f"   --server_type {server_type} "
+        f"   --with_sandbox {with_sandbox} "
+        f"   {extra_args} "
```
Kipok, line 28: as discussed offline, let's control all arguments from web interface itself, not here

nemo_skills/pipeline/start_server.py
```diff
@@ -16,10 +16,25 @@
 import typer
 
 from nemo_skills.pipeline.app import app, typer_unpacker
-from nemo_skills.pipeline.utils import add_task, check_if_mounted, get_cluster_config, get_exp, get_free_port
+from nemo_skills.pipeline.utils import (
+    add_task, check_if_mounted,
```
Kipok, line 20: recommend to setup precommit hook, so that things like this are formatted by black

Kipok (commented):
Thanks! Just a few small comments and can you also add an image or a small gif with how this looks? And another thing is you can mention this on top of the evaluation page for the openmathreasoning documentation, so people know that besides the benchmark numbers, it's also very easy to "talk" with the models for both tir and cot

Kipok (approved):
Thanks!

PR #465, opened 2025-05-14
Add a common implementation of pass@k and majority@k in the `BaseMetrics` class

shan18:
*Note: This PR needs some changes before it can be merged.*
*What needs to be fixed: Reordering of metrics correctly in `max_aggregations_to_print` such that they show up correctly with `ns summarize_results`*

### Summary

* New `get_prediction_results` abstract method in `BaseMetrics`
* Two new functions: `get_majority_at_k` and `get_pass_at_k` in `BaseMetrics` that compute the pass@k and majority@k metrics.
  - These functions get results from the `get_prediction_results` to compute the scores.
* Updated the evaluation code of the math, code, ifeval, and lean4 benchmarks to use these classes for computing metrics.


nemo_skills/evaluation/metrics/base.py
```diff
@@ -29,6 +30,92 @@ def get_metrics(self):
     def reset(self):
         pass
 
+    @abc.abstractmethod
+    def get_prediction_results(self, prediction):
+        pass
+
+    def get_majority_at_k(
+        self, agg_mode_dict, predicted_answers, pred_keys=None, predictions=None, prediction_results=None
```
Kipok, line 38: do we need prediction_results parameter? I see it only used in math metrics but seems that the provided value is the same as what is computed in this function anyway?

Kipok, line 38: I would probably also remove pred_keys and always assume that whatever is returned in get_prediction_results is what we should use. I think you only use pred_keys in math metrics and if metrics. In math metrics, from what I can see it might be possible to just completely remove it? Maybe I'm missing something, but seems that it will already contain exactly the right keys based on what's available for judge/sympy metrics and then we can just aggregate over all of them. And for the if metrics, to support that use-case, I'd instead change the self._get_best_prediction signature to also accept full predictions list and then it will have access to any parameters it needs. Let me know if I miss something here

Kipok, line 38: For agg_mode_dict and self._get_best_prediction, I would just use those methods from main class. So you can define get_best_prediction to have the general logic in the base class and then subclasses can override. And agg_mode_dict is anyway used everywhere, so let's just move it to the base class and use directly instead of passing as a parameter?

nemo_skills/evaluation/metrics/base.py
```diff
@@ -29,6 +30,92 @@ def get_metrics(self):
     def reset(self):
         pass
 
+    @abc.abstractmethod
+    def get_prediction_results(self, prediction):
+        pass
+
+    def get_majority_at_k(
```
Kipok, line 37: also for majority, isn't it only applicable to math metrics? If that's the case, we can probably move this function over there and also simplify the signature (e.g. we can remove predicted_answers) and take them from predictions assuming the "predicted_answer" key is used. But if you know other use-cases that support majority, then we can keep it here

nemo_skills/evaluation/metrics/base.py
```diff
@@ -29,6 +30,92 @@ def get_metrics(self):
...
+            prediction_results (Optional[list]): List of prediction results. If provided, it will ignore the `predictions` argument.
+        """
+        if prediction_results is None:
+            assert predictions is not None, "Either predictions or prediction_results must be provided"
+            prediction_results = [self.get_prediction_results(pred) for pred in predictions]
+
+        if pred_keys is None:
+            pred_keys = prediction_results[0].keys()
+
+        for k in range(len(prediction_results), 1, -1):
```
Kipok, line 57: after we change max aggregations logic, can probably have a normal loop here, not reversed to make it simpler

nemo_skills/evaluation/metrics/base.py
```diff
@@ -29,6 +30,92 @@ def get_metrics(self):
...
+                    agg_mode_dict[f"majority@{k}"][pred_field] += False
+                    continue
+
+                # Find the most common answer and its correctness
+                majority_result = Counter(valid_answers_and_results).most_common(1)[0][0]
+
+                # Update the metric
+                agg_mode_dict[f"majority@{k}"][pred_field] += majority_result[1]
+
+    def get_pass_at_k(
```
Kipok, line 77: most comments from majority method apply here as well

nemo_skills/evaluation/metrics/code_metrics.py
```diff
@@ -29,31 +35,24 @@ def update(self, predictions):
...
+        self.get_pass_at_k(self.agg_mode_dict, predictions=predictions)
 
     def get_metrics(self):
-        metrics_dict = {
-            "num_entries": self.total,
-            "passing_base_tests": self.total_correct / self.total * 100.0,
-            "passing_plus_tests": self.total_correct_plus / self.total * 100.0,
-        }
+        metrics_dict = {}
+        for agg_mode, agg_metric_dict in self.agg_mode_dict.items():
```
Kipok, line 42: seems that every class has this loop now, maybe let's move that over to base and then each metric will only need to define a new method that takes agg_metric_dict and returns the metric dictionary that's then assigned to metrics_dict[agg_mode] by the base class? But current way is also ok, I guess, not too much duplication

Also I feel that our variable names here could be improved, so feel free to change self.agg_mode_dict / agg_metric_dict to something more descriptive if you have better ideas

nemo_skills/evaluation/metrics/math_metrics.py
```diff
@@ -309,6 +200,7 @@ def get_metrics(self):
 
             metrics_dict[agg_mode]["no_answer"] = (agg_metric_dict["no_answer"] / self.total) * 100.0
 
+        print(metrics_dict.keys())
```
Kipok, line 203:
```suggestion
```

Kipok (commented):
thanks, added some comments! Also best to use a branch, not a fork as otherwise our tests cannot properly run

nemo_skills/evaluation/metrics/base.py
shan18: I thought this metric might be used by other benchmarks as well so I kept it in the main file. If not, I'll just move it to the math metric file.

shan18: No we don't. I made this a generic function. But now I'll move it inside the math metric file so will remove the unecessary parameters.

Kipok: Actually there is another one where it makes sense, the https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/evaluation/metrics/answer_judgement_metrics.py. I think you missed it in this PR, can you please change that class accordingly? And in that case we can keep it in the base metrics.

There is also arena_metrics and mtbench_metrics that would need changes I guess?

PR #472, opened 2025-05-17
Removing calib params from nemo to hf conversion

shtoshni:


shtoshni:
This is a minor fix for the ckpt conversion script. The calibration arguments are being passed to conversion scripts that don't support them.

Kipok (approved):

PR #473, opened 2025-05-20
Adding megatron inference backend + fix for logging

Kipok:


Kipok:
There is also now a new server_entrypoint parameter that can be used to customize the server script (apparently for megatron-lm people constantly change the server path for different models). @titu1994 FYI

PR #476, opened 2025-05-21
Fixing inheritance + Async support for contamination check

shtoshni:


nemo_skills/inference/check_contamination.py
```diff
@@ -154,6 +183,89 @@ def sync_loop(self, data):
             LOG.info("Contamination portion: %.2f%% (%d/%d)", 100 * num_contaminated / total, num_contaminated, total)
 
 
+    def async_loop(self, data):
```
Kipok, line 186: The current implementation doesn't reuse much from the base class (for both sync and async parts), which isn't the best. Do you think it's possible to change it to reuse most of the logic by moving customizations in the llm_generate method? So there you can just submit N generations instead of 1 and return a list as a value for "generation" field. And then you can restore it back to the current format and assign "contaminated" label inside "postprocess" method? The only thing that we'd need to do is to wrap self.llm.get_generations call into a new method similar to how we do with llm_generate. By default it will just do the current call, but you can customize it to wait for all N generations instead. That way all the logic of the main loop is fully reused and the amount of changes we need here is much smaller

nemo_skills/inference/check_contamination.py
```diff
@@ -54,19 +53,23 @@ class CheckContaminationConfig(GenerateSolutionsConfig):
     top_k: int | None = None
 
     def __post_init__(self):
```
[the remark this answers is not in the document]
shtoshni: we can.. i  was thinking of it, but was worried of introducing too tight of a dependency. What do you think?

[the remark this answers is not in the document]
Kipok: why is that a problem? If you need customization, just override it, but by default it's reused?

nemo_skills/inference/check_contamination.py
shtoshni: The refactoring was a bit non-trivial. But I think we got there in the end. Let me know if the current implementation reduces the redundancy.

nemo_skills/inference/llm_math_judge.py
```diff
@@ -109,10 +115,12 @@ def generate(cfg: LlmMathJudgeConfig):
     task.generate()
 
 
-HELP_MESSAGE = get_help_message(
-    LlmMathJudgeConfig,
-    server_params=server_params(),
-    sandbox_params=sandbox_params(),
+HELP_MESSAGE = (
+    LlmMathJudgeConfig.__doc__ + "\n\n" +
```
Kipok, line 119: Should we do this inside get_help_message, so that it automatically does it for any class?

nemo_skills/inference/generate.py
```diff
@@ -457,6 +464,21 @@ def sync_loop(self, data):
                     self.dump_outputs(outputs, data_points_batch, fout)
                     data_points_batch = []
 
+
+    def get_llm_generations(self, requests_in_progress, generations):
```
Kipok, line 468: I looked at the changes to this file and they seem ok to me. But please run some more tests to double check that normal async generations work properly (e.g. no corruption to the order or anything like that) as this is a very core functionality that we really don't want to break, especially if it's a silent break that doesn't result in an error but causes data corruption

nemo_skills/inference/check_contamination.py
```diff
@@ -100,55 +105,118 @@ def log_example_prompt(self, data):
...
+            # Get the LLM generations for all the queries remaining for this data point
+            gen_dict_list = self.llm.get_generations(gen_id_list)
+            # If all generations have finished, we can process the "generation" key, otherwise keep it as None
+            rem_gen_id_list = []
+            for gen_id, gen_dict in zip(gen_id_list, gen_dict_list):
+                if gen_dict['generation'] is None:
+                    # If any of the generations is not done, we cannot process the data point
+                    rem_gen_id_list.append(gen_id)
+                else:
+                    if "all_generations" not in generations[original_dp_idx]:
```
Kipok, line 192: please add a comment saying that this is ok to add partial generations here since we will add remaining next time this is called. Wasn't obvious to me when I first saw this code

nemo_skills/inference/check_contamination.py
```diff
@@ -100,55 +105,118 @@ def log_example_prompt(self, data):
...
+            # Update the requests in progress to include the remaining generation IDs
+            requests_in_progress[original_dp_idx] = rem_gen_id_list
+
+            if rem_gen_id_list:
+                # There are still generations to be processed
+                generations[original_dp_idx]["generation"] = None
+            else:
+                # All generations have finished
+                # If any of the generations is True, then the data point is considered contaminated
+                contaminated_count = len([generation for generation in generations[original_dp_idx]['all_generations'] if generation.strip() == "True"])
```
Kipok, line 206:
```suggestion
                contaminated = any([generation.strip() == "True" for generation in generations[original_dp_idx]['all_generations']])
```

nemo_skills/inference/check_contamination.py
```diff
@@ -100,55 +105,118 @@ def log_example_prompt(self, data):
...
+                generations[original_dp_idx]['generation'] = (contaminated_count > 0)
```
Kipok, line 207:
```suggestion
                generations[original_dp_idx]['generation'] = contaminated
```

nemo_skills/inference/check_contamination.py
```diff
@@ -29,19 +31,19 @@
 
 @nested_dataclass(kw_only=True)
 class CheckContaminationConfig(GenerateSolutionsConfig):
-    """Top-level parameters for the script"""
+    """LLM-based check contamination parameters.
```
Kipok, line 34: Eventually would be good to add a gpu test for this script (and retrieve_similar). If it's too much for this PR, feel free to just create an issue so that we can track this

shtoshni: Let me work on this in a separate PR. Created an issue here - https://github.com/NVIDIA/NeMo-Skills/issues/481

nemo_skills/inference/llm_math_judge.py
shtoshni: Done!

nemo_skills/inference/check_contamination.py
shtoshni: Cool! Added a few more comments.

nemo_skills/inference/check_contamination.py
```diff
@@ -100,55 +105,118 @@ def log_example_prompt(self, data):
...
+                        
+                proc_outputs.append(proc_output)
+                output_idx += query_per_data_point
+
+            return proc_outputs
+        else:
+            # Club the generation IDs into tuples for each data point
+            generation_ids = []
+            for idx in range(0, len(outputs), query_per_data_point):
+                # We mark each generation ID as "Unfinished" to indicate that it is not done yet
```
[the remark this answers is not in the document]
shtoshni: Oops we are not creating tuples anymore. Replaced it with 

 _\# Create a list of lists, where each inner list contains the generation IDs for a data point_

nemo_skills/inference/generate.py
shtoshni: I ran the model is both sync and async generations, and the generations look comparable to me, even if not exactly the same for all instances. The contamination results are also similar across generations.

nemo_skills/inference/check_contamination.py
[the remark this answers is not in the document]
Kipok: Did you push this?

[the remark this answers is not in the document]
shtoshni: Yes

Kipok (approved):
Thanks!

PR #480, opened 2025-05-22
Fixing judgement + Cluster logic in summarize_results

shtoshni:


nemo_skills/pipeline/summarize_results.py
```diff
@@ -90,22 +90,22 @@ def summarize_results(
...
-            tunnel,
-            get_unmounted_path(cluster_config, results_dir),
-            temp_dir,
-            remote_tar_dir=get_unmounted_path(cluster_config, remote_tar_dir),
-            verbose=verbose,
-        )
-        results_dir = Path(temp_dir) / Path(results_dir).name
+        if cluster_config.get("executor", "") == "local":
+            results_dir = get_unmounted_path(cluster_config, results_dir)
+        else:
```
Kipok, line 95: this should still be elif, because we want to support running this without config and then it just accesses the files on the local filesystem without unmounting

tests/gpu-tests/test_judge.py
```diff
@@ -66,3 +66,12 @@ def test_trtllm_judge():
     for line in lines:
         data = json.loads(line)
         assert 'judgement' in data
+
+    # Adding a summarization step to check that the results are formatted correctly
```
Kipok, line 70: did this actually fail without your fix? Just wondering if it's sufficient enough to catch an error or we need some other test cases for this

nemo_skills/pipeline/summarize_results.py
shtoshni: But "cluster is not None:" is already tested in the "if" above this, right?

Kipok: oh, right, sorry

Kipok (approved):
thanks!

PR #484, opened 2025-05-23
Raise error if the retrieve_similar contains any None values.

smahdavi4:


nemo_skills/inference/retrieve_similar.py
```diff
@@ -113,6 +113,8 @@ def retrieve_similar(cfg: RetrieveSimilarConfig):
     retrieve_from_list = read_data(cfg.retrieve_from, cfg.retrieve_key)
     compare_to_list = read_data(cfg.compare_to, cfg.retrieve_key)
 
+    assert all(retrieve_from_list), "retrieve_from_list contains none values"
```
Kipok, line 116: I'd do the same for compare_to as well. I think it's a lot more common that if there are any none values there that this is a mistake, rather than an expected behavior

Kipok (approved):
thanks!

PR #486, opened 2025-05-28
Genselect updates

shtoshni:
The `output_instance` in genselect copies the original instance before updating the evaluation keys based on the selected solution. We currently just update the `is_correct` based on the selected solution. This leads to `summarize_results` reporting a non-sensical `judge_correct` score. In this PR, we keep both the symbolic and judgement key around from the original solutions. The selected solution's judgement key is copied into the `output_instance`.


shtoshni:
It's a minor PR dealing mostly with genselect. Merging it for now.

PR #487, opened 2025-05-28
Minor fixes and test for CodeGen run_cmd refactor

titu1994:


tests/scripts/run_cmd_llm_infer_check.py
```diff
@@ -0,0 +1,17 @@
+from openai import OpenAI
```
Kipok, line 1: license header

tests/test_run_cmd_llm_infer.py
```diff
@@ -0,0 +1,65 @@
...
+import json
+import os
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+import pytest
+
+sys.path.append(str(Path(__file__).absolute().parents[1]))
```
Kipok, line 24:
```suggestion
```

tests/test_run_cmd_llm_infer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    outputs_map = {}
+    for server_type, model_path in model_info:
+        if not model_path:
+            continue
+
+        output_dir = f"/tmp/nemo-skills-tests/qwen2.5/{server_type}-run-cmd"
+        command = (
+            f"cd /nemo_run/code/tests/scripts/ && "
+            f"rm -f {os.path.join(output_dir, 'output.txt')} && "
```
Kipok, line 44: it's better to remove this file outside of the command. Otherwise if the run_cmd fails completely the assert can be passed if the file is present in the test folder. You can use the newly added docker_rm utility for this to avoid permission errors

nemo_skills/pipeline/eval.py
```diff
@@ -226,6 +226,7 @@ def eval(
         cluster_config,
         log_dir=log_dir,
         mount_map={output_dir: None},
+        create_remote_dir=True,
```
Kipok, line 229: this is actually unnecessary, by the way. The inference/generate.py will always create this folder for you, so just an extra ssh access here that we can avoid. Not a big deal, so feel free to keep if you'd like

titu1994: I can drop it here.

docs/pipelines/run-cmd.md
```diff
@@ -67,4 +67,4 @@ ns run_cmd \
     cd /nemo_run/code/ && python inference.py
 ```
 
-This will launch the LLM inference server, the
\ No newline at end of file
+This will launch the LLM inference server, the sandbox server and then run the inference code in the sandbox.
```
Kipok, line 70:
```suggestion
This will launch the LLM inference server, the sandbox server and then run the inference code.
```

tests/gpu-tests/test_run_cmd_llm_infer.py
```diff
@@ -0,0 +1,72 @@
...
+from pathlib import Path
+
+import pytest
+
+from tests.conftest import docker_rm
+
+
+def test_run_cmd_llm_infer():
+    """
+    Starts (if available) TRTLLM, Nemo, and VLLM servers, then sends the same prompt
```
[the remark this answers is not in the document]
titu1994: Fixed

Kipok (approved):
Thanks!

PR #493, opened 2025-05-29
Add a common implementation of pass@k and majority@k in the `BaseMetrics` class

shan18:
### Summary

* New `get_prediction_results` abstract method in `BaseMetrics`
* Two new functions: `get_majority_at_k` and `get_pass_at_k` in `BaseMetrics` that compute the pass@k and majority@k metrics.
  - These functions get results from the `get_prediction_results` to compute the scores.
* Updated the evaluation code of all the metrics to use these classes for computing metrics.
* Moved some common functionality from the `reset` and the `update` function back to the base class.
* Modified the `max_aggregations_to_print` function to `aggregations_to_print` to directly specify the order in which each metric should be printed for each benchmark.


Kipok:
merged in here https://github.com/NVIDIA/NeMo-Skills/pull/504

PR #501, opened 2025-06-03
Update nemo rl grpo templating

gwarmstrong:
Updates nemo rl grpo so that nemo_skills templating logic can be used with ns-style input,

so an input file would look something like:
```
{"expected_answer": "-17x + 21", "problem": "Simplify the expression: $2(x-3)^2 - 5(x+2)(x-1) + 3x^2 - 7$."}
{"expected_answer": "\\frac{3}{5}", "problem": "Evaluate $\\sin \\left( \\arctan \\frac{3}{4} \\right)$"}
```

and you can use the arguments to grpo like:
```
++data.prompt.prompt_config='qwen/math-cot' ++data.prompt.prompt_template='qwen-instruct'
```
to update the prompt config and template used


nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -110,9 +110,14 @@ policy:
 
 data:
   max_input_seq_length: ${policy.max_total_sequence_length} # upper bound, real truncation occurs at vllm.max_model_len
-  prompt_file: "/nemo_run/code/nemo_skills/training/nemo_rl/prompts/math.txt"
-  system_prompt_file: null
-  dataset_name: "OpenMathInstruct-2"
+  prompt:
+    prompt_config: 'qwen/math-cot'
```
Kipok: let's not set defaults and instead ask users to specify always. Otherwise too easy to make a mistake and not notice

```suggestion
    prompt_config: ???
```

nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -110,9 +110,14 @@ policy:
...
+    prompt_template: 'qwen-instruct'
```
Kipok:
```suggestion
    prompt_template: ???
```

tests/test_prompts.py
```diff
@@ -100,6 +100,59 @@ def test_generic_math_problem_augmentation_prompt():
...
 
 
+def test_qwen_math_prompt_dict():
+    prompt = get_prompt('qwen/math-cot', 'qwen-instruct')
+
+    filled = prompt.fill({'problem': "What's the meaning of life?"}, return_templated_dict=True)
+    expected_result = [
+        {
+            'role': 'system',
+            'content': '<|im_start|>system\nPlease reason step by step, and put your final answer within \\boxed{}.<|im_end|>\n',
```
Kipok, line 110: are you sure that's how this is supposed to be in nemo-rl? They want the content to be formatted, but still be presented as messages list?

nemo_skills/training/nemo_rl/start_grpo.py
```diff
@@ -70,54 +70,35 @@ def parse_args() -> tuple[argparse.Namespace, list[str]]:
...
-            {
-                "role": "user",
-                "content": data["problem"],
-            },
-            {
-                "role": "assistant",
-                "content": data[output_key],
-            },
-        ],
+        "problem": data["problem"],
```
Kipok: can we make this independent of the particular key (problem here). So we would just pass in all of the data as input to prompt.fill and whatever it returns we use as an input. And for the output, we'd ask user to provide a key (can be expected_answer) as default. That way it would work with any prompt we have and not assume a particular structure, right?

nemo_skills/training/nemo_rl/start_grpo.py
```diff
@@ -126,158 +107,84 @@ def prepare_openinstructmath2_dataset(
...
-    user_message["content"] = message[0]
-    message_log.append(user_message)
+@dataclass
+class NSTaskDataSpec(TaskDataSpec):
+    prompt_spec: dict[str, Any] | None = None
 
-    length = sum(len(m["token_ids"]) for m in message_log)
+    def copy_defaults(self, from_spec):
+        super().copy_defaults(from_spec)
+        default_attrs = {
```
Kipok: is this required to be duplicated here? Even if we already specify the same thing in the yaml config?

nemo_skills/prompt/utils.py
```diff
@@ -217,6 +218,7 @@ def fill(
         prefix_generation_to_response: bool = False,
         continue_prefix_generation: bool = False,
         multi_turn_key: str | None = None,
+        return_templated_dict: bool = False,
```
Kipok, line 221: please add some documentation for this new parameter

nemo_skills/training/nemo_rl/start_grpo.py
wedu-nvidia: I checked the original code and they also use the same copy_defaults function which will be called by dataset construction function later.

tests/test_prompts.py
wedu-nvidia: I think nemo-rl wanna wrap the data to following type
LLMMessageLogType = list[dict[str, Union[str, torch.Tensor]]]
So I think it should be correct

gwarmstrong: so, ulimately, the "role" field here is used to determine the loss mask, here: https://github.com/NVIDIA/NeMo-RL/blob/383ed0b049c44ffe067cbc3392f4f8cbca15515c/nemo_rl/algorithms/grpo.py#L440-L443

then after that step, it is flattened: https://github.com/NVIDIA/NeMo-RL/blob/383ed0b049c44ffe067cbc3392f4f8cbca15515c/nemo_rl/algorithms/grpo.py#L457-L463

so  yes, it has to be formatted here so the loss mask can be determined correctly by the grpo algorithm, though  the content will be unused after we tokenize here: https://github.com/NVIDIA/NeMo-Skills/pull/501/files#diff-233f6e659cee7bb2fa067c4d520d6fb6e582675caa6b3f095e361044a0f4aae5R179

nemo_skills/prompt/utils.py
gwarmstrong: done

nemo_skills/training/nemo_rl/configs/grpo.yaml
gwarmstrong: done

nemo_skills/training/nemo_rl/configs/grpo.yaml
gwarmstrong: done

nemo_skills/training/nemo_rl/start_grpo.py
gwarmstrong: yeah I think the reason for this is because nemo-rl's config parsing logic is a little bit custom and doesn't always fully resolve the `???` paramters, but that being said, I agree that at least this shouldn't be set by default and we should rely on user. I removed this function and it is fine as long as you don't mind an error like this:

```
7-nemo-run/0 ============================================================
7-nemo-run/0                   SETUP COMPLETE
7-nemo-run/0 ============================================================
7-nemo-run/0 
7-nemo-run/0 Traceback (most recent call last):
7-nemo-run/0   File "/nemo_run/code/nemo_skills/training/nemo_rl/start_grpo.py", line 327, in <module>
7-nemo-run/0     main()
7-nemo-run/0   File "/nemo_run/code/nemo_skills/training/nemo_rl/start_grpo.py", line 310, in main
7-nemo-run/0     grpo_train(
7-nemo-run/0   File "/opt/NeMo-RL/nemo_rl/algorithms/grpo.py", line 384, in grpo_train
7-nemo-run/0     for batch in dataloader:
7-nemo-run/0                  ^^^^^^^^^^
7-nemo-run/0   File "/opt/nemo_rl_venv/lib/python3.12/site-packages/torchdata/stateful_dataloader/stateful_dataloader.py", line 450, in __next__
7-nemo-run/0     return super().__next__()
7-nemo-run/0            ^^^^^^^^^^^^^^^^^^
7-nemo-run/0   File "/opt/nemo_rl_venv/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 708, in __next__
7-nemo-run/0     data = self._next_data()
7-nemo-run/0            ^^^^^^^^^^^^^^^^^
7-nemo-run/0   File "/opt/nemo_rl_venv/lib/python3.12/site-packages/torchdata/stateful_dataloader/stateful_dataloader.py", line 491, in _next_data
7-nemo-run/0     data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
7-nemo-run/0            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7-nemo-run/0   File "/opt/nemo_rl_venv/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
7-nemo-run/0     data = [self.dataset[idx] for idx in possibly_batched_index]
7-nemo-run/0             ~~~~~~~~~~~~^^^^^
7-nemo-run/0   File "/opt/NeMo-RL/nemo_rl/data/datasets.py", line 111, in __getitem__
7-nemo-run/0     datum_spec = task_data_processor(
7-nemo-run/0                  ^^^^^^^^^^^^^^^^^^^^
7-nemo-run/0   File "/nemo_run/code/nemo_skills/training/nemo_rl/start_grpo.py", line 152, in ns_data_processor
7-nemo-run/0     prompt = get_prompt(
7-nemo-run/0              ^^^^^^^^^^^
7-nemo-run/0   File "/nemo_run/code/nemo_skills/prompt/utils.py", line 380, in get_prompt
7-nemo-run/0     config = load_config(prompt_config, config_dir)
7-nemo-run/0              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7-nemo-run/0   File "/nemo_run/code/nemo_skills/prompt/utils.py", line 366, in load_config
7-nemo-run/0     with open(config_path, "rt", encoding="utf-8") as fin:
7-nemo-run/0          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7-nemo-run/0 FileNotFoundError: [Errno 2] No such file or directory: '/nemo_run/code/nemo_skills/prompt/config/???.yaml'
```

If we want to catch it at the point of calling hydra, we would need to either add our own parsing logic to catch it, or update nemo-rl.

nemo_skills/training/nemo_rl/start_grpo.py
gwarmstrong: that's a great point, I've updated a bunch of this logic now so everything passes through to `prompt.fill`, which has the added benefit that now the problem-specific logic is a lot easier to identify/encapsulate.

nemo_skills/training/nemo_rl/start_grpo.py
Kipok: I think that's ok for now, we have the same problem for other scripts as it's hard to validate all arguments before submitting a job since they often need to be run in the container to even get basic checks. We can refine this later if we find it to be a common problem

nemo_skills/prompt/utils.py
```diff
@@ -230,6 +232,8 @@ def fill(
             multi_turn_key: If specified, will read the list from input_dict[multi_turn_key]
                 and use it to construct the prompt. You input_dict should also have "assistant" key in all
                 turns except last containing assistant reply.
+            return_templated_dict: Indicates whether to return a messages list where hte template is used
```
Kipok:
```suggestion
            return_templated_dict: Indicates whether to return a messages list where the template is used
```

nemo_skills/prompt/utils.py
```diff
@@ -230,6 +232,8 @@ def fill(
...
+                to fill the prompt. If so, a list of dicts with 'role' and 'content'  keys will be returned.
```
Kipok:
```suggestion
                to fill the prompt. If so, a list of dicts with 'role' and 'content' keys will be returned. 
                In this case the final user and assistant messages will include special tokens.
```

Kipok (approved):
Thanks, looks great! We'd need to find some way to customize the environment and task (task is easy, but the environment probably will need some work to be able to pass through cmdline), but let's leave it for another PR when we have another use-case ready to try

PR #503, opened 2025-06-04
update to latest image

wedu-nvidia:
I updated the latest verl image and also install from latest verl

### DIFF
--- a/dockerfiles/Dockerfile.verl
+++ b/dockerfiles/Dockerfile.verl
@@ -1,19 +1,15 @@
-FROM verlai/verl:ngc-th2.4.0-cu124-vllm0.6.3-te1.7-v0.0.4
-
+FROM whatcanyousee/verl:ngc-cu124-vllm0.8.5-sglang0.4.6-mcore0.12.0-te2.3
 # Set working directory
 WORKDIR /opt
 
 # Install verl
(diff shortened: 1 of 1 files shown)


Kipok (commented):
Thanks! Can you please also update the warning message over here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/pipelines/training-verl-openrlhf.md?plain=1#L21 to only mention openrlhf?

dockerfiles/Dockerfile.verl
```diff
@@ -1,19 +1,15 @@
-FROM verlai/verl:ngc-th2.4.0-cu124-vllm0.6.3-te1.7-v0.0.4
-
+FROM whatcanyousee/verl:ngc-cu124-vllm0.8.5-sglang0.4.6-mcore0.12.0-te2.3
 # Set working directory
 WORKDIR /opt
 
 # Install verl
-ENV VERL_COMMIT=db57bb4331f86769719d3ab2a2e7c1d90d26dc1e
-RUN git clone https://github.com/titu1994/verl.git && \
+ENV VERL_COMMIT=main
```
Kipok: let's not use main and instead fix to whatever is the current latest commit. Otherwise it's hard to keep track of what was the version that worked when we last built this

wedu-nvidia: ok, revised as above

PR #504, opened 2025-06-04
Refactoring of the metrics logic

Kipok:


Kipok:
based on https://github.com/NVIDIA/NeMo-Skills/pull/493 to provide a bit more flexibility in updating metrics for majority / pass and reusing the base implementation

shtoshni:
Ok I see new updates.. Let me review once it passes all the tests.

Kipok:
it works fine, just the problem with the test - some operations like majority are non-deterministic, so not getting same output as locally. But should be good enough for review

nemo_skills/inference/llm_math_judge.py
```diff
@@ -101,6 +101,14 @@ def prefill_generation(self, data_point):
         else:
             return {"generation": judgement}
 
+    def dump_outputs(self, outputs, data_points, fout):
+        # removing num_generated_tokens to keep the original ones instead of the judge as it's often not relevant
```
shtoshni, line 105: Nice!

nemo_skills/evaluation/metrics/if_metrics.py
```diff
@@ -58,51 +34,35 @@ def update(self, predictions):
...
-            self._update_single_stat(self.strict_stats, [pred['strict_eval'] for pred in predictions])
-            self._update_single_stat(self.loose_stats, [pred['loose_eval'] for pred in predictions])
-        else:
-            # Single prediction
-            self.agg_mode = "greedy"
-
-            self._update_single_stat(self.strict_stats, [predictions[0]['strict_eval']])
-            self._update_single_stat(self.loose_stats, [predictions[0]['loose_eval']])
+        self._compute_pass_at_k(predictions=strict_predictions, eval_dict=self.strict_eval_dict)
+        self._compute_pass_at_k(predictions=loose_predictions, eval_dict=self.loose_eval_dict)
```
shan18, line 43: According to the original code, IFEval had a different pass at k implementation. For example, each prediction had 3 outputs:
```
pred-1: [True, False, False]
pred-2: [False, False, True]
...
```

And how pass@k was computed:
* Each output index was compared across all the predictions i.e.
  ```
  final_pred[0] = pred[0][0] or pred[1][0] or pred[2][0] or ...
  final_pred[1] = pred[0][1] or pred[1][1] or pred[2][1] or ...
  final_pred[2] = pred[0][2] or pred[1][2] or pred[2][2] or ...
  ```
* And then the code checked if all 3 outputs are set to `True`.

This is why IFEval had a custom pass@k function to override the base metrics class.

So maybe we want to add that via the `_update_score_metrics_for_pass` function

shan18: For reference, this was the old custom implementation:
```python
def _get_best_prediction(self, stats_dict, elems):
    """Will update using the pass@k strategy (just pass a single-element list to get greedy)."""
    # computing "pass@k" score
    follow_instruction_list = elems[0]['follow_instruction_list']
    for elem in elems:
        follow_instruction_list = [
            follow_instruction_list[i] or elem['follow_instruction_list'][i]
            for i in range(len(follow_instruction_list))
        ]

    if all(follow_instruction_list):
        stats_dict['prompt'] += 1

    stats_dict['instruction'] += sum(follow_instruction_list)
  ```

Comment (Kipok) [reply nemo_skills/evaluation/metrics/if_metrics.py:43]: you're right, it's not the same as before. There is no "standard" implementation for pass@k for ifeval and after some thinking I decided to go with this simpler one. I now actually think that this one is more correct, because it doesn't mix predictions from different generations together and instead returns the score for the best individual prediction

File: nemo_skills/evaluation/metrics/mtbench_metrics.py [hunk 3]
```diff
@@ -24,9 +24,6 @@
 
 
 class MtBenchMetrics(BaseMetrics):
```
shan18, line 26: There is no `_get_score_dict` function implemented so this would throw the `NotImplementedError` from the `BaseMetrics` class.

Kipok: It only throws that error if that function is actually used. I decided not to define it as abstract method to allow classes not to use this interface if they don't want to. So in this case it should work fine as that function is never called

nemo_skills/evaluation/metrics/arena_metrics.py
```diff
@@ -120,12 +116,12 @@ def get_metrics(self):
...
         return {self.agg_mode: metrics}
 
     def reset(self):
+        super().reset()
         self.scores = []  # list of lists
         self.lengths = 0
-        self.total = 0
-        # Set automatically
-        self.agg_mode = "greedy"
+        self.agg_mode = "pass@1"
```
shan18, line 127: * ~~This function might also throw the `NotImplementedError` since the `_get_score_dict` function is not there.~~
* Also, it does not use the new base metrics class to compute the pass@1

nemo_skills/evaluation/metrics/mtbench_metrics.py
```diff
@@ -128,10 +115,10 @@ def get_metrics(self):
...
         self.scores = defaultdict(list)
-        self.total = 0
-        self.agg_mode = "greedy"
+        self.agg_mode = "pass@1"
 
-    def max_metrics_to_print(self):
+    def metrics_to_print(self):
         """We are only printing the averages, but all other metrics can still be found in metrics.json"""
-        return 4
+        return ['num_entries', 'avg_tokens', 'average', 'average_turn1', 'average_turn2']
```
shan18: The math metric class return from this function a dict:
https://github.com/NVIDIA/NeMo-Skills/blob/d61d6f03929f907951f1cb61cf23c68d0bc0f25c/nemo_skills/evaluation/metrics/math_metrics.py#L147-L154

Confirming if this is correct.

Kipok: good catch, thanks!

nemo_skills/evaluation/metrics/arena_metrics.py
Kipok: that's right, the new interface doesn't work well with arena metrics even for pass@1 and since it's only 5 lines of code I thought it would be simpler to just keep it as the original logic

shan18 (approved):

nemo_skills/evaluation/metrics/base.py
```diff
@@ -13,29 +13,281 @@
...
+
+                # If no valid answers, it's incorrect
+                if not valid_answers_and_results:
+                    majority_score = 0
+                    majority_answer = None
+                else:
+                    # sorting to ensure reproducible scores in case of ties in majority
+                    valid_answers_and_results = sorted(valid_answers_and_results)
+                    # Find the most common answer and its correctness
+                    majority_answer, majority_score = Counter(valid_answers_and_results).most_common(1)[0][0]
```
shtoshni, line 153: Need a tiebreak logic here. Maybe allow for tiebreaks if a tiebreak method is implemented.

Kipok: right, but probably can be added in another PR? We can just add a new method that will do that and any class that needs to customize, will be able to override

nemo_skills/evaluation/metrics/if_metrics.py
```diff
@@ -21,35 +21,11 @@ class IFMetrics(BaseMetrics):
...
-
-        stats_dict['instruction']['total'] += len(instruction_id_list)
-        stats_dict['instruction']['correct'] += sum(follow_instruction_list)
-
-        for instruction_id, followed_or_not in zip(instruction_id_list, follow_instruction_list):
-            instruction_id = instruction_id.split(":")[0]
-            stats_dict['tier0']['total'][instruction_id] += 1
-            if followed_or_not:
-                stats_dict['tier0']['correct'][instruction_id] += 1
+    def _get_score_dict(self, prediction: dict) -> dict[bool | int | float]:
```
shtoshni:
```suggestion
    def _get_score_dict(self, prediction: dict) -> dict[str, bool | int | float]:
```

Kipok: While str will technically work, I'm not sure it's a good idea to use it. We take max over scores to get pass@k and max over strings is probably not going to work as expected

nemo_skills/evaluation/metrics/base.py
```diff
@@ -13,29 +13,281 @@
...
+            for metric_key, metric_value in agg_metric_dict.items():
+                if isinstance(metric_value, float):
+                    # by default we will return all float metrics as percentages
+                    metrics_dict[agg_mode][metric_key] = 100.0 * metric_value / self.total
+                else:
+                    metrics_dict[agg_mode][metric_key] = metric_value
+
+        return metrics_dict
+
+    def _get_score_dict(self, prediction: dict) -> dict[bool | int | float]:
```
shtoshni:
```suggestion
    def _get_score_dict(self, prediction: dict) -> dict[str, bool | int | float]:
```

nemo_skills/evaluation/metrics/base.py
```diff
@@ -13,29 +13,281 @@
...
+                    for pred_answer, correctness_dict in zip(predicted_answers[:k], score_dicts[:k])
+                    if pred_answer is not None
+                ]
+
+                # If no valid answers, it's incorrect
+                if not valid_answers_and_results:
+                    majority_score = 0
+                    majority_answer = None
+                else:
+                    # sorting to ensure reproducible scores in case of ties in majority
```
shtoshni, line 150: Are we sure that this comment makes sense i.e. post sorting, Counter is deterministic as well?

nemo_skills/evaluation/metrics/if_metrics.py
shtoshni: str is the key is what this suggestion is trying to indicate.

Kipok: oh, right, yeah, let me change that

nemo_skills/evaluation/metrics/base.py
Kipok: from the docs "Elements with equal counts are ordered in the order first encountered"

shtoshni:
Thanks, great PR!
Feel free to merge after addressing the minor nits.

PR #506, opened 2025-06-05
Add support for mounted cluster datasets + refactoring of eval parameters

Kipok:
1. Remove dataset / split from inference/generate.py (now input_file is required)
2. Rename DEFAULT_GENERATION_ARGS and DEFAULT_EVAL_ARGS to remove DEFAULT_ for consistency with other arguments (they are all defaults that can be overridden)
3. Add a new optional EVAL_SPLIT parameter that can be used to specify default split if it's not "test"
4. Remove validation / train_full from gsm8k (it's not been used for a while and we already don't have it for math)
5. Add support for using datasets on cluster directly


nemo_skills/inference/generate.py
```diff
@@ -259,10 +245,6 @@ def setup_llm(self):
         return llm
 
     def setup_prompt(self):
-        if self.cfg.prompt_config is None:
```
titu1994, line 262: So no more default prompt config ?

nemo_skills/pipeline/eval.py
```diff
@@ -179,7 +242,11 @@ def eval(
         help="Path to a custom dataset folder that will be searched in addition to the main one. "
         "Can also specify through NEMO_SKILLS_EXTRA_DATASETS.",
     ),
-    # extra_datasets_type: ExtraDatasetType = typer.Option(ExtraDatasetType.copy, help="How to handle extra datasets"),
+    extra_datasets_type: ExtraDatasetType = typer.Option(
+        os.getenv("NEMO_SKILLS_EXTRA_DATASETS_TYPE", "local"),
+        help="If you have extra datasets locally, set to 'local', if on cluster, set to 'cluster'."
+        "Can also specify through NEMO_SKILLS_EXTRA_DATASETS_TYPE environment variable.",
+    ),
```
titu1994, line 250: Typer has `envvar` arg in Option that reads env variables if you want that

nemo_skills/pipeline/eval.py
```diff
@@ -202,10 +269,20 @@ def eval(
...
         pass
+    try:
+        extra_datasets_type = extra_datasets_type.value
+    except AttributeError:
+        pass
 
     cluster_config = get_cluster_config(cluster, config_dir)
     cluster_config = resolve_mount_paths(cluster_config, mount_paths)
 
+    if extra_datasets_type == ExtraDatasetType.cluster and cluster_config['executor'] != 'slurm':
```
titu1994, line 281: Hmm why is this ? Mounting works in docker based local env

titu1994 (approved):
Overall very nice changes, minor questions only

nemo_skills/pipeline/eval.py
Kipok: I think some of the tunnel logic might not work with this, so easier to just disallow for now. We can fix this later if we find it convenient to use mounted locations locally

nemo_skills/inference/generate.py
Kipok: it's still set by ns eval, so no changes there. The only change is really when you used ns generate with dataset/split parameters but those are removed now. So shouldn't affect any of the use-cases

nemo_skills/pipeline/eval.py
Kipok: nice, let me use that

nemo_skills/pipeline/eval.py
titu1994: Sounds good

nemo_skills/inference/generate.py
titu1994: Ah ok that's a good change. Earlier having eval args in generate was confusing

PR #507, opened 2025-06-06
Improve Answer-Judge metrics

smahdavi4:
Fix return_none in answer judge + cover more valid judgements from judge + add f1/precision/recall.

### DIFF
--- a/nemo_skills/evaluation/metrics/answer_judgement_metrics.py
+++ b/nemo_skills/evaluation/metrics/answer_judgement_metrics.py
@@ -14,18 +14,49 @@
 
 from nemo_skills.evaluation.metrics.base import BaseMetrics
 from nemo_skills.evaluation.metrics.utils import is_correct_judgement
+from functools import partial
+from collections import defaultdict
 
+is_correct_judgement_or_none = partial(is_correct_judgement, return_none=True)
 
 class AnswerJudgementMetrics(BaseMetrics):
+    def __init__(self):
+        super().__init__()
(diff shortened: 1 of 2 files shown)


nemo_skills/evaluation/metrics/answer_judgement_metrics.py
```diff
@@ -57,16 +89,22 @@ def _update_score_metrics_for_pass(
     ):
         assert score_method == 'correct_judgements'
         # expected answer is always the same for all predictions, so just take the first one
-        gt_judgement = is_correct_judgement(predictions[0]['expected_judgement'])
-        pred_judgement = is_correct_judgement(predictions[0]['judgement'])
-        # if pass is not correct, means all predictions are the same and wrong
-        if not pass_score:
-            self._update_fp_fn(eval_dict[f"pass@{k}"], pred_judgement, gt_judgement)
+        gt_judgement = is_correct_judgement_or_none(predictions[0]['expected_judgement'])
+        pred_judgements = [is_correct_judgement_or_none(pred['judgement']) for pred in predictions]
```
Kipok:
```suggestion
        pred_judgements = [is_correct_judgement_or_none(pred['judgement']) for pred in predictions[:k]]
```
Right now we pass full list of predictions (in case you want to do some bootstrapping or things like this to calculate @ k metrics), so need to pick the right subset

smahdavi4:
Done. The metrics files had a lot of floats which did not match exactly on my local machine with the CI/CD, so had to change the tests to account for metric differences and add an epsilon tollerance.

PR #509, opened 2025-06-06
Adding ns prepare_data that can prepare and keep datasets on the cluster

Kipok:
With this you can ask to prepare the data on the cluster and then both evals and metrics can access split.jsonl and init.py there directly, so we don't need to keep uploading large datasets


docs/basics/index.md
```diff
@@ -238,7 +238,7 @@ model for a total of 64 samples)
 First prepare evaluation data
 
 ```bash
-python -m nemo_skills.dataset.prepare aime24 aime25
+ns prepare_data aime24 aime25
```
titu1994, line 241: Hmm does the script work for local data preparation too? If not, lets show both examples for local prep and cluster prep

nemo_skills/dataset/utils.py
```diff
@@ -64,32 +67,83 @@ def import_from_path(file_path, module_name=None):
...
-def get_dataset_module(dataset, extra_datasets=None, dataset_init_path=None):
+class ExtraDatasetType(str, Enum):
+    local = "local"
+    cluster = "cluster"
+
+
+def _get_dataset_module_from_cluster(cluster_config, mounted_path):
+    # getting tmp path to download init.py
+    with tempfile.TemporaryDirectory() as tmpdir:
+        tmp_path = str(Path(tmpdir) / f"init.py")
```
titu1994, line 78: Hmm shouldnt it be `__init__.py`?

nemo_skills/pipeline/prepare_data.py
```diff
@@ -0,0 +1,127 @@
...
+    exclusive: bool = typer.Option(True, "--not_exclusive", help="Disable --exclusive flag"),
+    check_mounted_paths: bool = typer.Option(False, help="Check mounted paths availability"),
+):
+    """Prepare datasets by running python -m nemo_skills.dataset.prepare.
+
+    Run `python -m nemo_skills.dataset.prepare --help` to see other supported arguments.
+    """
+    setup_logging(disable_hydra_logs=False, use_rich=True)
+    extra_arguments = f'{" ".join(ctx.args)}'
+    command = f"python -m nemo_skills.dataset.prepare {extra_arguments}"
```
titu1994, line 64: Hmm the dataset preparation command should be taken as input in typer (with this as default). That way a user can point to their own script to process the dataset.. Can be done later though.

titu1994: Also if its generating the dataset in local /nemo_run/code/nemo_skills/dataset/ dir and then copying onto cluster dataset dir .. then you need to delete the copy in /nemo_run/code/ns/dataset no? Or two copies will remain, one in data dir and one inside the nemo-run folder.

nemo_skills/pipeline/prepare_data.py
```diff
@@ -0,0 +1,127 @@
...
+                f"Make sure to set NEMO_SKILLS_DATA_DIR={data_dir} in your environment "
+                f"or pass it as an argument to `ns eval`."
+            )
+        if data_dir_env_var is None:
+            LOG.warning(
+                f"NEMO_SKILLS_DATA_DIR environment variable is not set. "
+                f"You might want to set it as NEMO_SKILLS_DATA_DIR={data_dir} "
+                f"to avoid always specifying it as a parameter to `ns eval`."
+            )
+        # TODO: automatically add it to cluster config based on user prompt?
```
titu1994, line 102: That would be nice. You can use is_mounted() and get_mounted() to test if its already in mount paths before adding it.

titu1994 (approved):
Overall looks great! Minor comments

docs/basics/index.md
titu1994: Scratch that, it works for local too.

Kipok: yes, by default it has the same behavior is `python -m nemo_skills.dataset.prepare` and runs things locally. To do it on cluster you need to provide --cluster parameter. I plan to add documentation for that using ruler as an example, so will do it after we merge that

nemo_skills/dataset/utils.py
Kipok: this doesn't matter, it's just a local file that we are loading, can be named anything. The mounted_path should be `__init__.py`

nemo_skills/pipeline/prepare_data.py
Kipok: > Hmm the dataset preparation command should be taken as input in typer (with this as default).

Let's do that in another PR

Kipok: > then you need to delete the copy in /nemo_run/code/ns/dataset no

There will be two copies, that's true. I thought about using mv instead of cp, but a little hesitant to do that as the dataset folder in code might contain files that we didn't create and so I don't want to be removing them. I agree that probably 99% of the cases that's safe to do, but I imagine there might be some edge-cases where it's not what the user expects (e.g. if reusing code across multiple experiments). And in any case, people need to be cleaning up the nemo-run folder from time to time. 

I'd suggest we go with copy behavior for now and if this becomes an issue we can address this later

PR #511, opened 2025-06-09
Azure OpenAI API and code execution

i-vainn:
This PR:
- Adds Azure OpenAI API model class
- Adds separate `code_tags` configs, removing them from template
- Adds code execution capability to the OpanAI models

### DIFF
--- a/docs/basics/inference.md
+++ b/docs/basics/inference.md
@@ -129,7 +129,7 @@ Click on :material-plus-circle: symbols in the snippet below to learn more detai
 
     sandbox = get_sandbox()  # localhost by default
     llm = get_code_execution_model(server_type="vllm", sandbox=sandbox)
-    prompt = get_prompt('generic/default', 'llama3-instruct', 'llama3') # (1)!
+    prompt = get_prompt('generic/default', 'llama3-instruct', code_tags='llama3') # (1)!
     prompt.config.system = ( # (2)!
         "Environment: ipython\n\n"
         "Use Python to solve this math problem."
--- a/docs/basics/prompt-format.md
+++ b/docs/basics/prompt-format.md
@@ -134,7 +134,7 @@ If you're implementing a new script, you can use the following code to create a 
 from nemo_skills.prompt.utils import get_prompt
 
 # The third parameter is optional and only needed for code execution
-prompt = get_prompt('generic/math', 'llama3-instruct', 'llama3')
+prompt = get_prompt('generic/math', 'llama3-instruct', code_tags='llama3')
 print(prompt.fill({'problem': "What's 2 + 2?"}))
 ```
 


nemo_skills/inference/server/code_execution_model.py
```diff
@@ -114,10 +115,17 @@ def _generate_single(
...
+            new_prompt = copy.deepcopy(prompt)
+        else:
+            new_prompt = copy.deepcopy(prompt)
 
         start_time = int(time.time())
-
+        
+        # For OpenAI models, don't add code_end to stop phrases because OpenAI API
+        # automatically removes stop phrases from the response, which breaks code execution detection
+        final_stop_phrases = (stop_phrases if stop_phrases else []) + [code_end]
```
Kipok: isn't that exactly the same as before?

nemo_skills/inference/server/code_execution_model.py
```diff
@@ -176,7 +184,19 @@ def _generate_single(
...
+
+            # openai don't show what stop word was triggered, so we assume that it was `code_end`
+            # if there's an unfinished code block
+            if is_openai_format and output_dict.get('finish_reason') == 'stop':
+                if output.count(code_end) + 1 == output.count(code_begin):
+                    output += code_end
+            # Update the prompt based on format
+            if is_openai_format:
+                request['prompt'].append({'role': 'assistant', 'content': output})
+                request['prompt'].append({'role': 'user', 'content': "continue"})
```
Kipok, line 191: I guess if we want to support vllm/sglang's continue_final_message parameter, then this should happen in the underlying model, not here

Kipok: If that's hard for some reason, can skip for now and do in another pr

nemo_skills/inference/server/model.py
```diff
@@ -793,14 +796,61 @@ def _stream_chunks(self, response):
...
+        base_url=None,
+        api_key=None,
+        api_version: str = "2024-02-15-preview",
+        max_retries: int = 3,
+        initial_retry_delay: float = 2.0,
+        **kwargs,
+    ):
+        # Call BaseModel.__init__ directly to bypass OpenAIModel.__init__
+        BaseModel.__init__(self, host=host, port=port, **kwargs)
+        from openai import AzureOpenAI
```
Kipok: why importing here? openai is always installed

nemo_skills/inference/server/model.py
```diff
@@ -793,14 +796,61 @@ def _stream_chunks(self, response):
...
+
+        if model is None:
```
Kipok, line 830: can skip the if by doing `model = model or os.getenv("NEMO_SKILLS_OPENAI_MODEL")`

Kipok (commented):
Thanks! Added a few comments. Please also update the docs here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/basics/prompt-format.md?plain=1

nemo_skills/inference/server/code_execution_model.py
i-vainn: Now vllm/sglang don't support chat completions. So I suggest we leave this for the next PR that will also implement chat completions.

docs/basics/inference.md
```diff
@@ -129,7 +129,7 @@ Click on :material-plus-circle: symbols in the snippet below to learn more detai
 
     sandbox = get_sandbox()  # localhost by default
     llm = get_code_execution_model(server_type="vllm", sandbox=sandbox)
-    prompt = get_prompt('generic/default', 'llama3-instruct') # (1)!
+    prompt = get_prompt('generic/default', 'llama3-instruct', 'llama3') # (1)!
```
Kipok:
```suggestion
    prompt = get_prompt('generic/default', 'llama3-instruct', code_tags='llama3') # (1)!
```

docs/basics/prompt-format.md
```diff
@@ -96,22 +94,47 @@ prompt the `gsm8k_standard_few_shot` examples from
...
 
-If you're implementing a new script, you can use the following code to create a prompt and then use it
+If you're implementing a new script, you can use the following code to create a prompt and then use it:
 
 ```python
 from nemo_skills.prompt.utils import get_prompt
 
-prompt = get_prompt('generic/math', 'llama3-instruct')
+# The third parameter is optional and only needed for code execution
+prompt = get_prompt('generic/math', 'llama3-instruct', 'llama3')
```
Kipok:
```suggestion
prompt = get_prompt('generic/math', 'llama3-instruct', code_tags='llama3')
```

Kipok (approved):

PR #513, opened 2025-06-10
Refactoring of the generate pipeline

Kipok:
1. There is now explicit --input_file and --input_dir parameters (only one of the two can be specified). This allows to fully unify llm-judge, reward, contamination and main generate modes. All instances of ++input_file and ++input_dir would need to be renamed to --input_file and --input_dir.
2. ns check_contamination is removed and replaced with ns generate --generation_type=check_contamination
3. ns generate --generation_type=genselect is replaced with a separate ns genselect pipeline
4. ns generate now accepts `--generation_module` parameter which can be used to run with any class that subclasses from GenerationTask. The only requirement is that it needs to define `GENERATION_TASK_CLASS` variable inside the module which we will use to identify the new class. FYI @titu1994 @SeanNaren 
5. Removed `get_generation_module` method of GenerationTask as we can always identify it automatically in code
6. Added `get_server_command_fn` method to GenerationTask to customize which server to use
7. Refactor eval pipeline to reuse most of the logic from generate pipeline. I originally wanted to fully reuse generate inside of eval, but I now think that's a bad idea as it will require too many changes (eval needs different generation parameters for different input files)
8. The default temperature is now set to 0.7 instead of 1.0 inside of generate to get the eval behavior (and it's more used anyway)
9. The above means that all generation scripts (judge, reward, contamination, eval) now automatically properly support .done files and chunking
10. Moved around some generation utilities (to reuse them in other files) and renamed get_generation_command to wait_for_server
11. Removed nemo-aligner based grpo code (it wasn't working anyway)


Kipok:
gpu tests are passing locally, ci is broken

nemo_skills/pipeline/genselect.py
```diff
@@ -0,0 +1,237 @@
...
+        "Can provide a list directly when using through Python",
+    ),
+    starting_seed: int = typer.Option(0, help="Starting seed for random sampling"),
+    preprocess_cmd: str = typer.Option(None, help="Command to run before generation"),
+    postprocess_cmd: str = typer.Option(None, help="Command to run after generation"),
+    partition: str = typer.Option(
+        None, help="Can specify if need interactive jobs or a specific non-default partition"
+    ),
+    time_min: str = typer.Option(None, help="If specified, will use as a time-min slurm parameter"),
+    genselect_args: str = typer.Option(None, help="Can specify extra arguments to prepare the data for genselect"),
```
shtoshni: Calling it `genselect_args` sounds weird. How about removing it and the `extra_arguments` can take care of this.

Kipok: how about genselect_preprocess_args? I think those are used in the preprocess tasks and main extra_arguments are used in generation

shtoshni: How about just `preprocess_args`? Calling it `genselect_preprocess_args` when using `genselect` pipeline would be redundant imo.

nemo_skills/pipeline/generate.py
```diff
@@ -557,126 +194,95 @@ def generate(
...
-        and 'generation_task_type' in ctx.obj
-    ):
-        generation_task = ctx.obj['generation_task_type']  # type: type(GenerationTask)
-        assert issubclass(
-            generation_task, GenerationTask
-        ), f"`generation_task_type` must be a subclass of GenerationTask"
-        cmd_script = generation_task.get_generation_module()
-        cmd_extra_args = generation_task.get_generation_default_args()
-        cmd_script = f"{cmd_script.strip()} {cmd_extra_args.strip()}"
+    if generation_module is not None and generation_type is not None:
```
titu1994, line 214: Hmm dont we always have to pass at least generation_type=generate explicitly then? What about check if both are none and instead defaulting generation_type to generate?

nemo_skills/pipeline/genselect.py
```diff
@@ -0,0 +1,237 @@
...
+import typer
+
+import nemo_skills.pipeline.utils as pipeline_utils
+from nemo_skills.pipeline.app import app, typer_unpacker
+from nemo_skills.utils import get_logger_name, setup_logging, str_ids_to_list
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+def get_genselect_cmd(
```
titu1994, line 26: duplicate from the util func?

Kipok: this one is different, it's genselect specific

nemo_skills/pipeline/generate.py
Kipok: that's how it works right now - if both are None, the generation_type will default to generate right below. If either is specified explicitly, we just use that

docs/pipelines/decontamination.md
```diff
@@ -52,15 +52,17 @@ run_cmd(
...
+generate(
+    cluster="local",
+    generation_type="check_contamination",
+    input_file="/workspace/math-contamination-retrieved.jsonl",
+    output_dir="/workspace/math-contamination-results",
+    model="meta/llama-3.1-405b-instruct",
+    server_type="openai",
+    server_address="https://integrate.api.nvidia.com/v1",
+)
+
```
shtoshni, line 65:
```suggestion
```
```

Comment (shtoshni) [inline docs/pipelines/decontamination.md]: Close the code block. Missing ticks

File: nemo_skills/pipeline/generate.py [hunk 5]
```diff
@@ -557,126 +194,95 @@ def generate(
...
+        raise ValueError("Cannot specify both generation_module and generation_type. ")
+    if generation_module is None:
+        generation_module = GENERATION_MODULE_MAP[generation_type or GenerationType.generate]
 
+    generation_task = importlib.import_module(generation_module)
```
shtoshni, line 219: Need a `try/except` block to catch ill-specified `generation_module`

Kipok: why? It will just fail with a clear error in that case

nemo_skills/inference/llm_math_judge.py
```diff
@@ -56,26 +41,6 @@ class LlmMathJudgeConfig(GenerateSolutionsConfig):
...
-            seed = f'-rs{self.random_seed}' if self.random_seed is not None else ''
-            self.input_file = path.join(self.input_dir, f"output{seed}.jsonl")
-            self.output_file = path.join(self.output_dir, f"output{seed}.jsonl")
-        elif self.input_file is not None and self.input_dir is None:
-            if self.output_file is None:
-                raise ValueError("Output file should be provided if providing `input_file`")
-        else:
-            raise ValueError("`input_file` and `input_dir` cannot be provided at the same time")
-
-    def _get_disallowed_params(self):
```
shtoshni, line 72: We're allowing `code_execution` in LLM math judge?

nemo_skills/inference/reward_model.py
```diff
@@ -15,35 +15,23 @@
...
 
 LOG = logging.getLogger(get_logger_name(__file__))
 
 
 @nested_dataclass(kw_only=True)
 class RewardModelConfig(GenerateSolutionsConfig):
-    """LLM reward model parameters. 
-For the full list of supported parameters, use 'python -m nemo_skills.inference.generate --help'
+    """LLM reward model parameters.
+    For the full list of supported parameters, use 'python -m nemo_skills.inference.generate --help'
```
shtoshni, line 32:
```suggestion
    For the full list of supported parameters, use 'python -m nemo_skills.inference.reward_model --help'
```

nemo_skills/inference/llm_math_judge.py
Kipok: yes, it might be useful for judging if two answers are equivalent

nemo_skills/inference/reward_model.py
Kipok: it should be generate here as that one has extra parameters, not visible with rm --help

nemo_skills/pipeline/genselect.py
```diff
@@ -0,0 +1,237 @@
...
+    )
+    has_tasks = False
+    extra_arguments_original = extra_arguments
+
+    with pipeline_utils.get_exp(expname, cluster_config) as exp:
+        # Add the preprocessing command for genselect
+        preprocess_args = f" ++num_random_seeds={len(random_seeds)} ++output_dir={output_dir} " + (
+            preprocess_args if preprocess_args is not None else ""
+        )
+        preprocess_cmd = f"python -m nemo_skills.inference.genselect_preprocess {preprocess_args}"
```
shtoshni: The `preprocess_cmd` term is overloaded here.. I would change this variable name to say `task_preprocess_cmd` since this is a preprocessing task that we run before running the generations. Thus, from CLI you can still specify a generation job specific `preprocess_cmd`.

nemo_skills/pipeline/openrlhf/ppo.py
```diff
@@ -352,7 +352,7 @@ def ppo_openrlhf(
 
     server_config = None
     if server_type is not None:
-        get_random_port = server_gpus != 8 and not exclusive
+        get_random_port = server_gpus != 8 and not exclusive and server_type != "megatron"
```
shtoshni: This logic is repeated in 5 places right now. I would suggest moving this to `pipeline/utils/server.py`

nemo_skills/pipeline/utils/generation.py
```diff
@@ -0,0 +1,328 @@
...
+            else:
+                out = subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE).stdout.decode("utf-8")
+            outputs.append(out)
+        output = "\n".join(outputs).strip()
+    else:
+        command = f"bash -c '{'; '.join(check_commands)}'"
+        if cluster_config['executor'] == 'slurm':
+            output = get_tunnel(cluster_config).run(command).stdout.strip()
+        else:
+            output = subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE).stdout.decode("utf-8")
```
shtoshni, line 95: Can shorten the logic to always use for loop:

```python
request_size = 16
outputs = []
for i in range(0, len(check_commands), request_size):
    group = check_commands[i : i + request_size]
    command = f"bash -c '{'; '.join(group)}'"
    if cluster_config['executor'] == 'slurm':
        outputs.append(get_tunnel(cluster_config).run(command).stdout.strip())
    else:
        outputs.append(subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE).stdout.decode("utf-8"))
output = "\n".join(outputs).strip()
```

Kipok: please do that in another pr

PR #519, opened 2025-06-13
Support for OpenAI reasoning models

i-vainn:


nemo_skills/inference/server/model.py
```diff
@@ -729,31 +814,26 @@ def _generate_single(
...
-                        stop=stop_phrases,
-                        messages=prompt,
-                        logprobs=top_logprobs is not None,
-                        top_logprobs=top_logprobs,
-                        timeout=timeout,
-                        stream=stream,
-                    )
-                else:
-                    raise
+                if hasattr(e, 'body') and isinstance(e.body, dict) and 'detail' in e.body:
```
Kipok: we can probably remove this part, don't think it's relevant anymore

Kipok (approved):
thanks - we should really do some refactoring here and try to unify logic between different classes to make sure changes like this don't have to touch so many code lines

nemo_skills/inference/server/model.py
i-vainn: Do you mean completely remove this `except openai.BadRequestError as e:` part?

PR #520, opened 2025-06-16
add openai chat completion support for vllm/sglang

makeshn:
file - generate.py

* added prompt_format to GenerateSolutionsConfig - options would be ns and openai - added check in _post_init_validate_params.

* Modified setup_prompt() and log_example_prompt() to avoid modifying prompts if prompt_config is openai.

file - model.py

* added new function in vllm to add support for chatcompletionsendpoint.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -152,9 +152,10 @@ class GenerateSolutionsConfig:
             raise ValueError(f"prompt_format must be either 'ns' or 'openai', got '{self.prompt_format}'")
         
         if self.prompt_format == "openai":
-            assert self.prompt_config is None, "prompt_config is not supported for OpenAI server"
-            assert self.prompt_template is None, "prompt_template is not supported for OpenAI server"
-        
+            assert self.prompt_config is None, "prompt_config is not supported for prompt_format == 'openai'"
+            assert self.prompt_template is None, "prompt_template is not supported for prompt_format == 'openai'"
+        else:
+            assert self.prompt_config is not None, "prompt_config is required when prompt_format == 'ns'"
         for param, default_value in self._get_disallowed_params():
             if getattr(self, param) != default_value:
                 raise ValueError(f"{param} must be {default_value}")
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -55,6 +55,7 @@ class GenerateSolutionsConfig:
     output_file: str  # Where to save the generations
     prompt_config: str  # How to format the data into prompts
     prompt_template: str | None = None  # not required for OpenAI server
+    prompt_format: str = "ns"  # to specify the format of the prompt, "ns" for NeMo-Skills format or "openai" for OpenAI chat format
     code_tags: str | None = None # required when using code execution
     examples_type: str | None = None  # to be able to customize few-shot examples
 
@@ -147,6 +148,9 @@ class GenerateSolutionsConfig:
 
     def _post_init_validate_params(self):
         """Validate that certain parameters are restricted to certain values"""
+        if self.prompt_format not in ["ns", "openai"]:
+            raise ValueError(f"prompt_format must be either 'ns' or 'openai', got '{self.prompt_format}'")
+        
         for param, default_value in self._get_disallowed_params():
             if getattr(self, param) != default_value:
                 raise ValueError(f"{param} must be {default_value}")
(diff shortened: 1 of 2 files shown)


nemo_skills/inference/generate.py
```diff
@@ -55,6 +55,7 @@ class GenerateSolutionsConfig:
     output_file: str  # Where to save the generations
     prompt_config: str  # How to format the data into prompts
```
Kipok: Let's make prompt_config optional as it's not used when prompt_format != ns, right?

nemo_skills/inference/generate.py
```diff
@@ -147,6 +148,9 @@ def _post_init_validate_server(self):
 
     def _post_init_validate_params(self):
         """Validate that certain parameters are restricted to certain values"""
+        if self.prompt_format not in ["ns", "openai"]:
```
Kipok, line 151: Please add assert that prompt_config and prompt_template are both set to None if format != ns

nemo_skills/inference/generate.py
```diff
@@ -367,12 +384,21 @@ def fill_prompt(self, data_point, data):
...
 
     def llm_generate(self, data_points, data, is_async=False):
-        generation_params = {
-            "prompts": [self.fill_prompt(dp, data) for dp in data_points],
-            "stop_phrases": combine_stop_phrases(self.prompt.stop_phrases, self.extra_stop_phrases),
-            **asdict(self.cfg.inference),
-            **self.extra_generate_params,
-        }
+
+        if self.cfg.prompt_format == "openai":
```
Kipok: is this change needed? Shouldn't the original code work fine since the fill_prompt will just return the data_point?

nemo_skills/inference/server/model.py
```diff
@@ -896,7 +896,7 @@ def __del__(self):
 
     def _generate_single(
         self,
-        prompt: str | dict,
+        prompt: str | dict | list,
```
Kipok: I think the dict type is incorrect here, can you please change all throughout this file to `str | list`?

Kipok (commented):
Thanks, overall looks great, but added a few small comments!

Also just so that you know, if prompt_template isn't provided (which should be the case for you), we automatically switch to `server_type: openai` inside setup_llm method in this class. This means that instead of vllm client it will use openai client and so the changes you made for chat completion aren't technically necessary. But since you made them anyway, maybe add a check there that if serve_type is vllm, then we don't switch?

nemo_skills/inference/generate.py
makeshn: done

nemo_skills/inference/generate.py
makeshn: done

nemo_skills/inference/generate.py
makeshn: done

nemo_skills/inference/server/model.py
makeshn: done

makeshn:
@Kipok made the changes - I did not realize that it would default to openai client - I had prompt template not None which led to errors and I added the chat completions part in vllm to resolve that issue. Please let me know if you would like me to refactor and remove the vllm.

nemo_skills/inference/generate.py
```diff
@@ -147,6 +148,13 @@ def _post_init_validate_server(self):
 
     def _post_init_validate_params(self):
         """Validate that certain parameters are restricted to certain values"""
+        if self.prompt_format not in ["ns", "openai"]:
+            raise ValueError(f"prompt_format must be either 'ns' or 'openai', got '{self.prompt_format}'")
+        
+        if self.prompt_format == "openai":
+            assert self.prompt_config is None, "prompt_config is not supported for OpenAI server"
+            assert self.prompt_template is None, "prompt_template is not supported for OpenAI server"
+
```
```suggestion
        else:
            assert self.prompt_config is not None, "prompt_config is required when prompt_format == 'ns'"
```

nemo_skills/inference/generate.py
```diff
@@ -147,6 +148,13 @@ def _post_init_validate_server(self):
 
     def _post_init_validate_params(self):
         """Validate that certain parameters are restricted to certain values"""
+        if self.prompt_format not in ["ns", "openai"]:
+            raise ValueError(f"prompt_format must be either 'ns' or 'openai', got '{self.prompt_format}'")
+        
+        if self.prompt_format == "openai":
+            assert self.prompt_config is None, "prompt_config is not supported for OpenAI server"
```
```suggestion
            assert self.prompt_config is None, "prompt_config is not supported for prompt_format == 'openai'"
```

nemo_skills/inference/generate.py
```diff
@@ -147,6 +148,13 @@ def _post_init_validate_server(self):
...
+            assert self.prompt_template is None, "prompt_template is not supported for OpenAI server"
```
```suggestion
            assert self.prompt_template is None, "prompt_template is not supported for prompt_format == 'openai'"
```

Kipok (approved):

Kipok:
> @Kipok made the changes - I did not realize that it would default to openai client - I had prompt template not None which led to errors and I added the chat completions part in vllm to resolve that issue. Please let me know if you would like me to refactor and remove the vllm.

all good, we wanted to add native support anyway, but ideally reusing the same class for both of those. Probably @i-vainn will take a pass at refactoring this further to unify the code sometimes soon

Issue #525, opened 2025-06-18
merge_new_summary.py import error


xinyu-dev:
**Version**: 
NeMo-Skill 0.6.1

**Issue**: 
[merge_new_summary.py](https://github.com/NVIDIA/NeMo-Skills/blob/main/recipes/openmathreasoning/scripts/merge_new_summary.py) has an import error: 
```python
from nemo_skills.code_execution.math_grader import extract_answer
```
Error: `code_execution` module not found

**Remedy**: 
See [my branch ](https://github.com/xinyu-dev/NeMo-Skills/blob/250614-tir-0.6.1/recipes/openmathreasoning/scripts/merge_new_summary.py)

Note that by simply changing the above import to 

```python
from nemo_skills.evaluation.math_grader import extract_answer
```

Fixes the the import issue, but there is another issue that the output JSONL file not found. Hence I added [this line](https://github.com/xinyu-dev/NeMo-Skills/blob/7201820e341a925ce79e95f51be1623506a9267c/recipes/openmathreasoning/scripts/merge_new_summary.py#L170) to create a new output file.

Kipok:
@xinyu-dev thanks for reporting this, can you please send a PR with your fixes?

@shtoshni can you please review the changes?

Kipok:
fixed in https://github.com/NVIDIA/NeMo-Skills/pull/528

PR #528, opened 2025-06-19
updated merge_new_summary

xinyu-dev:
Updated the merge_new_summary.py issue when running the openmath TIR pipeline. See[ Issue 525. ](https://github.com/NVIDIA/NeMo-Skills/issues/525)

Issues include: 
1. import not correct
2. a bug saying the output jsonl file isn't found.

not a bug but improvements made for my cases
3. Find the start and end tags. If either is not found, return entire generation instead of None. 
4. Between the trimmed traces and summary, add a word "Summary: " for better readability.


recipes/openmathreasoning/scripts/merge_new_summary.py
```diff
@@ -62,20 +63,21 @@ def select_best_summary(valid_summaries):
...
-
-    end_tag_position = reasoning_generation.find(end_tag)
-    if end_tag_position == -1:
-        if strict_end_tag:
-            return None
-        else:
-            reasoning_generation = reasoning_generation + end_tag
-            reasoning_trace = reasoning_generation
+        # If no start tag found, wrap the entire generation with tags
+        reasoning_trace = start_tag + "\n" + reasoning_generation + "\n" + end_tag
```
shtoshni: Just to be safe we should also check if the end_tag is not present and only then add the end_tag if it's missing. So you can make the `else` to another `if` where we check if the `end_tag_position == -1` and add the `end_tag` if required.

recipes/openmathreasoning/scripts/merge_new_summary.py
```diff
@@ -62,20 +63,21 @@ def select_best_summary(valid_summaries):
...
     else:
-        reasoning_trace = reasoning_generation[:end_tag_position + len(end_tag)]
+        end_tag_position = reasoning_generation.find(end_tag)
+        if end_tag_position == -1:
+            if strict_end_tag:
+                return None
+            else:
+                reasoning_generation = reasoning_generation + end_tag
```
shtoshni: It would be good to follow a consistent format while adding these tags. A few lines up, a newline is added along with the tag. Would be good to keep the tag addition format consistent.

recipes/openmathreasoning/scripts/merge_new_summary.py
```diff
@@ -62,20 +63,21 @@ def select_best_summary(valid_summaries):
...
+                reasoning_trace = reasoning_generation
```
shtoshni: can remove this line by directly assigning the value to `reasoning_trace` in the prior line.

recipes/openmathreasoning/scripts/merge_new_summary.py
```diff
@@ -141,7 +143,7 @@ def format_reasoning_trace_with_summary(reasoning_file, summary_dir, start_tag,
             # Select the best summary
             best_summary = select_best_summary(valid_summaries)                
             # Combine the trimmed reasoning trace with the best summary
-            combined_generation = trimmed_reasoning_trace + best_summary["generation"]
+            combined_generation = trimmed_reasoning_trace + "\n\nSummary:\n" + best_summary["generation"]
```
shtoshni: is this added string required? The `end_tag` indicates the start of summary as well. We probably don't need this added string but let me know if you had other reasons in mind.

recipes/openmathreasoning/scripts/merge_new_summary.py
```diff
@@ -94,8 +96,8 @@ def trim_reasoning_generation(reasoning_generation, start_tag, end_tag, strict_e
                 reasoning_trace[rightmost_match.end():].split("\n\n")[0]
             )
 
-            # If the end tag is not present, add it
-            if end_tag not in reasoning_trace:
+            # If we originally had tags and the end tag is not present, add it
+            if start_tag_position != -1 and end_tag not in reasoning_trace:
```
shtoshni: We will most likely not reach this code if we structure the above `if else` part.

recipes/openmathreasoning/scripts/merge_new_summary.py
shtoshni: If you want it for your usecase, we can keep the "Summary" separator optional. So we can have a param say "summary_start_tag". Could be None by default but you can optionally set it to  "\n\nSummary:\n"

shtoshni:
@xinyu-dev - Do you want to take a look at my comments?

xinyu-dev:
Thanks so much for your comments, @shtoshni ! I revised it accordingly. Let me know if there is any issues.

shtoshni:
Hey @xinyu-dev, 

Simplified and fixed the code about handling tags. 

Earlier code:
```
if start_tag_position == -1:
    # If no start tag found, wrap the entire generation with tags
    reasoning_trace = start_tag + "\n" + reasoning_generation
    if end_tag_position == -1:
        reasoning_trace += "\n" + end_tag
else:
    if end_tag_position == -1:
        if strict_end_tag:
            return None
        reasoning_trace = reasoning_generation + "\n" + end_tag
    else:
        reasoning_trace = reasoning_generation[: end_tag_position + len(end_tag)]
```

New code:
```
# Handle missing start tag
if start_tag_position == -1:
    reasoning_trace = f"{start_tag}\n{reasoning_generation}"
else:
    reasoning_trace = reasoning_generation

# Handle missing end tag
if end_tag_position == -1:
    if strict_end_tag:
        return None
    reasoning_trace += f"\n{end_tag}"
else:
    reasoning_trace = reasoning_trace[:end_tag_position + len(end_tag)]
```

In the former, we miss out on returning `None` if `strict_end_tag` in the case when both `start_tag_position` and `end_tag_position` are -1. 
Also, the logic about adding `end_tag` was repeated. 
Hopefully, this implementation is okay with you!

xinyu-dev:
Thanks so much! Appreciate your review and work!

PR #529, opened 2025-06-19
Add support for MMLU 5-shot eval for base models

shan18:
* Bugfix: Update the key names in the few shot examples of `mmlu-pro`
* Add support for evaluating base models on mmlu 5-shot setting.
* Using the `dev` split from the mmlu dataset for the few shot prompts.


nemo_skills/prompt/few_shot_examples/__init__.py
```diff
@@ -15,11 +15,13 @@
...
 
 examples_map = examples_gsm8k.copy()
 examples_map.update(examples_math)
 examples_map.update(examples_lean4)
 examples_map.update(examples_mmlu_pro)
-assert len(examples_map) == len(examples_gsm8k) + len(examples_math) + len(
-    examples_lean4
-) + len(examples_mmlu_pro), "Duplicate keys in examples!"
+examples_map.update(examples_mmlu)
+assert len(examples_map) == len(examples_gsm8k) + len(examples_math) + len(examples_lean4) + len(
```
Kipok: can you please split this into multiple expressions by adding a variable for the length? Or some other way, the current assert is really hard to read with its weird three-line formatting

Kipok (approved):
thanks, looks good to me, just a small comment about formatting! But might be good to have a review from @ekmb as well

nemo_skills/prompt/few_shot_examples/__init__.py
shan18: Done

PR #530, opened 2025-06-21
Adding dapo and config file ability

tmfs10:
Adding ability to call arbitrary verl modules (including dapo module) as well as pass a yaml config file for settings

### DIFF
--- a/nemo_skills/pipeline/verl/ppo.py
+++ b/nemo_skills/pipeline/verl/ppo.py
@@ -42,6 +42,7 @@ class PPOVerlTask:
     timeout: str
     extra_arguments: str = ""
     logging_params: str = ""
+    script_module: str = "verl.trainer.main_ppo"
 
     def get_ray_launch_cmd(self):
         cmd = "ray job submit --address='http://127.0.0.1:8265' -- "
@@ -116,7 +117,7 @@ class PPOVerlTask:
         return cmd
 
     def get_script_module(self):
-        return "verl.trainer.main_ppo"  # Must use https://github.com/titu1994/verl/
+        return self.script_module
 
     def get_job_cmd(self):
         ray_job_cmd = self.get_ray_launch_cmd()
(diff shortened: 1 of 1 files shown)


wedu-nvidia (commented):
Thanks, Sid, for adding this. So the user needs to define the Verl config for Dapo, right?
But if we would like to support dapo, can we have a separate file, e.g., dapo.py to support just like we did for ppo? What do you think @gwarmstrong ?

tmfs10:
> Thanks, Sid, for adding this. So the user needs to define the Verl config for Dapo, right? But if we would like to support dapo, can we have a separate file, e.g., dapo.py to support just like we did for ppo? What do you think @gwarmstrong ?

For dapo, you don't need to supply a verl config. You just need to give the script_module option as recipe.dapo.main_dapo and it will pick up on the default dapo config. You can override params with extra_args.

You can also of course supply your custom yaml config (I added in that option) but it's not necessary.

The reason I did it this way is there are a lot of recipes in main verl branch (https://github.com/volcengine/verl/tree/main/recipe). It is possible we may want to try those. It seems the most economical to just give the option to change the script module instead of writing a new file for every option you may want to try. In addition, this makes it easier for us to add in our own entrypoints without having to modify skills every time.

wedu-nvidia:
> > Thanks, Sid, for adding this. So the user needs to define the Verl config for Dapo, right? But if we would like to support dapo, can we have a separate file, e.g., dapo.py to support just like we did for ppo? What do you think @gwarmstrong ?
> 
> For dapo, you don't need to supply a verl config. You just need to give the script_module option as recipe.dapo.main_dapo and it will pick up on the default dapo config. You can override params with extra_args.
> 
> You can also of course supply your custom yaml config (I added in that option) but it's not necessary.
> 
> The reason I did it this way is there are a lot of recipes in main verl branch (https://github.com/volcengine/verl/tree/main/recipe). It is possible we may want to try those. It seems the most economical to just give the option to change the script module instead of writing a new file for every option you may want to try. In addition, this makes it easier for us to add in our own entrypoints without having to modify skills every time.

Thanks so much for the reply here. Can you also provide a sample code to submit the jobs.

wedu-nvidia (approved):

PR #534, opened 2025-06-25
add summarize results support for varying length

wedu-nvidia:
Add support for results summarization under varying length


nemo_skills/evaluation/metrics/utils.py
```diff
@@ -20,11 +20,15 @@
...
-def read_predictions(predictions, line_idx, file_handles):
+def read_predictions(predictions, line_idx, file_handles, sequence_length = None):
     data = []
     for file_idx, prediction in enumerate(predictions):
         try:
             prediction_dict = json.loads(prediction)
+            if sequence_length != None:
+                if int(prediction_dict['num_generated_tokens']) > sequence_length: 
+                    prediction_dict['is_correct'] = False
+                    prediction_dict['predicted_answer'] = None
```
Kipok: Maybe instead we modify the base class for metrics to add a new method with something like `get_incorrect_sample` that will return `{'is_correct': False, 'predicted_answer': None}` for math metrics, but e.g. `{'passing_base_tests': False, 'passing_plus_tests: False}` for code metrics and similar different dicts for other classes. And then we call that method to fill-in the dictionaries here (and base model can raise NotImplementedError, so that if it's not implemented, we have a clear error, instead of current incorrect behavior, since this logic is math-specific)

wedu-nvidia: sure, I can do that.

wedu-nvidia: I've updated the code—let me know if it aligns with your expectations.

nemo_skills/evaluation/metrics/base.py
```diff
@@ -77,6 +77,31 @@ def reset(self):
         self.avg_tokens = 0
         self.eval_dict = defaultdict(lambda: defaultdict(float))
 
+    @classmethod
+    def _get_incorrect_sample(cls, prediction, metric_type):
```
Kipok: Instead of this big if, let's move this into subclasses. So the base just throws NotImplementedError, but the subclasses should provide a valid version matching the ifs you have here. And then you don't need to pass in metric_type and probably also don't need to pass in prediction, the class should just return a new dict with all necessary information for an incorrect sample

Kipok: and then you can use self.calculators['all'].get_incorrect_sample() to get that dict. Probably best not to pass it into the read_predictions as it's a more general function but just go over predictions one more time inside compute_metrics.py and replace anything that's not passing length constraint

wedu-nvidia: ok, sure. It is a better an idea and I will do

PR #536, opened 2025-06-27
Adding scicode + refactoring for evaluators and letting use custom generation module in eval

Kipok:


nemo_skills/inference/eval/scicode.py
```diff
@@ -0,0 +1,183 @@
...
+                llm_output['generation'] = ''
+            extracted_python = extract_python_script(llm_output['generation'])
+            previous_llm_code[cur_step] = extracted_python
+            # TODO: save those as separate entries so that we can preserve intermediate progress on reruns
+            task_solutions[f"{problem_id}.{cur_step}"] = f'{previous_code}\n{extracted_python}'
+
+        # generation is a dict["problem_id.subtask_step": full_solution] here
+        return {'generation': task_solutions, 'num_generated_tokens': total_generated_tokens}
+
+    def llm_generate(self, data_points, data, is_async=False):
```
Kipok, line 137: @shtoshni this and get_llm_generations is what you need to support multi-turn generations

Kipok:
GPU tests are passing locally and I'm able to match scicode scores as reported by aai for qwen3-reasoning (although not fully matching for gpt4.1). I think it should be safe to use, and we can fix any issues in the follow up prs if needed

PR #537, opened 2025-06-30
Update STEM benchmarks to follow AAI methodology

ekmb:


nemo_skills/evaluation/evaluator/mcq.py
```diff
@@ -24,11 +25,27 @@
...
+        # extract prediction from boxed{}
+        parsed = extract_answer(text)
+        if parsed is not None and len(parsed) != 1:
+            match = re.findall(r"\b[A-J]\b(?!.*\b[A-J]\b)", parsed, re.DOTALL)
+            if len(match) > 0:
+                parsed = match[-1].strip()
+
+        # adapted from https://artificialanalysis.ai/methodology/intelligence-benchmarking#intelligence-index-evaluation-suite-overview
+        if parsed is None:
+            match = re.findall(r"(?i)[\*\_]{0,2}Answer[\*\_]{0,2}\s*:[\s\*\_]{0,2}\s*([A-Z])(?![a-zA-Z0-9])", text)
```
dgtm777, line 38: Why do we look here for A-Z answers, and in the previous condition for A-J only?

nemo_skills/prompt/config/generic/mcq-10choices-aai-boxed.yaml
```diff
@@ -0,0 +1,8 @@
+# Default prompt for multuple choice benchmarks with ten answer choices (e.g. MMLU-Pro). It's intended for models trained to follow \boxed{{}} format and for generaral instruction-following models.
```
Kipok, line 1: if prompts are specific for some particular evaluation setup, let's probably move them to the config/eval folder

Kipok:
merging as is to unblock experiments and let's fix any remaining issues in a follow up pr

PR #538, opened 2025-06-30
AIMO-2 and trtllm-serve async inference 

alessiodevoto:
### Summary 
This PR implements the inference pipeline of [AIMO-2 paper](https://arxiv.org/abs/2504.16891) using `ns serve` with the`trtllm-serve` option. This is more flexible as avoids the dependency on specific `trtllm` versions.

### Main changes
The main changes concern:
- `OpenAIAPIModel` -> added tracking for ongoing requests, so that a generation can be interrupted by closing the corresponding API request. This is backward compatible (the tracking is optional).
- `CodeExecutionWrapper ` -> the main goal here is to stream requests and stop after N seconds / N requests completed / N requests with the same answer. It is important to also return all responses generated up to interruption. I am not sure the code I added here should be added in this place as it is a bit AIMO-2 specific.


nemo_skills/inference/model/base.py
```diff
@@ -313,19 +319,82 @@ def __init__(
...
+        Cancel a specific generation by ID.
+        Returns True if the generation was found and cancelled, False otherwise.
+        """
+        generation_info = self._unregister_generation(gen_id)
+        if generation_info is None:
+            return False
+        
+        try:
+            response = generation_info['response']
+            if hasattr(response, 'close'):
```
Kipok: what's the case when this would be False?

nemo_skills/inference/model/base.py
```diff
@@ -313,19 +319,82 @@ def __init__(
...
+                response.close()
+            return True
+        except Exception as e:
```
Kipok: If there is some error here, I'd rather fail by default, instead of emitting a warning. Users can have a try - except on their side if they want this behavior, but I think it's safer to not ignore the errors as a default behavior

nemo_skills/inference/model/base.py
```diff
@@ -414,50 +520,64 @@ def _parse_chat_completion_response(self, response) -> dict:
...
+                    if not cur_delta:
+                        result["generation"] = ""
+
+                yield result
+        finally:
+            # Always unregister the generation when streaming is complete
+            self._unregister_generation(gen_id)
+
+    # Backward compatibility method
+    def close_active_api_calls(self):
```
Kipok: why do we need this? I don't think we had that method before

alessiodevoto: Right, removed

nemo_skills/inference/model/base.py
alessiodevoto: Agree, I will change this

nemo_skills/inference/model/base.py
alessiodevoto: In the case where the response is not a streaming one ---> This made me realize that there is no point in registering non-streaming requests (unless we use openai.AsyncOpenAI as a server, but that would require major modifications I think). So I refactored everything to only keep track of streaming requests.

nemo_skills/inference/model/code_execution.py
```diff
@@ -544,4 +544,130 @@ def _stream_single(
                 else:
                     current_full_prompt += formatted_code_output
             else:
-                break
\ No newline at end of file
+                break
+    
+
+    @staticmethod
```
alessiodevoto: Removing all this as we agreed

PR #539, opened 2025-06-30
Add SGLang to Verl

smahdavi4:
To use SGlang with verl, it requires setting this environment variable.


Kipok:
if it's just for qwen3, should we then not have it default and instead each user can put inside their cluster config?

PR #560, opened 2025-07-08
Support for client side parsing for BFCL

shtoshni:
By supplying the params: 
- `use_client_parsing` -> True
- `model_name` -> Model name such as "Qwen/Qwen3-4B-FC"

the bfcl implementation will use BFCL's handling of prompt formatting and tool parsing.

### DIFF
--- a/nemo_skills/inference/eval/bfcl.py
+++ b/nemo_skills/inference/eval/bfcl.py
@@ -16,7 +16,7 @@ import logging
 import sys
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import field
-
+import json
 import hydra
 from dataclasses import asdict, field
 from omegaconf import OmegaConf
@@ -27,6 +27,7 @@ from nemo_skills.inference.model import server_params
 from nemo_skills.utils import get_help_message, get_logger_name, nested_dataclass, setup_logging
 
 from nemo_skills.dataset.bfcl_v3.utils import func_doc_language_specific_pre_processing, convert_to_tool
+from bfcl_eval.constants.model_config import local_inference_model_map
 
 LOG = logging.getLogger(get_logger_name(__file__))
 
@@ -41,8 +42,10 @@ class BFCLGenerationConfig(GenerateSolutionsConfig):
     server: dict = field(default_factory=dict)
 
     remove_thinking: bool = True
+    use_client_parsing: bool = False
+    model_name: str | None = None
 
-    
+
     def _post_init_validate_params(self):
         """Validate that certain parameters are restricted to certain values"""
         if self.prompt_format not in ["ns", "openai"]:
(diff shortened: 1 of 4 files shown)


Kipok (approved):
Thanks! If you didn't yet, please double check that you get expected scores from one of the models, e.g. qwen3-8b

Kipok:
oh, just see that it's a draft - let me know when it's ready for review and I can have another look

nemo_skills/pipeline/utils/eval.py
shtoshni, line 1: Made this change because the original code fails when we want to evaluate a particular subtask such as "bfcl_v3.simple" because the eval folder has the structure `bfcl_v3/simple`, and simple is not a benchmark.

nemo_skills/inference/eval/bfcl.py
```diff
@@ -41,8 +43,10 @@ class BFCLGenerationConfig(GenerateSolutionsConfig):
     server: dict = field(default_factory=dict)
 
     remove_thinking: bool = True
+    use_client_parsing: bool = False
```
Kipok, line 46: I'd say this should default to True? Seems like a more common way as most models are supported

nemo_skills/inference/eval/bfcl.py
```diff
@@ -55,6 +59,40 @@ def _post_init_validate_params(self):
         for param, default_value in self._get_disallowed_params():
             if getattr(self, param) != default_value:
                 raise ValueError(f"{param} must be {default_value}")
+        
+        if self.use_client_parsing:
+            if self.model_name is None:
+                raise ValueError("model_name is required when use_client_parsing is True")
+            
+            # Add FC by default
+            if "-FC" not in self.model_name[-3:]:
```
Kipok, line 68: Should we have an argument to disable this? Right now it's impossible to have a model name without -FC, right?

nemo_skills/pipeline/utils/eval.py
Kipok: Can you please share the command that failed with old version? I'm not sure if your change is correct

nemo_skills/inference/eval/bfcl.py
shtoshni: Actually, supporting models without FC doesn't make much sense since all the top models support FC mode anyways, and the performance is always much better - https://gorilla.cs.berkeley.edu/leaderboard.html. I have not tested non-FC mode for models, so was avoiding their support. Can be done and may require a few more cases to be added, but I don't see much point in supporting this mode.

nemo_skills/inference/eval/bfcl.py
shtoshni: Sure, we can make it false. I don't know what the frequent use cases for the alignment team are. I would imagine they test the other models once, and then just focus on their own models.

Kipok: Right, but we aren't only adding this for the "alignment team", we are adding it for anyone using NeMo-Skills. And I think broadly, its's more common to evaluate models that are already supported in bfcl

PR #563, opened 2025-07-09
Prepare pyproject.toml

activatedgeek:
A `pyproject.toml` is now the recommended approach for Python packages. I've updated the current `pyproject.toml` to reflect all build configurations in `setup.py`.

Also see [Is setup.py deprecated?](https://packaging.python.org/en/latest/discussions/setup-py-deprecated).

This also helps with more modern and super fast project managers like [`uv`](https://docs.astral.sh/uv/) and external tooling support for the future.

TODO:
  - [x] Remove `setup.py` reference from Dockerfile.


Kipok:
can we then remove setup.py completely? Otherwise we have two sets of equivalent configurations and always need to remember to change both

activatedgeek:
@Kipok `setup.py` should now be removed.

titu1994:
Why should setup.py be removed ? It is not deprecated as per your own link, and toml config alone does not allow code to be executed. 

Adding support for project.toml does not equate to removal of setup.py. Dynamic execution at setup time can be required often enough

activatedgeek:
> Adding support for project.toml does not equate to removal of setup.py. Dynamic execution at setup time can be required often enough

We will only require `setup.py` if there are custom build steps for `setuptools`. However, `pyproject.toml` will not hinder us because we can just switch away from `setuptools` to other build managers and allow dynamic injection of build steps. `hatch` allows it via just pointing to other modules and fairly easy to migrate to given that we do not have any non-standard build rules in the package right now.

In any case, the key consideration here is not the fact that we want to replace `setup.py` with `pyproject.toml`, but rather that it is the preferred mode of modern Python package management. And it has a very healthy ecosystem of plugin tooling and widespread adoption.

I consider the deletion of `setup.py` is merely a convenience so that we don't forget about keeping track in both files. Unless there are ongoing plans for custom build steps or supporting old Python distributions, it is unnecessary to keep it around.

> Need to test properly before merging that installation works well in different envs

I can test, but will require some guidance on what to look for.

Kipok (approved):

PR #569, opened 2025-07-11
Add recipe for OpenCodeReasoning

titu1994:
Recipe for replicating OpenCodeReasoning dataset using R1

### DIFF
--- a/recipes/opencodereasoning/scripts/output_processing.py
+++ b/recipes/opencodereasoning/scripts/output_processing.py
@@ -1,0 +1,143 @@
+import ast
+
+
+def check_generation(response, keep_explanations=False, do_ast_check=False) -> bool:
+    """Obtained from https://github.com/nickrosh/evol-teacher/blob/main/generate_evol.py
+
+    Check the generated instruction with several checks. If it returns True,
+    then the instruction should be discarded
+
+    Reviews the generation from a language model and checks if it is valid.
+
+    Args:
+        response: The response to be checked.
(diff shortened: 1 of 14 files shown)


recipes/opencodereasoning/scripts/prepare_questions.py
```diff
@@ -0,0 +1,106 @@
...
+        os.makedirs(output_dir)
+
+    # Load the OCR2 dataset and prepare questions
+    ocr2_dataset = load_dataset("nvidia/OpenCodeReasoning-2")
+
+    # Save only subset that contains unique question_ids
+    unique_values = set()  # To keep track of unique values encountered
+    first_occurrence_indices = []
+
+    for ocr2_ds in [ocr2_dataset["python"]]:
```
titu1994, line 79: This step is currently too disk-space intensive - we need to download 200GB of external datasets, and then 300GB of OCR 2 just to get 34K questions. Ill think about how to rework this later - maybe release just a small question-id dataset for this indexing - at least we can skip the 300GB download of OCR 2 then .

recipes/opencodereasoning/scripts/output_processing.py
```diff
@@ -0,0 +1,143 @@
+import ast
```
Kipok, line 15: license

titu1994: Done

Kipok (approved):

PR #592, opened 2025-07-16
Fixes for genselect pipeline

Kipok:


nemo_skills/inference/genselect.py
```diff
@@ -119,10 +109,8 @@ def postprocess(self):
         single_answer_instances = [json.loads(line) for line in open(single_answer_instances_file, "r")]
 
         input_file = self.cfg.output_file
-        if self.cfg.dataset is not None:
-            benchmark_dir = self.cfg.dataset
-        else:
-            benchmark_dir = "math"
+        # TODO: use last part of input_dir?
```
shtoshni, line 112: We need to make sure that the name that we extract is an actual benchmark so that summarize_results doesn't complain. I think that can be done via a common implementation in utils which can be beneficial for `summarize_results` pipeline as well.

shtoshni: Obviously that's for another PR

PR #593, opened 2025-07-16
Support for PutnamBench dataset

fchen97:
Successfully added the dataset to the repository.
Make the format consistent with other formal datasets such as miniF2F and ProofNet.
655 out of 660 theorems pass lean4 compilation. The remaining 5 that fail the compilation are likely due to Mathlib or lean4 version issues.


Kipok (approved):
let's maybe use lower-case with a dash for the name? Like `putnam-bench`. Just to keep it consistent with other datasets where we don't use uppercase

PR #599, opened 2025-07-18
Support custom containers in run_cmd

ludwig-n:
Currently, `ns run_cmd` fails if the `container` argument is not defined in the cluster config. Now it will fall back to passing the `container` argument directly to SlurmExecutor or DockerExecutor, allowing the use of dockerhub paths or .sqsh files.


Kipok (approved):
thanks!

PR #607, opened 2025-07-22
update grpo config para

wedu-nvidia:
Update grpo config to be compatible latest nemo-rl


Kipok:
should we upgrade the container for this to work without custom nemo-rl mounted?

wedu-nvidia:
I updated the grpo config and dockerfile to support latest nemo-rl

PR #612, opened 2025-07-23
Lean4 TIR execution support

gwarmstrong:
Supports Lean4 Compilation during generation


nemo_skills/inference/model/code_execution.py
```diff
@@ -56,6 +64,8 @@ def __init__(self, model: BaseModel, sandbox: Sandbox, config: CodeExecutionConf
         else:
             self._can_cancel_generations = False
 
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model.model)
```
Kipok: This is not always possible. E.g. we can use closed model like gpt or claude. Or it might be a custom model checkpoint which doesn't work with from_pretrained. So best not to put this logic into this general class or at least handle errors gracefully. But I'd probably not use this yet, we do want to add tokenizer on the client side, but I think it should be inside main model, not code execution model and probably done a bit differently

nemo_skills/inference/model/code_execution.py
```diff
@@ -208,27 +218,21 @@ def _generate_single(
...
+                code_execution_time_start, execution_dict = self.execute_generated_code(prompt, code_begin, code_end, output, session_id)
                 remaining_code_executions = None
                 if self.config.add_remaining_code_executions:
                     remaining_code_executions = effective_max_code_executions - generation_index - 1
                 # adding code output to the prompt
                 code_output = format_code_output(
                     execution_dict, code_output_begin, code_output_end, code_output_format, remaining_code_executions
                 )
-                
+                request['tokens_to_generate'] = max_new_tokens - len(self.tokenizer.encode(request['prompt'], add_special_tokens=True))
```
Kipok: I don't think this is the right change. tokens_to_generate means whatever comes after prompt. So we shouldn't subtract the prompt, but should subtract the code output. But in any case, let's probably remove it from this pr

nemo_skills/inference/model/code_execution.py
```diff
@@ -249,6 +252,36 @@ def _generate_single(
...
+        execution_dict, session_id = self.sandbox.execute_code(
+                    generated_code=extracted_code,
+                    language=self.config.code_execution_language,
+                    timeout=self.config.code_execution_timeout,
+                    max_output_characters=self.config.max_code_output_characters,
+                    session_id=session_id,
+                    traceback_verbosity=self.config.sandbox_traceback_verbosity,
+                )
+
+        return code_execution_time_start,execution_dict
```
Kipok:
```suggestion
        return code_execution_time_start, execution_dict
```

nemo_skills/inference/model/code_execution.py
```diff
@@ -249,6 +252,36 @@ def _generate_single(
...
+                        "import Mathlib",
+                        "",
+                        "set_option maxHeartbeats 0",
+                        "",
+                        "open BigOperators",
+                        "open Real",
+                        "open Nat",
+                        "open Topology",
+                        "",
+                    ])
```
fchen97: I think it's better to support custom headers. For example some times I want to have
```
import Mathlib
open Filter Topology
```

nemo_skills/prompt/config/lean4/formal-proof-reasoning-execution.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+user: |-
+  In your solution you MUST strictly follow these instructions:
+  1. You may check intermediate proof steps with Lean4.
+  2. For Lean4 code  use the following template,
+    ```lean4
+    /-- Lean4 code --/
+    ```
+  3. The final proof's theorem MUST match the theorem in the prompt.
+  4. Put your final proof within **FINAL ANSWER**```lean4 ```
```
fchen97:
```suggestion
```
I think it overlaps with the following sentence. And it is also missing a ```\n```

nemo_skills/prompt/config/lean4/formal-proof-reasoning-execution.yaml
fchen97, line 1: It seems that one needs to change a few functions to support two code_begin tags: _"\```lean4\n" and "\```lean\n"_

gwarmstrong: Great, I created an issue to track it #613

nemo_skills/inference/model/code_execution.py
gwarmstrong: The issue I was having, and the reason I put this here, is that I was regularly getting errors that break the whole generation loop when the entire context length exceeds the model max (32k tokens I believe, in the case of LIMO). So this was to prevent issues when, due to output being formatted in, the "tokens to generate" plus existing content would  exceed model max and break an entire generation run.

That being said, this can probably also be reasonably controlled with overall max tokens to generate. I'll remove it for now and run another test with LIMO to make sure.

nemo_skills/inference/model/code_execution.py
gwarmstrong: Agreed this is ideal. It was a little tricky to incorporate them in the config the last time I tried, but I'll take another pass at it.

gwarmstrong: added in 9e28c2817819fda206acad5b671809843ccb5989

You can use it in the generate pipeline with something like:
```python
generate(
    ...
    ctx=wrap_arguments(
        ...
        '++server.code_execution.code_execution_headers=\'["import Mathlib", "open Filter Topology"]\''
    )
)

File: nemo_skills/inference/model/code_execution.py (hunk 2 above)
Comment (gwarmstrong) [reply nemo_skills/inference/model/code_execution.py]: Okay I updated and it seems fine now with this code removed--it's possible it had interactions with other bugs in my initial implementation. I am running a longer run just to make sure.

File: nemo_skills/inference/model/code_execution.py (hunk 1 above)
Comment (gwarmstrong) [reply nemo_skills/inference/model/code_execution.py]: Removed in 2ec5f0b0d0861716b4dd8a6b88a0618632aaa84a

File: nemo_skills/inference/model/base.py [hunk 6]
```diff
@@ -54,11 +54,13 @@ def __init__(
         port: str = '5000',
         ssh_server: str | None = None,
         ssh_key_path: str | None = None,
+        model: str | None = None,
     ):
         self.server_host = host
         self.server_port = port
         self.ssh_server = ssh_server
         self.ssh_key_path = ssh_key_path
+        self.model = model
```
Kipok: Remove this then? And all other places where you pass the model in

nemo_skills/inference/model/code_execution.py
gwarmstrong: @Kipok Okay longer run has been running for a while and it seems everything is effectively controlled with `++inference.tokens_to_generate`

nemo_skills/inference/model/code_execution.py
fchen97: Thank you! But I think it can be a data field instead? If you look at miniF2F/proofNet/PutnamBench, there is a data field called 'header' for each data entry. Do you think it makes sense to have different header for different theorems?

nemo_skills/inference/model/code_execution.py
```diff
@@ -34,6 +37,8 @@
 class CodeExecutionConfig:
     max_code_output_characters: int = 1000
     code_execution_timeout: float = 10.0
+    code_execution_language: str = 'python'  # could be python, lean4
```
i-vainn, line 40: Have you considered dynamically detecting the language instead (e.g. based on `code_begin` & `code_end`)? I'm wondering if this could be useful in the future

nemo_skills/inference/model/code_execution.py
gwarmstrong: I think we can support both paths. On one hand, yes often datasets will have imports as a data field.

On the other hand--how do you know a priori what imports a theorem should use? E.g., in benchmark setting, is it always appropriate to give known headers that vary on a per-theorem basis?  Perhaps it is more appropriate to apply the same header globally.

I think both paths are valid--I will write something to also expose the headers from data field.

nemo_skills/inference/model/code_execution.py
gwarmstrong: Agree it could be useful in the future. I think the scope is outside of this PR because it interacts substantially with prompting, training and templating.

nemo_skills/inference/model/base.py
gwarmstrong: Removed

nemo_skills/inference/model/code_execution.py
gwarmstrong: On closer inspection, getting entry-specific headers would require modifying a lot of generation modules to accept an additional parameter that only applies to very limited scenarios, which therefore requires some additional care that I believe is outside the scope of this PR. Alternatively, the header can be passed by adding it to the prompt and later parsed upon code extraction. However, this will be prompt specific, so I encourage you to implement it for your specific needs, then perhaps we can generalize.

PR #615, opened 2025-07-25
Add pypy3 code execution

aleksficek:
In this MR I add the following:
- [x] Option to perform pypy3 code execution (specified with `language="python"`)
- [x] Added ability to pass `std_input` as part of pypy3 integration
- [x] Added and modified tests to support implementation
- [x] Added setting the sandbox container memory limit (at high loads this was taking down cluster nodes)
- [x] Stress tested the two approaches locally

**IPython vs pypy3 performance**
Experiment: 1000 requests with 50 parallel threads
Total Duration: 51.90 seconds vs 16.94 seconds 
Requests per second: 19.27 vs 59.03
Average duration per request: 0.052 seconds vs 0.017 seconds

### DIFF
--- a/nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
+++ b/nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
@@ -17,10 +17,7 @@
 # NOTE: needs to run from the root of the repo!
 
 SANDBOX_NAME=${1:-'local-sandbox'}
-SANDBOX_MEMORY_LIMIT=${2:-'16g'}
 
-docker build --tag=${SANDBOX_NAME} --build-arg="UWSGI_PROCESSES=$((`nproc --all` * 10))" --build-arg="UWSGI_CHEAPER=`nproc --all`" -f dockerfiles/Dockerfile.sandbox .
+docker build --tag=${SANDBOX_NAME} --build-arg="UWSGI_PROCESSES=$((nproc --all * 10))" --build-arg="UWSGI_CHEAPER=nproc --all" -f dockerfiles/Dockerfile.sandbox .
 
-docker run --network=host --rm --memory=${SANDBOX_MEMORY_LIMIT} --restart unless-stopped --name=local-sandbox \
-    ${NEMO_SKILLS_SANDBOX_MEM_LIMIT:+-e NEMO_SKILLS_SANDBOX_MEM_LIMIT=${NEMO_SKILLS_SANDBOX_MEM_LIMIT}} \
-    ${SANDBOX_NAME}
+docker run --network=host --rm --memory=${NEMO_SKILLS_SANDBOX_MEM_LIMIT:-"16g"} --restart unless-stopped --name=local-sandbox ${SANDBOX_NAME}


nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -40,6 +52,30 @@ def execute_python(generated_code, timeout):
 
     return queue.get()
 
+def execute_python(generated_code, std_input, timeout):
```
titu1994: I don't know if we need to default to pypy3 as engine. 
Dependencies may be problematic and not all libraries can work with pypy3. 

Maybe keep same python as older code and keep option to use pypy3 as a separate flask request ?

aleksficek: @wasiahmad needs pypy3 for livecodebench. I have it working as part of the container being built and with the tests but I can make it so that you specific language as "pypy3" specifically to run using it.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -20,14 +20,26 @@
...
 import sys
 import tempfile
+import signal
 from io import StringIO
 
 from flask import Flask, request
 
 app = Flask(__name__)
 
+MEM_LIMIT_BYTES = 10 * 1024 ** 3          # 10 GiB
```
Kipok: Can you accept this from `NEMO_SKILLS_SANDBOX_MEM_LIMIT` env var if it's provided? We had a request to let people customize this as there is probably not a single limit that works for any use-cases

nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
```diff
@@ -17,7 +17,8 @@
 # NOTE: needs to run from the root of the repo!
 
 SANDBOX_NAME=${1:-'local-sandbox'}
+SANDBOX_MEMORY_LIMIT=${2:-'16g'}
```
Kipok: Why do we need this part?

nemo_skills/code_execution/sandbox.py
```diff
@@ -246,6 +247,9 @@ def strip_ansi_codes(text):
     }}
 print(json.dumps(to_return))
 """
+        elif language in ["python", "pypy3", "python3"]:
+            TO_EXECUTE = generated_code
```
Kipok: Should we raise an error here if session_id is not None, similar to Lean? As current logic doesn't reuse previous code snippets, right?

tests/test_code_execution.py
```diff
@@ -36,116 +36,127 @@ def _get_sandbox(sandbox_type):
...
 
-    output, session_id = sandbox.execute_code(code)
+    output, _ = sandbox.execute_code(code, language=language)
     assert output == {'process_status': 'completed', 'stderr': '', 'stdout': '1\n2x3\n'}
-    assert session_id is not None
 
-    code = "print(2)\n15"
-    output, session_id2 = sandbox.execute_code(code, session_id=session_id)
+    code = "print(2)\nprint(15)"
+    output, _ = sandbox.execute_code(code, language=language)
```
Kipok: We actually want to keep session_id here to check that output of previous cells is not present in the current cell. So I'd recommend creating a separate tests for non-ipython languages

nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
aleksficek: The `MEM_LIMIT_BYTES` in the current container sets memory limits per code execution process. We need to also force a ceiling for the entire container and protect against:
1. Zombie processes that survive reaping
2. Many expensive concurrent requests

Without this hard limit I've had my container take down cluster nodes because the container would consume all of the nodes' memory.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
aleksficek: Yes but I think that would be better suited for the `SANDBOX_MEMORY_LIMIT` var which is passed to docker run. See response to your comment below and let me know what you think.

nemo_skills/code_execution/sandbox.py
aleksficek: Good point, added

tests/test_code_execution.py
aleksficek: Sounds good, I restored and made a separate test

nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
Kipok: Actually now that I think about this, the cluster sandbox doesn't go through this script and instead launches the container directly, so you'd need to add this variable to cluster config or set as one of the defaults inside pipeline/utils. So this should have no effect on cluster jobs and not even affect local jobs, only when you directly use this script to start the sandbox. So perhaps it's not needed after all? Or if it is needed, you'd need to pass it differently to the cluster jobs. But perhaps can be done in another pr

aleksficek: Sounds good, I can make this change in a different PR then

PR #621, opened 2025-07-28
Soft failure for BFCL

shtoshni:
Added the logic for soft failure. The model exits the conversation when it is about to run out of context. Have tested it on Qwen3-4B, and the soft fail logic worked.


Kipok (approved):
Thanks!

PR #622, opened 2025-07-29
Ipython session affinity

gwarmstrong:
# Add Ipython with Session Affinity

* Enables code execution with Nginx-based session affinity and per-session IPython backends
* Supports both persistent and stateless execution modes
* Maintains backward compatibility with single-server mode
* Includes REST APIs for session management and automated cleanup
* Validated with 240 concurrent session servers and 240 concurrent sessions on 2000 total sessions


i-vainn:
Thanks! A few comments below

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -20,13 +20,133 @@
...
+# ============================================================================
+# IPYTHON SESSION BACKEND
+# ============================================================================
+
+try:
+    from IPython.terminal.interactiveshell import TerminalInteractiveShell
+    IPYTHON_AVAILABLE = True
+except ImportError:
+    IPYTHON_AVAILABLE = False
+    print("IPython not available - session persistence will not work!")
```
i-vainn: Do we really need `IPYTHON_AVAILABLE` here? It might be better to fail explicitly if IPython isn't available.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -103,13 +223,58 @@ def execute():
...
+def health():
+    """Health check endpoint for load balancer"""
+    return {"status": "healthy", "port": os.environ.get('FLASK_PORT', '6000')}
+
+
 if __name__ == '__main__':
     log = logging.getLogger('werkzeug')
     log.setLevel(logging.WARNING)
+
+    print(f"Starting sandbox server on port {port}")
```
i-vainn: Is `port` defined anywhere?

i-vainn:
Currently `sandbox.py` still re-executes the whole code history to track the state, right?

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
gwarmstrong: Yeah you're right, probably better to remove

gwarmstrong:
> Currently `sandbox.py` still re-executes the whole code history to track the state, right?

Right--the client will need updates to actually use the session.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
gwarmstrong: good catch, this was a holdover from debugging, can be removed

dockerfiles/Dockerfile.sandbox
```diff
@@ -46,20 +46,48 @@ ENV PATH="/lean4/my_project:$PATH"
...
 COPY requirements/code_execution.txt /app/requirements.txt
 RUN pip install --no-cache-dir -r /app/requirements.txt
-COPY nemo_skills/code_execution/local_sandbox/local_sandbox_server.py /app/main.py
 
 # For scicode eval
 RUN mkdir /data && pip install gdown && \
     python -c "import gdown; url = f'https://drive.google.com/uc?id=17G_k65N_6yFFZ2O-jQH00Lh6iaw3z-AW'; gdown.download(url, '/data/test_data.h5', quiet=False)"
 
+# IPython is installed via requirements.txt
+
```
gwarmstrong:
```suggestion
```

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -20,12 +20,117 @@
...
+from IPython.terminal.interactiveshell import TerminalInteractiveShell
 
 from flask import Flask, request
 
 app = Flask(__name__)
 
+# Global dictionary to store IPython shells by session_id
+sessions = {}
+session_lock = threading.Lock()
+SESSION_TIMEOUT = 3600  # 1 hour timeout
```
Kipok: take this from env var? Should we have no timeout by default? Users are generally expected clean up sessions on the client side

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -20,12 +20,117 @@
...
+                print(f"Error cleaning up session {session_id}: {e}")
+
+def get_or_create_session(session_id):
+    """Get existing IPython session or create a new one (fast startup)"""
+    current_time = time.time()
+    with session_lock:
+        if session_id not in sessions:
+            try:
+                # Create NEW IPython shell instance for each session (not singleton!)
+                shell = TerminalInteractiveShell()  # Remove .instance() to create separate instances
```
Kipok:
```suggestion
                shell = TerminalInteractiveShell()
```

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -20,12 +20,117 @@
...
+
+        # Get or create session
+        session_data = get_or_create_session(session_id)
+        shell = session_data['shell']
+
+        # Capture stdout/stderr
+        old_stdout = sys.stdout
+        old_stderr = sys.stderr
+
+        stdout_capture = StringIO()
```
Kipok: why not reuse our previous logic exactly here? It didn't need e.g. to add result.result to the output and used Ipython's own io module to capture output. It also supported important things like traceback_verbosity which we should preserve

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -103,12 +208,55 @@ def execute():
     timeout = request.json['timeout']
     language = request.json.get('language', 'python')
 
+    # Get session_id from JSON body first, then from header (for nginx compatibility)
+    session_id = request.json.get('session_id') or request.headers.get('X-Session-ID')
```
Kipok: why do we need this or case? One http session can have multiple session_ids, I don't think we should assume that requests correspond to the same session if it's not specified

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -103,12 +208,55 @@ def execute():
...
+
     if language == 'python':
-        return execute_python(generated_code, timeout)
+        if session_id:
```
Kipok: why do we need a case with session_id not specified? If people don't want to use session, then should just pick "python" as a language which will be faster. But for ipython, I'd have session required

nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
```diff
@@ -17,7 +17,24 @@
...
 
-docker build --tag=${SANDBOX_NAME} --build-arg="UWSGI_PROCESSES=$((nproc --all * 10))" --build-arg="UWSGI_CHEAPER=nproc --all" -f dockerfiles/Dockerfile.sandbox .
+docker build --tag=${SANDBOX_NAME} --build-arg="UWSGI_PROCESSES=$((`nproc --all` * 10))" --build-arg="UWSGI_CHEAPER=nproc --all" -f dockerfiles/Dockerfile.sandbox .
 
-docker run --network=host --rm --memory=${NEMO_SKILLS_SANDBOX_MEM_LIMIT:-"16g"} --restart unless-stopped --name=local-sandbox ${SANDBOX_NAME}
+echo "Starting sandbox in $DEPLOYMENT_MODE mode..."
+
+DEPLOYMENT_ARGS=""
+
+if [ "$DEPLOYMENT_MODE" = "multi-worker" ]; then
```
Kipok: Do we need two different modes? Is there a downside in using multi-worker always?

tests/test_session_affinity.py
```diff
@@ -0,0 +1,447 @@
+#!/usr/bin/env python3
```
Kipok: license

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
gwarmstrong: sure, updated

nemo_skills/code_execution/local_sandbox/start_local_sandbox.sh
gwarmstrong: I was concerned about potential perf implications from using multi-worker. I'll do a perf test and if that turns out alright then I suppose we can remove the old deployment version.

gwarmstrong: So at the moment, if given enough resources, there is very little performance difference between the two modes. But if there code being executed consumes a lot of memory, it will tend to lead to a timeout in the multi-worker setting (which is time consuming) rather than failing quickly like the previous single deployment method. I'll see if I can tweak it to avoid this behavior, in which case, I would feel more comfortable moving to multi-worker completely.

gwarmstrong: The other difference would be the job scheduling. 'single' deployment mode has a slight edge because it can assign requests to the first available worker, effectively balancing workloads across workers. Whereas the 'multi-worker' version with session-affinity will assign a job to a replica without regard the current workload of that replica, even when there is no session-id. This could lead to some inefficiency in scheduling.

Kipok:
@i-vainn is all ready for another review or are you still doing changes?

Kipok:
looks good to me, except two things

1. Please resolve conflicts
2. @i-vainn please discuss with @gwarmstrong on whether it's ok for us to have a single deployment mode. If it's important to be able to switch, this should be supported in the slurm as well, because current this argument is not used when things are running on cluster as far as I understand as that start script isn't used there

nemo_skills/code_execution/sandbox.py
```diff
@@ -116,17 +116,17 @@ def __init__(
         )
         self.ssh_server = os.getenv("NEMO_SKILLS_SSH_SERVER", ssh_server)
         self.ssh_key_path = os.getenv("NEMO_SKILLS_SSH_KEY_PATH", ssh_key_path)
-        # will keep state of code sessions
-        self.sessions = {}
-
-    def clear_session(self, session_id):
```
Kipok, line 122: instead of removing this, should we call it inside code execution model when we are done with the current session? And change this function to actually delete session from the sandbox (there is some new api for that). This way we can avoid clogging the cpus with inactive sessions? Or does it happen like this in some other way already?

i-vainn: We have an automatic cleanup on each execute call, but this is a good point, I'll add it to the code execution model

Kipok (approved):
thanks, a lot of great work here! Let's just hope it doesn't break anything we didn't fully test :)

PR #624, opened 2025-07-29
Add faster python code execution

aleksficek:
**_This is a PR resubmitting https://github.com/NVIDIA/NeMo-Skills/pull/615_**

In this MR I add the following:
- [x] Option to perform pypy3 code execution (specified with `language="python"`)
- [x] Added ability to pass `std_input` as part of pypy3 integration
- [x] Added and modified tests to support implementation
- [x] Added setting the sandbox container memory limit (at high loads this was taking down cluster nodes)
- [x] Stress tested the two approaches locally

**IPython vs pypy3 performance**
Experiment: 1000 requests with 50 parallel threads
Total Duration: 51.90 seconds vs 16.94 seconds 
Requests per second: 19.27 vs 59.03
Average duration per request: 0.052 seconds vs 0.017 seconds


Kipok (approved):
Looks great, thanks! Will update container and rerun the ci, hopefully all tests pass after that

PR #626, opened 2025-07-30
Remove sync logic

smahdavi4:
Refactoring the generation pipeline. Removing sync logic first.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -105,9 +105,7 @@ class GenerateSolutionsConfig:
     # and so on
     multi_turn_key: str | None = None
 
-    # set to False if you want to use synchronous loop instead of async. Async loop means we will send all
-    # data to engine at the same time (batch size is ignored) and then write the output as soon as it's ready
-    # to `output_file`-async (and put it back in order after all generations are done)
+    # Must always be True, sync is no longer supported
     use_async_loop: bool = True
     async_position_key: str = "_async_position"  # key to use for preserving position in async loop in data dict
 
@@ -249,17 +247,12 @@ class GenerationTask:
 
         self.extra_stop_phrases = OmegaConf.to_container(self.cfg.extra_stop_phrases, resolve=True)
 
-        self.use_async_loop = (
-            self.cfg.use_async_loop
-            and self.cfg.server["server_type"] not in ["nemo", "megatron"]
-            and self.cfg.multi_turn_key is None
+        assert self.cfg.use_async_loop, "Sync loop is no longer supported"
+        LOG.info(
+            "Async loop is maintaining %d generations in parallel. "
+            "Use max_concurrent_requests to control the number of concurrent requests.",
+            self.cfg.max_concurrent_requests,
         )
-        if self.use_async_loop:
-            LOG.info(
-                "Async loop is maintaining %d generations in parallel. "
-                "Use max_concurrent_requests to control the number of concurrent requests.",
-                self.cfg.max_concurrent_requests,
-            )
 
     def setup_llm(self):
         # TODO: DRY with the check in the validation config
@@ -441,40 +434,6 @@ class GenerationTask:
 
         return generate_method(**generation_params)
 
-    # TODO: rewrite mtbench to have turns separated in data file and remove this method
-    def llm_generate_multi_turn(self, data_points, data):
-        # TODO: this will not be efficient if different elements have different number of turns
-        # (effective batch size gets smaller). Need to rewrite it to ensure batch size is filled
-        # no matter the turns. Also even the below implementation can probably be simplified
-        turn_data_points = deepcopy(data_points)
-        dp_indices = list(range(len(turn_data_points)))
-        cur_turn = 1
-        outputs = [{"generation": []} for _ in range(len(data_points))]
-        while dp_indices:
-            # updating the turns to only have data up-to the current turn
-            # and adding any generated assistant messages
-            for dp_index in dp_indices:
-                turn_data_points[dp_index][self.cfg.multi_turn_key] = data_points[dp_index][self.cfg.multi_turn_key][
-                    :cur_turn
-                ]
-                for turn_idx in range(cur_turn - 1):
-                    turn_data_points[dp_index][self.cfg.multi_turn_key][turn_idx]['assistant'] = outputs[dp_index][
-                        "generation"
-                    ][turn_idx]
-            # getting a new set of generations
-            turn_outputs = self.llm_generate([turn_data_points[dp_index] for dp_index in dp_indices], data)
-            # adding assistant answers to the generations
-            for pos_index, dp_index in enumerate(dp_indices):
-                outputs[dp_index]["generation"].append(turn_outputs[pos_index]["generation"])
-
-            # removing any indices that got through all turns
-            dp_indices = []
-            for dp_index, (output, dp) in enumerate(zip(outputs, data_points)):
-                if len(output["generation"]) < len(dp[self.cfg.multi_turn_key]):
-                    dp_indices.append(dp_index)
-            cur_turn += 1
-        return outputs
-
     def dump_outputs(self, outputs, data_points, fout):
         for output, original_data_point in zip(outputs, data_points):
             # to make it easier to follow up with evaluation and limit accidental errors, we are adding
@@ -510,30 +469,6 @@ class GenerationTask:
         """Prefill generation in case LLM is not required."""
         # Override this method to customize the prefilling behavior.
         return None
-
-    def sync_loop(self, data):
-        with open(self.cfg.output_file, "at", encoding="utf-8", buffering=1) as fout:
-            data_points_batch = []
-            for idx, data_point in tqdm(enumerate(data), total=len(data), desc="Remaining generations"):
-                prefill_output = self.prefill_generation(data_point)
-                if prefill_output is not None:
-                    # We can bypass the LLM and directly dump the prefilled output
-                    self.dump_outputs([prefill_output], [data_point], fout)
-                else:
-                    data_points_batch.append(data_point)
-
-                if len(data_points_batch) == self.cfg.max_concurrent_requests or idx == len(data) - 1:
-
-                    for data_point in data_points_batch:
-                        # registering current time to calculate total generation time
-                        data_point['generation_start_time'] = time.time()
-
-                    if self.cfg.multi_turn_key is None:
-                        outputs = self.llm_generate(data_points_batch, data)
-                    else:
-                        outputs = self.llm_generate_multi_turn(data_points_batch, data)
-                    self.dump_outputs(outputs, data_points_batch, fout)
-                    data_points_batch = []
 
     def get_llm_generations(self, requests_in_progress, generations):
         """Get the completed LLM generations that were submitted asynchronously."""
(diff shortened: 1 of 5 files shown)


nemo_skills/inference/generate.py
```diff
@@ -441,40 +434,6 @@ def llm_generate(self, data_points, data, is_async=False):
 
         return generate_method(**generation_params)
 
-    # TODO: rewrite mtbench to have turns separated in data file and remove this method
```
Kipok, line 444: Please also remove mt-bench dataset and all related logic as it's not going to work anymore. But we don't really need it, so it's ok to remove (and we can add it back with a different implementation later if needed)

nemo_skills/inference/generate.py
```diff
@@ -105,9 +105,7 @@ class GenerateSolutionsConfig:
     # and so on
     multi_turn_key: str | None = None
 
-    # set to False if you want to use synchronous loop instead of async. Async loop means we will send all
-    # data to engine at the same time (batch size is ignored) and then write the output as soon as it's ready
-    # to `output_file`-async (and put it back in order after all generations are done)
+    # Must always be True, sync is no longer supported
     use_async_loop: bool = True
```
Kipok: Should we remove this parameter completely then?

nemo_skills/inference/generate.py
```diff
@@ -629,10 +564,7 @@ def generate(self):
 
         data = self.load_data()
 
-        if self.use_async_loop:
-            data = self.skip_completed_samples_async(data)
-        else:
-            data = self.skip_completed_samples_sync(data)
+        data = self.skip_completed_samples_async(data)
```
Kipok: Do you think we should rename those functions as well to remove "async" part? Can do in another pr when you change the structure more substantially

nemo_skills/inference/generate.py
Kipok: And eventually we should also remove all logic related to multi-turn from prompt utils https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/prompt/utils.py#L231. And from the docs here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/basics/prompt-format.md?plain=1#L176 (instead can mention there that users can pass in `++prompt_format=openai` to bypass our prompting logic and use openai messages as an input in generation pipeline).

Not necessary to do this in this pr

smahdavi4: I removed mtbench, let's keep the multi-turn prompt for another pull request

nemo_skills/inference/generate.py
smahdavi4: yeah i'll do it in the next pr when I change the async logic to asyncio

nemo_skills/inference/generate.py
smahdavi4: Don't we need some backward compatibility for users who set this?

smahdavi4:
I added the async logic to all the task generation classes. please also take a look and let me know if it looks good. I have tested the generation itself and it works fine, but I haven't tested the bfcl, arena hards, decontamination to make sure they work. Lmk if any change is required while I test those.

nemo_skills/inference/generate.py
Kipok: Well, it defaulted to True, so it's only really meaningful to set it to False, but that's not supported anymore. So I don't really see much point in keeping this parameter as it will result in error anyway

nemo_skills/inference/generate.py
```diff
@@ -511,49 +434,57 @@ def prefill_generation(self, data_point) -> dict | None:
...
+                
+                # Thread-safe output writing
+                async with self.output_lock:
+                    self._dump_outputs([output], [data_point], fout)
+                    pbar.update(1)
+                    
+            except Exception as e:
+                LOG.error(f"Error processing data point: {e}")
+                # Still update progress bar even on error
+                pbar.update(1)
```
Kipok, line 464: why do we need to update on error?

nemo_skills/inference/model/base.py
```diff
@@ -34,6 +35,14 @@
 LOG = logging.getLogger(get_logger_name(__file__))
 
 
+# Allow asyncio to use higher thread pool size
```
Kipok: this will be removed after we change clients?

nemo_skills/inference/check_contamination.py
```diff
@@ -120,83 +121,34 @@ def prefill_generation(self, data_point):
...
-                contaminated = any(
-                    [generation.strip() == "True" for generation in generations[original_dp_idx]['all_generations']]
-                )
-                generations[original_dp_idx]['generation'] = contaminated
-
-        # Return the remaining requests in progress and the generations
-        return (requests_in_progress, generations)
+    async def process_single_datapoint(self, data_point, all_data):
+        """Process a single data point by running contamination checks on all similar items."""
+        prefill_result = self.prefill_generation(data_point)
```
Kipok: Is this needed? The parent class should already call prefill before we get here, right?

nemo_skills/inference/generate.py
```diff
@@ -422,60 +397,8 @@ def fill_prompt(self, data_point, data):
...
-            # removing any indices that got through all turns
-            dp_indices = []
-            for dp_index, (output, dp) in enumerate(zip(outputs, data_points)):
-                if len(output["generation"]) < len(dp[self.cfg.multi_turn_key]):
-                    dp_indices.append(dp_index)
-            cur_turn += 1
-        return outputs
-
-    def dump_outputs(self, outputs, data_points, fout):
+    def _dump_outputs(self, outputs, data_points, fout):
```
Kipok: I feel the current naming with and without _ is a bit inconsistent. Typically, underscore is used for public methods that can be called outside of the class. In that case, I'd say everything should start with underscore here except "generate" as no other function can really be called in isolation. So either we only have generate without underscore or we have everything without underscore (which was the original naming where we basically just don't differentiate between private and public). So we should be consistent in naming here ideally

nemo_skills/inference/generate.py
Kipok: so should we do it now then as you added asyncio changes? I'd probably remove async from all methods to keep naming simpler

nemo_skills/inference/model/base.py
smahdavi4: Assuming we move everything to native async, yes, but if other clients do not support async, we need to keep it for them (I assume nemo and megatron are one of them?).

Kipok: Well, it's anyway just an http request on the client side, so for anything that doesn't follow openai api we can always use aiohttp or something like that. But I think for now we will just support only openai-compatible bits and will see if we need to add some other support later

nemo_skills/inference/model/base.py
```diff
@@ -272,6 +273,17 @@ def generate(
 
         return all_generations
 
+    async def generate_asyncio(self, *args, **kwargs) -> dict:
+        # Configure the executor for the current event loop
+        loop = asyncio.get_running_loop()
+        if not hasattr(loop, '_nemo_skills_executor_configured'):
+            loop.set_default_executor(ThreadPoolExecutor(max_workers=1024))
```
```suggestion
            loop.set_default_executor(ThreadPoolExecutor(max_workers=2048))
```

nemo_skills/inference/generate.py
```diff
@@ -562,53 +488,30 @@ def async_loop(self, data):
...
-                self.dump_outputs(outputs_to_dump, data_points_to_dump, fout)
-                time.sleep(1)  # Prevent excessive API overload
+            # Wait for all tasks to complete
+            if tasks:
+                await asyncio.gather(*tasks, return_exceptions=True)
 
             pbar.close()
 
-        self.restore_async_order()
+        self._restore_async_order()
```
Kipok: should we rename this one back without underscore then?

nemo_skills/inference/generate.py
```diff
@@ -511,49 +431,55 @@ def prefill_generation(self, data_point) -> dict | None:
...
+            try:
+                # Generate output for this single data point
+                output = await self.process_single_datapoint(data_point, all_data)
+                
+                # Thread-safe output writing
+                async with self.output_lock:
+                    self.dump_outputs([output], [data_point], fout)
+                    pbar.update(1)
+                    
+            except Exception as e:
```
Kipok: does the exception not get printed otherwise? If it's already part of the stacktrace, do we need another LOG.error or maybe just let it fail without try-except?

nemo_skills/inference/model/base.py
```diff
@@ -272,6 +273,17 @@ def generate(
...
         return all_generations
 
+    async def generate_asyncio(self, *args, **kwargs) -> dict:
+        # Configure the executor for the current event loop
+        loop = asyncio.get_running_loop()
+        if not hasattr(loop, '_nemo_skills_executor_configured'):
+            loop.set_default_executor(ThreadPoolExecutor(max_workers=2048))
+            loop._nemo_skills_executor_configured = True
+
+        result = await asyncio.to_thread(self.generate, *args, **kwargs)
```
fchen97, line 283: Hi Sadegh,

I noticed that you’re still calling self.generate, which uses self.generate_async and the executor within the model. I’m wondering if it might be simpler to call self._generate_single directly and manage threading only at the loop level.

Additionally, I sometimes need to limit the number of max_workers during code compilation. It would be helpful if we could make max_workers configurable.

nemo_skills/inference/model/code_execution.py
```diff
@@ -460,6 +464,11 @@ def generate(
 
         return all_generations
 
+    async def generate_asyncio(self, *args, **kwargs) -> dict:
+        result = await asyncio.to_thread(self.generate, *args, **kwargs)
```
fchen97, line 468: Similarly, here I feel calling `self._generate_single` instead of `self.generate` may be preferred.

fchen97 (commented):
I think adding a `self._generate_single` wrapper function to the BaseModel class could be helpful for scenarios where I want to manage threading outside the BaseModel itself. Below is a function I propose to add to the BaseModel class. And you should call `self._generate_single_completion` instead of `self.generate` for single completion.

```
def _generate_single_completion(
        self,
        prompt: str,
        tokens_to_generate: int = 2048,
        temperature: float = 0.0,
        top_p: float = 0.95,
        top_k: int = 0,
        min_p: float = 0.0,
        repetition_penalty: float = 1.0,
        random_seed: int = 0,
        stop_phrases: list[str] | None = None,
        top_logprobs: int | None = None,
        timeout: int | None = None,
        remove_stop_phrases: bool = True,
        stream: bool = False,
        reasoning_effort: str | None = None,
        tools: list[dict] | None = None,
        include_response: bool = False,
    ) -> list[dict]:
        """Same as generate_async but for a single completion"""
        request = {
            'prompt': prompt,
            'tokens_to_generate': tokens_to_generate,
            'temperature': temperature,
            'top_p': top_p,
            'top_k': top_k,
            'min_p': min_p,
            'repetition_penalty': repetition_penalty,
            'random_seed': random_seed,
            'stop_phrases': stop_phrases,
            'top_logprobs': top_logprobs,
            'timeout': timeout,
            'stream': stream,
            'reasoning_effort': reasoning_effort,
            'include_response': include_response,
        }

        if tools is not None:
            request['tools'] = tools

        self.preprocess_request(request)

        return self._generate_single(**request)
```

nemo_skills/inference/model/code_execution.py
smahdavi4: Hi, this is a temporary fix to make the refactoring managable, while not breaking any code. I am refactoring the client side, and adding native async support. I think the number of workers should be determined by the `max_concurrent_requests` in the GenerateTask side, and not the client side. I'll make sure to add a sync option as well.

PR #627, opened 2025-07-30
Fix cannot have large concurrency 

fzyzcjy:
Before this change: I only observed ~100 requests in parallel
After fix: for gsm8k can easily make all >1000 req in parallel

I am confused why the existing parallelism does not work, and here is a quick workaround

this PR is not polished, I want to firstly know whether nemo skills is willing to accept PR / does the direction of this looks good

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# Contributing To Nemo-Skills
+
+Thanks for your interest in contributing to Nemo-Skills!
+
+## Running Tests
+
+TBD
+
+## Code Quality
+
+- Follow the existing code style and conventions
+- Write tests for new features
+- Update documentation to reflect your changes
+- Ensure all tests pass before submitting a PR
+- Do not add arbitrary defaults for configs, be as explicit as possible.
+
+
+## Signing Your Work
+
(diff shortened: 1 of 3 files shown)


Kipok:
Thanks @fzyzcjy ! We did notice this as well recently and also not sure why there is this limit of 100 requests. Maybe the httpx client doesn't get used for some reason or there is some other parameter that we need to change to bypass it. Didn't really have time to debug this properly yet.

I think we can accept this as a temporary workaround as I don't really see a downside to your approach. And we will try to find the real root issue and fix it later.

fzyzcjy:
Thanks for the reply! Then looking forward to being merged, and feel free to ping me if need minor modifications

Kipok:
Just one small thing - can you please add the same client array to AzureOpenAIModel over here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/model/azure.py#L58? 

I will run some more tests with this tomorrow to double check there are no issues, and we can merge after that

Kipok:
@fzyzcjy all looks good to me, except can you please force push your commits with "-s" flag so that they are signed-off? Just added some more details on that in https://github.com/NVIDIA/NeMo-Skills/blob/main/CONTRIBUTING.md#signing-your-work

Thanks again for your contribution!

fzyzcjy:
done

Kipok (approved):
Thanks!

PR #633, opened 2025-08-01
Refactor clients

smahdavi4:
This merge request refactors the clients to go native async as well as use the litellm client for universal openai interface.


nemo_skills/inference/model/code_execution.py
```diff
@@ -321,153 +278,27 @@ def generate_async(
...
+            if output['generation'] is not None:
+                output['generation'] = trim_after_stop_phrases(output['generation'], stop_phrases_for_request)
+        
+        return output
+
+    async def generate_asyncio(self, prompt, *args, **kwargs) -> dict:
+        # Configure the executor for the current event loop
+        loop = asyncio.get_running_loop()
+        if not hasattr(loop, '_nemo_skills_executor_configured'):
+            loop.set_default_executor(ThreadPoolExecutor(max_workers=2048))
```
fchen97: Maybe add `max_workers` to `CodeExecutionConfig`.

And do

```suggestion
            loop.set_default_executor(ThreadPoolExecutor(max_workers=self.config.max_workers))
```

fchen97 (commented):
Thanks! Just wonder if you could set `max_workers` configurable in CodeExecutionWrapper.

nemo_skills/inference/model/code_execution.py
smahdavi4: Would it be beneficial since we already have `max_concurrent_requests` in the GenerationTask? Are there any usecases that `max_concurrent_requests` does not serve?

fchen97: Oh, I just want it to be configurable. `max_concurrent_requests` should work as well.

nemo_skills/inference/generate.py
```diff
@@ -259,7 +259,7 @@ def __init__(self, cfg: GenerateSolutionsConfig):
 
     def setup_llm(self):
         # TODO: DRY with the check in the validation config
-        if self.cfg.prompt_template is None and self.cfg.server["server_type"] in ["nemo", "megatron"]:
+        if self.cfg.server["server_type"] in ["nemo", "megatron"]:
```
Kipok: I think this if can now be removed fully

nemo_skills/inference/model/__init__.py
```diff
@@ -17,38 +17,30 @@
...
 # Utilities
-from .utils import RequestException, trim_after_stop_phrases
-from .vllm import VLLMModel, VLLMRewardModel
+from .vllm import VLLMModel
 
 # Model registry
 models = {
-    'trtllm': TRTLLMModel,
+    'trtllm': VLLMModel,
     'trtllm-serve': VLLMModel,
```
Kipok: Remove this one as it's no longer a separate option

nemo_skills/inference/model/base.py
```diff
@@ -13,21 +13,17 @@
...
-from collections.abc import Generator
-from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Dict, Optional, Union
-
 import httpx
-import openai
-import requests
-from openai import DefaultHttpxClient, Stream
+
+# TODO: Remove this once added to the docker image
```
Kipok: Please remove, I updated the image

nemo_skills/inference/model/base.py
```diff
@@ -67,174 +71,142 @@ def __init__(
...
-        top_logprobs: int | None = None,
-        timeout: int | None = None,
-        stream: bool = False,
-        reasoning_effort: str | list[int] | None = None,
-        tools: list[dict] | None = None,
-        include_response: bool = False,
-        extra_body: dict = None,
-    ) -> dict:
-        """If the engine supports inflight-batching of requests, you only need to define this method.
+        assert model is not None, "model is required"
```
Kipok: If model is required, let's just make it a required parameter without default?

nemo_skills/inference/model/base.py
```diff
@@ -67,174 +71,142 @@ def __init__(
...
+        litellm.aclient_session = httpx.AsyncClient(limits=httpx_limits)
 
-        We will call it in threads on the list of prompts.
-        """
+
+    def __del__(self):
+        if self._tunnel:
+            self._tunnel.stop()
+
+    def _maybe_apply_stop_phrase_removal(self, result: dict, remove_stop_phrases: bool, stop_phrases: list[str] | list[list[str]] | None) -> None:
```
```suggestion
    def _maybe_apply_stop_phrase_removal(self, result: dict, remove_stop_phrases: bool, stop_phrases: list[str] | None) -> None:
```

nemo_skills/inference/model/base.py
```diff
@@ -67,174 +71,142 @@ def __init__(
...
+        if remove_stop_phrases and isinstance(result, dict) and result.get('generation') is not None:
```
```suggestion
        if remove_stop_phrases:
```
if we were asked explicitly to remove, then we shouldn't silently not do that if something is misformed, instead better to just fail with error

nemo_skills/inference/model/base.py
```diff
@@ -67,174 +71,142 @@ def __init__(
...
-        if request["temperature"] == 0:
-            request["temperature"] = 1.0
-            request["top_k"] = 1
-            request["top_p"] = 1.0
+    @abc.abstractmethod
+    def _build_completion_request_params(self, **kwargs) -> dict:
+        pass
 
-    def generate_async(
+    def _prepare_generation_params(
```
Kipok: Can we fully remove this method? Let's just set timeout to 10000 as default wherever it's defined and same for top_k, just change default to -1

nemo_skills/inference/model/base.py
```diff
@@ -67,174 +71,142 @@ def __init__(
...
+        """Prepare generation parameters for both sync and async methods."""
+        if kwargs['timeout'] is None:
+            # litellm uses 10min as default None
+            # So we change it to 10000 seconds for consistency with other clients
+            kwargs['timeout'] = 10000 
+        if kwargs['top_k'] == 0:
+            kwargs['top_k'] = -1 # litellm doesn't support top_k=0
+        return kwargs
+
+    async def generate_asyncio(
```
Kipok: maybe just generate_async as a name? But up to you if you think this is better

nemo_skills/inference/model/base.py
```diff
@@ -244,265 +216,72 @@ def generate(
...
-    """
-    Base class for models using an OpenAI-compatible API.
-    Handles client setup, SSH tunneling, and a unified generation flow with generation tracking.
-    """
+        result = self._generate_single_sync(**request)
+        self._maybe_apply_stop_phrase_removal(result, remove_stop_phrases, stop_phrases)
+        return result
 
-    def __init__(
+    async def _generate_single_async(
```
Kipok: maybe call it just _generate_async? Single is always the case now and public functions are also single, so feels a bit redundant here

Kipok: also what's the value of this as a separate function if it's only ever called inside generate_async? Why not put all of this logic there directly and have one fewer function?

nemo_skills/inference/model/base.py
```diff
@@ -244,265 +216,72 @@ def generate(
...
+            if stream:
+                result = self._stream_completion_chunks_async(response)
+            else:
+                result = self._parse_completion_response(response, include_response=include_response, **kwargs)
+        else:
+            raise TypeError(f"Unsupported prompt type: {type(prompt)}")
+        return result
 
-    def _generate_single(
+    def _generate_single_sync(
```
Kipok: same comments here as for async case

nemo_skills/inference/model/code_execution.py
```diff
@@ -321,153 +277,26 @@ def generate_async(
...
+        if remove_stop_phrases:
+            stop_phrases_for_request = request['stop_phrases']
+            if output['generation'] is not None:
+                output['generation'] = trim_after_stop_phrases(output['generation'], stop_phrases_for_request)
+        
+        return output
+
+    async def generate_asyncio(self, prompt, *args, **kwargs) -> dict:
+        # Configure the executor for the current event loop
+        loop = asyncio.get_running_loop()
```
Kipok: do we need this logic at all? Shouldn't it now be possible to have a proper async interface here, by just calling the async function from self.model? Should hopefully be just a few lines change to the current implementation

nemo_skills/inference/model/vllm.py
```diff
@@ -116,94 +115,8 @@ def _build_chat_request_params(
             "stream": stream,
             "timeout": timeout,
             "extra_body": self._build_request_body(top_k, min_p, repetition_penalty, extra_body=extra_body),
+            "toold": tools,
```
Kipok:
```suggestion
            "tools": tools,
```

nemo_skills/pipeline/utils/server.py
```diff
@@ -275,16 +202,7 @@ def get_server_command(
...
-        server_entrypoint = server_entrypoint or "nemo_skills.inference.server.serve_trt"
-        server_start_cmd = (
-            f"FORCE_NCCL_ALL_REDUCE_STRATEGY=1 python -m {server_entrypoint} "
-            f"    --model_path {model_path} "
-            f"    --port {server_port} "
-            f"    {server_args} "
-        )
-        num_tasks = num_gpus
-    elif server_type == 'trtllm-serve':
+    elif server_type in ['trtllm-serve', 'trtllm']:
```
Kipok:
```suggestion
    elif server_type in ['trtllm']:
```
let's just use trtllm now instead of trtllm-serve

Kipok: please also remove from all other places where "trtllm-serve" is used and use "trtllm" instead

requirements/main.txt
```diff
@@ -39,3 +39,5 @@ tqdm
 transformers
 typer >= 0.13
 wandb
+litellm==1.71.1
```
Kipok:
```suggestion
litellm
```
better not pin version unless we have a good reason to

nemo_skills/inference/model/code_execution.py
smahdavi4: the code execution sandbox does not support async for now. so make it native async would make it wait for code execution to finish before executing the next event.

nemo_skills/pipeline/utils/server.py
```diff
@@ -276,17 +201,8 @@ def get_server_command(
...
-            f"FORCE_NCCL_ALL_REDUCE_STRATEGY=1 python -m {server_entrypoint} "
-            f"    --model_path {model_path} "
-            f"    --port {server_port} "
-            f"    {server_args} "
-        )
-        num_tasks = num_gpus
-    elif server_type == 'trtllm-serve':
-        server_entrypoint = server_entrypoint or "trtllm-serve"
-        if num_nodes > 1 and server_entrypoint == "trtllm-serve":
+        server_entrypoint = server_entrypoint or "trtllm"
```
Kipok: this has to still be trtllm-serve as that's what the entrypoint actually is

nemo_skills/code_execution/sandbox.py
```diff
@@ -177,27 +124,36 @@ def __init__(
...
-
-            sshtunnel_request = sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path)
-            output = sshtunnel_request.post(
-                url=self._get_execute_url(),
-                data=json.dumps(request),
-                timeout=timeout,
-                headers={"Content-Type": "application/json"},
-            )
+            
+            def ssh_request():
```
Kipok, line 136: can we just use the same logic of doing tunnel as in the base model? That should work with async, right?

smahdavi4: it should work, but do we want to initialize two tunnels? one inside the model one inside the sandbox? would that work?

Kipok: not sure, would need to test. We can leave this for later

PR #641, opened 2025-08-05
Adding IFBench to NeMo-Skills

Davood-M:
Note that to test this you need to re-create the NeMo-Skills container using the updated `Dockerfile.nemo_skills`.
After running the IFBench evaluation, the metrics file looks similar to IFEval.


Kipok (approved):
Looks great, thanks!

PR #644, opened 2025-08-06
Fix timeout bug for LEAN4 code execution

fchen97:
- The default Nginx timeout is 60 seconds, which is insufficient for long-running processes. Updated the Nginx configuration to set the timeout to 4 hours (14,400 seconds).  
- Bug: After a timeout occurs, the LEAN4 subprocess continued running in the background.    
  Fix: On timeout during LEAN4 code execution, the entire process tree is now forcefully terminated.  

Code has been tested on both local environment and cluster environment.

Docker image needs to be rebuilt to incorporate these changes.

Help from @gwarmstrong.


fchen97:
Going to redo the pull request from branch

PR #646, opened 2025-08-07
AAI update

shtoshni:
Tested IFBench. We're able to reproduce the numbers precisely. 
Need to add AA-LCR in a separate PR.


Kipok (commented):
please also change scoring function accordingly https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/aai/aai_score.py

Kipok (approved):

PR #652, opened 2025-08-07
Improve client

smahdavi4:
* Adding reasoning effort to vllm models (particularly targetting gpt-oss)
* Make random_seed None by default (to encourage randomness for people not using it)
* Disallow more parameters in text completion


Kipok (approved):
Thanks!

PR #655, opened 2025-08-08
Online GenSelect

shtoshni:
This is an online version of GenSelect where the model explores parallel generation, followed by running a GenSelect run.


nemo_skills/inference/model/online_genselect.py
```diff
@@ -0,0 +1,178 @@
...
+        self,
+        prompt: Union[str, List],
+        random_seed: int = 0,
+        **solution_kwargs,
+    ) -> Dict:
+        """
+        Generate multiple solutions and use Self-GenSelect to choose the best one.
+        """
+        # Step 1: Generate multiple solutions
+        random.seed(random_seed)
```
smahdavi4: I think we need a stateful generator instance per async call, otherwise each instance will override the random seed in the middle of generations of others.

shtoshni: Good point. Created a local random instance. Let me know if this looks okay.

smahdavi4: looks good, thanks!

smahdavi4 (approved):

PR #656, opened 2025-08-08
Fixes for code execution

Kipok:


nemo_skills/inference/model/code_execution.py
```diff
@@ -142,7 +142,7 @@ async def _generate_single(
                 if request['timeout'] <= 0:
                     break
 
-            output_dict = self.model.generate_sync(**request, remove_stop_phrases=False)
+            output_dict = await self.model.generate_async(**request, remove_stop_phrases=False)
```
Kipok, line 145: @smahdavi4 this should be await, right? Just double checking this is indeed a typo and there is no reason this can't be async

smahdavi4: yeah my bad, i missed changing this part.

PR #661, opened 2025-08-11
Fix a small bug in generation with chunks

fchen97:
When generating with large number of chunks, the command may exceed maximum length. Fix the bug to batch the commands.


Kipok (approved):
Thanks!

PR #666, opened 2025-08-13
Allow generation type or custom module in `eval` pipeline

activatedgeek:
This PR introduces two new CLI arguments for `ns eval`.

```shell
│    --generation_type                                          [generate|reward|math_judge|check_contamination]  Type of generation to perform [default: None]         │
│    --generation_module                                        TEXT                                              Path to the generation module to use. If not          │
│                                                                                                                 specified, will use the registered generation module  │
│                                                                                                                 for the generation type (which is required in this    │
│                                                                                                                 case).                                                │
│ 
```

### DIFF
--- a/nemo_skills/inference/factory.py
+++ b/nemo_skills/inference/factory.py
@@ -1,0 +1,15 @@
+from enum import Enum
+
+
+class GenerationType(str, Enum):
+    generate = "generate"
+    reward = "reward"
+    math_judge = "math_judge"
+    check_contamination = "check_contamination"
+
+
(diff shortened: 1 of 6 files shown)


nemo_skills/inference/factory.py
```diff
@@ -0,0 +1,15 @@
+from enum import Enum
+
+
+class GenerationType(str, Enum):
+    generate = "generate"
+    reward = "reward"
```
Kipok:
```suggestion
```
this should've been removed in one of previous prs, let's remove now

PR #671, opened 2025-08-14
Add SWE-bench inference & evaluation

ludwig-n:


Kipok (approved):
Thanks!

Kipok:
@ludwig-n when you get a chance, can you please update description of this PR with an example eval command and expected score? We will eventually port it to the evaluation docs we are going to have at some point in the near future

PR #673, opened 2025-08-15
Remove prompt template

Kipok:
A bunch of breaking changes
1. For generation / eval ++prompt_template=X is no longer supported. In 99% of cases you didn't need it anyway, so can be safely removed. If you know that for some reason you did need it, use `++use_completions_api=True` instead. It will use a template from model's tokenizer. You can override with `++tokenizer=<name or path>`.
2. For sft data prep, `hf_model_name` is renamed to `tokenizer` for consistency with inference.
3. Any instances of logic like this `elem['output'] += prompt.config.template.assistant_end` should be replaced with this `elem['output'] = prompt.add_assistant_end_suffix(elem['output'])`
4. All empty system messages are removed from prompt configs. Now by default we use whatever is defined in jinja. If you want to override with empty system message, set it explicitly to `""`. It doesn't work for all models, e.g. llama3.1 doesn't allow full override, so this is no longer possible in latest code. But we don't really use those models either, so I thought it's ok. We can bring back this logic if we need it in the future
5. `extra_stop_phrases` is replaced with a single `stop_phrase`. We didn't have any use-cases where multiple were required
6. `multi_turn_key` is removed
7. `prefix_generation_to_response` and `continue_prefix_generation` are replaced with a simpler `start_assistant_response_key` (just starts assistant response with data from that key)


docs/basics/inference.md
```diff
@@ -36,32 +36,27 @@ Click on :material-plus-circle: symbols in the snippet below to learn more detai
     from nemo_skills.prompt.utils import get_prompt
 
     llm = get_model(model="meta-llama/Llama-3.1-8B-Instruct", server_type="vllm")  # localhost by default
-    prompt = get_prompt('generic/default', 'llama3-instruct') # (1)!
-    prompt = prompt.fill({'question': "What's 2 + 2?"})
+    prompt_obj = get_prompt('generic/default') # (1)!
+    prompt = prompt_obj.fill({'question': "What's 2 + 2?"})
     print(prompt) # (2)!
-    output = llm.generate_sync(prompt=prompt)
+    output = llm.generate(prompt=prompt)
```
smahdavi4: `llm.generate_sync`?

docs/basics/inference.md
```diff
@@ -124,23 +118,22 @@ Click on :material-plus-circle: symbols in the snippet below to learn more detai
...
     ```
 
-    1.   Here we use [generic/default](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt/config/generic/default.yaml) config
-         and [llama3-instruct](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt/template/llama3-instruct.yaml) template.
+    1.   Here we use [generic/default](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt/config/generic/default.yaml) config.
 
          Note how we are updating system message on the next line (you can also include it in the config directly).
 
-         See [nemo_skills/prompt](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt) for more config/template options
+         See [nemo_skills/prompt](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt) for more config options
```
shtoshni:
```suggestion
         See [nemo_skills/prompt/config](https://github.com/NVIDIA/NeMo-Skills/tree/main/nemo_skills/prompt/config) for more config options
```

docs/releases/openmathreasoning/training.md
```diff
@@ -258,7 +259,7 @@ def apply_format(elem, prompt, is_tir):
         elem['input'] = prompt.fill({'problem': elem['input'], 'total_code_executions': total_code_executions})
     else:
         elem['input'] = prompt.fill({'problem': elem['input']})
-    elem['output'] += prompt.config.template.assistant_end
+    elem['output'] += prompt.add_assistant_end_suffix(elem['output'])
```
shtoshni:
```suggestion
    elem['output'] = prompt.add_assistant_end_suffix(elem['output'])
```

nemo_skills/dataset/ruler/prepare.py
```diff
@@ -167,6 +168,18 @@ def prepare_task(task):
...
+    parser.add_argument(
+        "--max_seq_length",
+        type=int,
+        required=True,
+        help="Sequence length to check with RULER.",
+    )
+    parser.add_argument(
+        '--template_tokens',
+        type=int,
+        default=10,
```
shtoshni, line 180: unrelated to this PR, but just to be on the safe side, the default can be kept higher. Keeping it to 100ish tokens won't make a difference if your document context is 128K+

Kipok: yeah, we can tweak it if we find cases when 10 tokens isn't enough. I will do another PR with ruler specific fixes as current logic is still a bit broken

PR #676, opened 2025-08-15
Reproducing Llama Nemotron Results with NeMo-Skills

shtoshni:


docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
+---
+date: 2025-08-15
+readtime: 15 
+---
+
+
+# Reproducing Llama-Nemotron-Super-49B-V1.5 Evals 
+
+In this tutorial, we will reproduce the evals for the Llama-3.3-Nemotron-Super-49B-v1.5 model using NeMo-Skills.
+For an introduction to the NeMo-Skills framework, we recommend going over [our introductory tutorial](./omr-simple-recipe.md).
```
Kipok: Maybe getting started page is a better link here?

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+
+
+We assume you have `/workspace` defined in your [cluster config](../../basics/cluster-configs.md) and are
```
Kipok, line 13: Let's instead say that we recommend defining `HF_HOME` in your cluster config to some mounted directory (and maybe give an example like in getting started). Then we don't need to ask to download the model and can just reference it with `nvidia/Llama-3_3-Nemotron-Super-49B-v1_5`, which is probably more convenient for users

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+
+
+## Evaluation commands
+
+Llama-3.3-Nemotron-Super-49B-v1.5 can perform inference in both reasoning on and off modes. 
+We detail the evaluation commands and results for both the modes. 
+Note that you might not get exactly the same numbers as reported here because of the stochastic nature of LLM generations. 
+
+!!! note 
+    The commands provided here assume you're working with a local machine where benchmarks/subsets are evaluated sequentially which will take a very long time. If running on slurm, you can set `--num_jobs` to a bigger number or just set it to -1 to run each benchmark and their random seeds as an independent job which in case of Llama-Nemotron-Super-49B-V1.5 requires one node per job.
```
Kipok: Let's remove num_jobs from all commands, that way it will default to 1 locally and -1 on slurm, which provides the best default behavior. Also is it true that you need 8 GPUs for this model? It's supposed to fit into a single H100. Maybe add a note about that and use 1 (or 2) by default?

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+    ++inference.tokens_to_generate=65536 \
+    ++inference.temperature=0.6 \
+    ++inference.top_p=0.95 \
+    ++system_message=''
+```
+
+#### Command for HLE Eval (Reasoning on)
+
+
+For HLE, because symbolic comparison is not sufficient to determine the correctness of the output, we use the recommended `o3-mini-20250131` model as the judge. Note that this model is the default in NeMo-Skills, and we have just added this argument for illustration purposes. To evaluate for the [Artificial Analysis Index (AAI) setting, please use the gpt-4o-20240806 model as the judge](https://artificialanalysis.ai/methodology/intelligence-benchmarking#intelligence-index-evaluation-suite-overview){target="_blank"}.
```
Kipok: Add a note that OPENAI_API_KEY is required. Maybe we also add a short explanation how to self-host the judge model to promote that usage?

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+    --server_args="--tool-parser-plugin \"/workspace/Llama-3_3-Nemotron-Super-49B-v1_5/llama_nemotron_toolcall_parser_no_streaming.py\" \
+                    --tool-call-parser \"llama_nemotron_json\" \
+                    --enable-auto-tool-choice"
+```
+
+
+
+### Reasoning-on Results
+
+We use the `summarize_results` pipeline to calculate the evaluation metrics, for all but BFCL where the metrics are calculated as part of the evaluation job itself.
```
Kipok: This should automatically run now, so maybe just say that after evaluation jobs are done, the following results will be reported? Or we can add that to get this printout again, they can run an explicit summarize_results command

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+
+---------------------------------------- mmlu-pro ---------------------------------------
+evaluation_mode   | num_entries | avg_tokens | gen_seconds | symbolic_correct | no_answer
+pass@1[avg-of-16] | 12032       | 4879       | 12516       | 81.44%           | 0.05%
+majority@16       | 12032       | 4879       | 12516       | 83.05%           | 0.00%
+pass@16           | 12032       | 4879       | 12516       | 91.32%           | 0.00%
+
+-------------------------------------------------- hle --------------------------------------------------
+evaluation_mode   | num_entries | avg_tokens | gen_seconds | judge_correct | symbolic_correct | no_answer
+pass@1[avg-of-15] | 2158        | 12111      | 7782        | 7.75%         | 2.40%            | 64.13%
```
Kipok: change to 16? Can just update value to not confuse readers, no need to rerun as it's not going to change much

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+majority@15       | 2158        | 12111      | 7782        | 4.31%         | 3.43%            | 49.91%
+pass@15           | 2158        | 12111      | 7782        | 27.80%        | 10.10%           | 49.91%
+```
+
+!!!note
+    The `majority` metric for most reasoning benchmarks typically improves over the corresponding `pass@1` numbers. For HLE, the `majority` number is lower than `pass@1` which can be counterintuitive but it has to with our metric calculation logic. For HLE, the final answer is contained in the generated solution but it is not easily extractable by rule-based systems as in the case of math where the model is instructed to put the final answer in \boxed{}. Thus, for certain questions the `predicted_answer` field is null but the LLM-as-a-judge is still able to evaluate the generated solution. The majority metric performs clustering over `predicted_answer` which currently incorrectly removes from consideration some of the correct solutions for which the `predicted_answer` is None.
```
Kipok, line 203:
```suggestion
    The `majority` metric for most reasoning benchmarks typically improves over the corresponding `pass@1` numbers. For HLE, the `majority` number is lower than `pass@1` which can be counterintuitive. This is due to our metric calculation logic. For HLE, the final answer is contained in the generated solution but it is not easily extractable by rule-based systems as in the case of math where the model is instructed to put the final answer in \boxed{}. Thus, for certain questions the `predicted_answer` field is null but the LLM-as-a-judge is still able to evaluate the generated solution. The majority metric performs clustering over `predicted_answer` which currently incorrectly removes from consideration some of the correct solutions for which the `predicted_answer` is None.    
```

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -0,0 +1,441 @@
...
+| overall_live                | 2251        | 83.34%   |
+| live_ast                    | 1351        | 82.68%   |
+| live_irrelevance            | 882         | 84.47%   |
+| live_relevance              | 18          | 77.78%   |
+| overall_multi_turn          | 800         | 46.38%   |
+
+```
+
+!!! note
+    Currently `summarize_results` doesn't support benchmarks like BFCL v3 which have their specific logic of combining subset scores to arrive at the overall score. This table was created by formatting the `metrics.json` file from `/workspace/llama_nemotron_49b_1_5_tool_calling/bfcl_v3/metrics.json` using an LLM.
```
Kipok:
```suggestion
    Currently `summarize_results` doesn't support benchmarks like BFCL v3 which have their specific logic of combining subset scores to arrive at the overall score. This table was created by formatting the `metrics.json` file from `/workspace/llama_nemotron_49b_1_5_tool_calling/bfcl_v3/metrics.json`.  
```

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
Kipok: Or I guess then tool-calling isn't going to work? Maybe then let's keep your current way but add a comment that we are downloading model explicitly to get tool-parser script

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
shtoshni: Yes, it automatically runs. Let me just clarify that this command is merely for printing.

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
shtoshni: Added a note to clarify this.

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
shtoshni: Okay, with 1 GPU, I ran into OOMs. Not the case with 2 GPUs. So let's just use 2.

Kipok (approved):
Should we list this in the "News" section in readme? Otherwise looks good!

PR #677, opened 2025-08-15
Add SFT data translation module

shuoyangd:
This PR adds an extra translation module that is functionally equivalent to the codebase that was used to translate the Nemotron-Post-Training-Dataset-v2 (data release coming soon).

How to use:

```
ns generate --cluster=local --server_type=vllm --model=Qwen/Qwen2.5-14B-Instruct --server_gpus=1 --input_file=[input_jsonl_file] --output_dir=[output_dir] --generation_module=nemo_skills.inference.translate_jsonl -- target_lang=[iso_639_1_language_code] prompt_config=/path/to/nemo_skills/prompt/config/translation/qwen25.yaml
```

### DIFF
--- a/nemo_skills/inference/translate_jsonl.py
+++ b/nemo_skills/inference/translate_jsonl.py
@@ -1,0 +1,431 @@
+import logging
+from dataclasses import dataclass, field, asdict, is_dataclass
+from pathlib import Path
+from typing import List
+
+import hydra
+
+from nemo_skills.inference.generate import (
+    GenerateSolutionsConfig,
+    GenerationTask,
+    InferenceConfig,
+)
+from nemo_skills.utils import setup_logging, nested_dataclass
+
+LOG = logging.getLogger(__name__)
+MAX_CONTEXT_LENGTH = 500
+
+
+def is_line_translatable_content(line: str) -> bool:
+    """Determines if the content of a line is considered translatable."""
+    stripped_line = line.strip()
+    if not any(char.isalpha() for char in stripped_line):
+        return False
+    if stripped_line.startswith('<') and stripped_line.endswith('>'):
+        return False
+    return True
+
+
+def _get_all_nested_fields(obj, path: str) -> List[str]:
+    """A helper function that finds all string values for a field specified by `path`."""
+    parts = path.split(".")
+
+    def finder(current_obj, current_path, collected_values):
+        if not current_path:
+            return
+
+        key = current_path[0]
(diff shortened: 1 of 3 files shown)


nemo_skills/inference/translate_jsonl.py
```diff
@@ -0,0 +1,431 @@
...
+                    'target_lang': full_language_name(self.cfg.target_lang),
+                    'src': 'Sample text to translate',
+                }
+            )
+            filled_prompt = self.fill_prompt(sample_data_point, data)
+            LOG.info("Example prompt:\nData dictionary: %s\nPrompt: %s", sample_data_point, filled_prompt)
+        else:
+            LOG.info("No data available for example prompt")
+
+    async def process_single_datapoint(self, data_point, all_data):
```
Kipok: is the only difference from parent function to handle dict value? If that's the case, let's just update it in parent class and remove this method?

nemo_skills/inference/translate_jsonl.py
```diff
@@ -0,0 +1,431 @@
...
+            pass  # do nothing
+
+        return text
+
+    def postprocess(self):
+        """
+        Reassemble individual translation results back into original document structure.
+        Reads the output file written by base class, reconstructs documents, and writes back.
+        """
+        import json
```
Kipok: what's the benefit of putting all imports inside methods? I'd rather put all on top, the only reason when you need to put them inside is if something isn't globally installed, but that's not the case here and in most other places in this file

nemo_skills/inference/translate_jsonl.py
shuoyangd: Moved to parent method

nemo_skills/inference/translate_jsonl.py
shuoyangd: Moved imports out of functions

shuoyangd:
@shtoshni Good catch, likely because I created from a fork instead of a branch (per [here](https://github.com/NVIDIA/NeMo-Skills/blob/2263a6d9fa4f92459b1a2b157b4b5ea378fbb07e/CONTRIBUTING.md?plain=1#L91))?

If so lmk if I should re-create the PR or run the test in some other way.

shtoshni:
@shuoyangd - Yes, please create a branch because running this test requires the secret key!

shuoyangd:
Closing this one to proceed with #721

PR #683, opened 2025-08-18
Suggestion: add more flexibility to the endpoint definition

Glorf:
Change 1: allow use of api key in all kind of servers, not only for the specific base_url (eg for NVCF invocations or localhost)
Change 2: use a single env variable regardless of the base url
Change 3: trim the chat/completions part if provided to create base url


Kipok:
@Glorf got it, thanks for explaining this more.

> Anyways, perhaps a flag which provides name of the env_var storing the key and which is then passed to the openai client would make sense?

I think this is a good option, so let's just add a new `api_key_env_var` parameter or something like that to the base model. If it's specified, it's using that env var, if not, following the current logic.

Also can you please re-create the PR from a branch, not fork, so that all tests can run? Just sent you an invite

Glorf:
Ok, will do!
Closing

PR #685, opened 2025-08-18
fix data cache bug

wedu-nvidia:
Fixed the caching bug where different dataset files in the same folder would always reuse the same cache. Now the cache directory is generated per filename, so each data file has its own cached dataset.

### DIFF
--- a/nemo_skills/training/nemo_rl/start_sft.py
+++ b/nemo_skills/training/nemo_rl/start_sft.py
@@ -75,11 +75,10 @@ class PromptResponseDataset:
         cache_dir = data_path.parent / ".cache" / f"{split_name}_{data_path.stem}"
         sig_file = cache_dir / "signature.json"
 
-        file_size = str(data_path.stat().st_size)
-
         if cache_dir.exists() and sig_file.exists() and not self.force_reprocess:
             with open(sig_file) as f:
                 old_sig = json.load(f)["size"]
+            file_size = str(data_path.stat().st_size)
             if old_sig == file_size:
                 print(f"[Cache] Loading {split_name} dataset from: {cache_dir}")
                 return load_from_disk(str(cache_dir))


nemo_skills/training/nemo_rl/start_sft.py
```diff
@@ -72,15 +72,15 @@ def __init__(
     def load_or_process_split(self, path: str, split_name: str) -> Dataset:
         data_path = Path(path)
 
-        # Cache path: same folder as data file
-        cache_dir = data_path.parent / ".cache" / split_name
+        # Build cache dir using split name + file stem
```
Kipok: can we instead do md5sum or something like this? To force hash to be re-generated if I update the contents of the file, but keep the name the same

wedu-nvidia: sure, it sounds more robust.

wedu-nvidia: I use the file size as a quick pre-check for caching. In practice, it’s uncommon for the file contents to change without a change in size.

nemo_skills/training/nemo_rl/start_sft.py
```diff
@@ -71,16 +72,21 @@ def __init__(
...
-        cache_dir = data_path.parent / ".cache" / split_name
+        file_size = str(data_path.stat().st_size)
 
-        # Use cached version unless forced to reprocess
-        if cache_dir.exists() and not self.force_reprocess:
-            print(f"[Cache] Loading {split_name} dataset from: {cache_dir}")
-            return load_from_disk(str(cache_dir))
+        if cache_dir.exists() and sig_file.exists() and not self.force_reprocess:
+            with open(sig_file) as f:
+                old_sig = json.load(f).get("size", "")
```
Kipok:
```suggestion
                old_sig = json.load(f)["size"]
```

nemo_skills/training/nemo_rl/start_sft.py
```diff
@@ -71,16 +72,21 @@ def __init__(
 
     def load_or_process_split(self, path: str, split_name: str) -> Dataset:
         data_path = Path(path)
+        cache_dir = data_path.parent / ".cache" / f"{split_name}_{data_path.stem}"
+        sig_file = cache_dir / "signature.json"
 
-        # Cache path: same folder as data file
-        cache_dir = data_path.parent / ".cache" / split_name
+        file_size = str(data_path.stat().st_size)
```
Kipok: move inside the if so that it's only run when necessary

wedu-nvidia: done

PR #696, opened 2025-08-20
Add support for api_key_env_var

Glorf:
Follow-up of the discussion in https://github.com/NVIDIA/NeMo-Skills/pull/683


Kipok:
@Glorf your original implementation was a bit incomplete, so I pushed some changes (and also fixed an unrelated bug we had with incorrect env name used in CI). Can you please check if that works for you and we can merge after that

PR #697, opened 2025-08-20
Update CONTRIBUTING.md

### LINKED ISSUE #691: Update contribution guidelines and CI checks
Key points to highlight from @Kipok in `CONTRIBUTING.md`.

> * Please always create PRs from a branch, not from a fork. Our CI depends on some secret variables (e.g. HF_TOKEN or NVIDIA_API_KEY) and they cannot be used on fork PRs.
> * Please always sign your commits. This is as simple as adding -s flag when doing git commit command.
> * Please enable pre-commit hook to ensure your code is automatically formatted according to our code style.

TODO:

- [x] Add CI checks equivalent to the pre-commit hooks.

activatedgeek:
Fixes #691


pyproject.toml
```diff
@@ -55,16 +55,7 @@ exclude = ["tests", "tests.*"]
 [tool.setuptools.dynamic]
 version = { attr = "nemo_skills.version.__version__" }
 dependencies = {file = ["requirements/main.txt"]}
-optional-dependencies = {core = { file = ["requirements/main.txt"] }, all = { file = ["requirements/main.txt"] }}
-
-[tool.isort]
-profile = "black"  # black-compatible
-line_length = 119  # should match black parameters
```
Kipok, line 62: don't we need to carry over this setting? That's probably non-default in ruff, right?

Kipok:
I would probably run precommit as a separate workflow. While both are required, it's often helpful to see test failures even if code style is not fixed yet, so that you can fix both at the same time or discuss test failures separate from the lint checks

pyproject.toml
activatedgeek: Fixed. A lot of other options are defaults from `black` as per [this doc](https://docs.astral.sh/ruff/configuration/#configuring-ruff).

activatedgeek:
Updated. Perhaps once merged to `main`, you might want to enforce rebase on existing PRs ready to be merged so that we can verify any issues. Post-commit, I did verify that it checks for all files on PR (also printing those in the logs) but good to do another real check.

PR #698, opened 2025-08-20
Support gemini api

hsiehjackson:
* Support gemini api based on `litellm`
* Support `reasoning_effort` for gemini models including disable thinking and different budgets


Kipok (commented):
please add this new server type here as well https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/utils/server.py#L37

nemo_skills/inference/model/gemini.py
```diff
@@ -0,0 +1,115 @@
...
+        self.model = model
+        self.litellm_kwargs = dict(
+            model=model_litellm,
+            max_retries=max_retries,
+        )
+        httpx_limits = httpx.Limits(
+            max_keepalive_connections=2048, max_connections=2048
+        )
+        litellm.client_session = httpx.Client(limits=httpx_limits)
+        litellm.aclient_session = httpx.AsyncClient(limits=httpx_limits)
```
smahdavi4: we can call super.init to reuse all the initialization logic and avoid code copying (please see `AzureOpenAIModel`).

hsiehjackson: We don't need most of the logics in the initialization of `OpenAIModel`. Also, if we use `OpenAIModel`, then several names of env vars need to be changed like `NEMO_SKILLS_OPENAI_MODEL` and `NEMO_SKILLS_OPENAI_BASE_URL`. I can change to inherit `BaseModel`, then the next issue is that it requires to use `base_url`.

smahdavi4: Extending `BaseModel` makes sense, maybe we can pass `base_url=""` to the base model and add a logic if the base url was an empty string, we do not include it in the litellm_kwargs.
Also, I think the [api key is reusable across clients](https://github.com/BerriAI/litellm/blob/55dcaded7244a74a7f5a6d7625707427ea4c70d9/litellm/llms/gemini/google_genai/transformation.py#L132) (litellm would use it instead of GEMINI_API). so to keep consistent with other clients, we can pass `api_key` instead of reading directly from the environment variable (if the api_key is None we can read from env variable similar to OpenAI clients).

nemo_skills/inference/model/gemini.py
```diff
@@ -0,0 +1,115 @@
...
+
+        elif reasoning_effort == "dynamic":
+            reasoning_effort = None
+            # https://github.com/BerriAI/litellm/blob/v1.75.0-nightly/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py#L451-L465
+            params["thinking"] = {
+                "type": "enabled",
+                "budget_tokens": -1,
+            }
+        
+        if self.model == "gemini-2.5-pro":
```
smahdavi4: Wouldn't this raise a server side error? If so, we can remove it from our side.

nemo_skills/inference/model/gemini.py
```diff
@@ -0,0 +1,115 @@
...
+        max_retries: int = 3,
+        **kwargs,
+    ):
+        """
+        model:
+            - gemini-2.5-pro: thinking budget 128-32768 (default: dynamic thinking)
+            - gemini-2.5-flash: thinking budget 0-24576 (default: dynamic thinking)
+            - gemini-2.5-flash-lite: thinking budget 0-24576 (default: no thinking)
+        """
+        assert os.getenv("GEMINI_API_KEY") is not None, "GEMINI_API_KEY is not set"
```
Kipok: @hsiehjackson sorry just merged a change to the logic of getting api keys, can you put this inside `_get_api_key` method like for other models now?

nemo_skills/inference/model/gemini.py
```diff
@@ -0,0 +1,104 @@
...
+        **kwargs,
+    ):
+        """
+        model:
+            - gemini-2.5-pro: thinking budget 128-32768 (default: dynamic thinking)
+            - gemini-2.5-flash: thinking budget 0-24576 (default: dynamic thinking)
+            - gemini-2.5-flash-lite: thinking budget 0-24576 (default: no thinking)
+        """
+
+        api_key = self._get_api_key(api_key, api_key_env_var="GEMINI_API_KEY", base_url=None)
```
Kipok: can you instead define an explicit function like for azureopenai which will take api key from parent module and only if it's None will use GEMINI_API_KEY and fail if it's not defined? We want to allow users to provide api_key_env_var in case e.g. they use different naming convention for their gemini. So parent super call will take care of that, but if it's still None after, this means we should check the standard env var

Kipok: if you define it as a method, you don't need to call it here as base class will call it in its init

hsiehjackson: Oh I see. The function is used to let user create their own env variable name.

nemo_skills/inference/model/gemini.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(
+        self,
+        *args,
+        model: str,
+        api_key: str | None = None,
+        **kwargs,
+    ):
+        """
+        model:
+            - gemini-2.5-pro: thinking budget 128-32768 (default: dynamic thinking)
```
smahdavi4: All the defaults are "disable" now, right? (unless we want to define `reasoning_effort=None` as dynamic.

hsiehjackson: Yes, if we don't set `reasoning_effort`, we disable thinking as default.

smahdavi4 (approved):

PR #699, opened 2025-08-21
fix: scicode missing functions

jubick1337:


Kipok (approved):
Thanks! Did you check if this has a significant impact on accuracy on e.g. qwen3 or other models?

jubick1337:
> Thanks! Did you check if this has a significant impact on accuracy on e.g. qwen3 or other models?

Preliminary check on `OpenReasoning-Nemotron-7B` subtask accuracy increased from 17.3% to 19.7%

PR #702, opened 2025-08-21
add skip home check

wedu-nvidia:
we enforce skip home check when submitting jobs. 
it is recommended to set 
`HF_HOME=/path/to/your/hf_home" `
in the cluster config env_vars.


nemo_skills/pipeline/convert.py
```diff
@@ -236,6 +236,7 @@ def convert(
     log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs."),
     exclusive: bool = typer.Option(False, help="If set will add exclusive flag to the slurm job."),
     check_mounted_paths: bool = typer.Option(False, help="Check if mounted paths are available on the remote machine"),
+    hf_home_check: bool = typer.Option(False, help="If False, skip checking HF_HOME in cluster_config."),
```
Kipok: we should do the check by default, so let's maybe call this `skip_hf_home_check` and if it's False, we check for env var?

Kipok (commented):
can you please also update `ns setup` pipeline to add another prompt for users asking to setup HF_HOME and if mentioning there that it's required by default, but they can use this parameter to opt-out of the check?

nemo_skills/pipeline/convert.py
wedu-nvidia: ok, I thought you wanna skip check by default. 
I will change it back.

nemo_skills/pipeline/prepare_data.py
```diff
@@ -54,6 +54,11 @@ def prepare_data(
     log_dir: str = typer.Option(None, help="Custom location for slurm logs"),
     exclusive: bool = typer.Option(False, help="If set will add exclusive flag to the slurm job."),
     check_mounted_paths: bool = typer.Option(False, help="Check mounted paths availability"),
+    skip_hf_home_check: bool = typer.Option(
+        False,
+        "--skip-hf-home-check",
```
Kipok: why do we need this special version of the name?

wedu-nvidia:
I added a prompt for HF_HOME in the ns setup pipeline and adjusted configurations to ensure all unit tests pass successfully. 
Please let me know if this revision looks good to you.

Kipok (approved):

PR #713, opened 2025-08-22
Adding MCP clients

gwarmstrong:
#### Overview
This PR introduces a first-pas at using Model Context Protocol (MCP) tools across different servers and model providers via a standard internal interchange format. It includes:
- A minimal client abstraction over multiple MCP transports.
- An adapter layer to bridge our internal tool format with different model schemas and tool-call outputs.
- Several example MCP servers (both “standard” and custom) and a demo loop that wires it together with a local vLLM model.

This is a draft intended to get feedback on the design and direction before hardening APIs and adding state, auth, and declarative configuration.

### What’s included

- **Internal adapters (`adapters.py`)**
  - `ToolSchemaAdapter`, `ToolCallInterpreter`, `ToolResponseFormatter` interfaces and an `AdapterRegistry`.
  - Implementations for OpenAI-style schemas and Qwen-compatible formatting:
    - `OpenAISchemaAdapter`, `OpenAICallInterpreter`, `OpenAIResponseFormatter`
    - `QwenResponseFormatter` (re-uses OpenAI schema + interpreter), pre-registered in `registry` for model types "openai" and "qwen".
  - Designed so we can also support GPT-OSS models via the same interfaces.

- **Clients and manager (`clients.py`)**
  - Unified client interface: `MCPClient` with `list_tools()` and `call_tool()`.
  - Transports:
    - `MCPHttpClient` (simple HTTP JSON endpoints: `/list_tools`, `/call_tool`)
    - `MCPStreamableHttpClient` (MCP streamable HTTP via `mcp.client.streamable_http`)
    - `MCPStdioClient` (MCP stdio via `mcp.client.stdio`)
  - `MCPClientManager` to:
    - Register multiple clients by name.
    - Merge and cache tool lists from all clients.
    - Resolve tool name → client mapping (latest registered client wins on collisions).
    - Execute a tool by name regardless of which client serves it.
  - Quality-of-life:
    - `hide_args` masking for schemas (via `MCPClientMeta` wrapping). This allows the client to continue to manipulate the tool call (e.g., to adjust timeout) without exposing these parameters to the model.

- **Example servers (`servers/`)**
  - `server_math.py` (FastAPI):
    - Tools: `add`, `mul` with a simple JSON schema contract.
    - Intentionally biased results (+2, ×2) to test model/tool reasoning in the demo.
  - `server_string.py` (FastAPI):
    - Tool: `concat`.
  - `plane_tickets.py` (FastMCP, streamable HTTP, mounted under Starlette at `/plane`):
    - Tools: `list_available_cities`, `list_destination_cities`, `book_trip`.
    - Demonstrates first-class MCP server with `stateless_http=True`.
  - `python_tool.py` (FastMCP over stdio) with `LocalSandbox`:
    - Tool: `execute(code, session_id?, timeout=10)` returning `ExecutionResult`.
    - Used via stdio client; arguments `session_id` and `timeout` are hidden in the exposed schema.

- **Demo (`demo.py`)**
  - Wires up:
    - vLLM model (`VLLMModel`) targeting a local OpenAI-compatible endpoint.
    - Four clients: `math` (HTTP on 8001), `string` (HTTP on 8002), `plane` (streamable HTTP on 8003/plane), `python` (stdio).
    - `MCPClientManager` to merge/resolve tools; name collisions are resolved by “last registered wins.”
    - Adapter registry for `model_type='qwen'` (OpenAI schema + Qwen response formatting).
  - Runs a multi-round tool-call loop:
    - Converts merged tools to model schema.
    - Calls model, interprets tool calls, executes tools, formats tool outputs, and appends them back into the conversation.

### How it works (end-to-end)
1. Start multiple servers; each exposes tools in its own way (FastAPI JSON, FastMCP stdio/HTTP).
2. Register clients in `MCPClientManager`.
3. `list_all_tools()` merges and caches tools across clients into a common internal format.
4. Adapter registry converts the internal tools into the model’s required schema.
5. Model generates with `tools=...`; any tool calls are interpreted via the adapter and executed via the manager.
6. Tool results are formatted back into the conversation using the model-specific response formatter.
7. The loop continues until no further tool calls.

### Local run
- Start servers (in separate shells):
```bash
uvicorn nemo_skills.mcp.servers.server_math:app --port 8001
uvicorn nemo_skills.mcp.servers.server_string:app --port 8002
uvicorn nemo_skills.mcp.servers.plane_tickets:app --reload --port 8003
# Launch sandbox server with docker
```
- Ensure a vLLM/OpenAI-compatible server is running at `127.0.0.1:8000` with model `Qwen/Qwen3-8B`.
- Run the demo:
```bash
python -m nemo_skills.mcp.demo
```

### Design notes
- **Standard interchange**: A single internal representation for tools, calls, and results, adapted at the edges for different models and MCP servers.
- **Multiple transports**: HTTP JSON, streamable HTTP (MCP), and stdio (MCP) supported behind a common interface.
- **Merging and collisions**: Tools from multiple clients are merged and a model can seemlessly call between multiple clients.
- **Schema masking**: Allows the client to continue to manipulate the tool call (e.g., to adjust timeout) without exposing these parameters to the model.
- **Model-agnostic adapters**: Pluggable schema, interpreter, and response formatting; `openai` and `qwen` supported now (only `qwen` tested so far).

### Known gaps / next steps
- **Statefulness**: All clients are currently stateless. Add session persistence where appropriate (e.g., python REPL continuity) and define lifecycle APIs in clients and manager.
- **Tool name scoping**: Support servers exposing tools with the same `name`. Options:
  - Namespacing (e.g., `server.tool`).
- **Declarative registration**: Replace imperative client registration in `demo.py` with a declarative spec (YAML/JSON/env) that maps to client types, endpoints, auth, and policy. This is required to run with NeMo-Skills pipelines.
- **Authentication**: Needs concrete MCP servers to validate client implementations. There are recommended auth flows (e.g., tokens, OAuth), but servers can also have one-off auth. Until we have a good example server, client-side auth support will remain limited and provisional.
- **Session dynamics**: Our current sessions could be more dynamic with respect to tool and server availability (e.g., refreshing, health/availability awareness, and re-discovery over time).


nemo_skills/mcp/servers/python_tool.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import Field
+from typing import Annotated, Optional
+from mcp.server.fastmcp import FastMCP
+from nemo_skills.code_execution.sandbox import LocalSandbox
+
+logger = logging.getLogger(__name__)
+
+mcp = FastMCP(name="python_tool")
+
+sandbox = LocalSandbox()
```
activatedgeek: We need a way from the `eval`/`generate` launch arguments for the sandbox server arguments.

nemo_skills/mcp/servers/server_math.py
activatedgeek, line 1: I don't think we need this file. The other files act as example enough if the purpose is to have a reference implementation.

nemo_skills/mcp/servers/server_string.py
activatedgeek, line 1: Same here. I don't think we need to add such servers that aren't really useful for real examples.

nemo_skills/mcp/servers/weather.py
activatedgeek, line 1: Same here. I don't think we need to add such servers that aren't really useful for real examples.

nemo_skills/mcp/servers/plane_tickets.py
activatedgeek, line 1: I don't think we should add trivial examples. The Python tool should be sufficient as a reference implementation.

activatedgeek (changes requested):

nemo_skills/mcp/servers/plane_tickets.py
gwarmstrong: For sure--they're placeholders for now and useful for testing, but these will get removed along with the demo.py script for the final PR.

nemo_skills/mcp/servers/python_tool.py
activatedgeek: Specifically, I am talking about this kind of a structure:

https://github.com/NVIDIA/NeMo-Skills/blob/8214c42a14f8d815eb100c920d4bb33f6e9a5d01/nemo_skills/inference/generate.py#L265

activatedgeek:
> Start servers (in separate shells):

@gwarmstrong For ease of use, may be we can have a single entrypoint for the MCP server, such that any tool registered (i.e. the MCP decorator called via imports) would be enabled? In addition, we perhaps need switches to enable the specific registered servers.

And subsequently, this single command can be launched via the sandbox container. I'm not sure if the sandbox container handles multiple processes, but it should be easy to use a process manager within the container.

nemo_skills/mcp/servers/plane_tickets.py
```diff
@@ -0,0 +1,123 @@
...
+from starlette.routing import Mount
+from mcp.server.fastmcp import FastMCP
+
+"""
+Run with:
+    uvicorn nemo_skills.mcp.servers.plane_tickets:app --reload --port 8003
+"""
+
+# Initialize FastMCP server
+mcp = FastMCP(name="plane_tickets", stateless_http=True)
```
Kipok: just to double check, the plan is to have this as part of the sandbox, right? So it will keep its current api, but also add the new endpoints for tools? The current examples use stdio for communication, which I think we won't actually need for the initial implementation, but instead we should test an http protocol there

nemo_skills/mcp/clients.py
```diff
@@ -0,0 +1,229 @@
...
+    return wrapped
+
+def inject_hide_args(init_func):
+    @functools.wraps(init_func)
+    def wrapper(self, *args, hide_args=None, **kwargs):
+        self._hide_args = hide_args or {}
+        return init_func(self, *args, **kwargs)
+    return wrapper
+
+class MCPClientMeta(type):
```
Kipok, line 146: can you please explain what's the purpose of this metaclass?

nemo_skills/mcp/adapters.py
```diff
@@ -0,0 +1,125 @@
...
+class OpenAICallInterpreter(ToolCallInterpreter):
+    def parse(self, tool_call):
+        fn = tool_call.function
+        tool = fn.name
+        return {"tool_name": tool, "args": json.loads(fn.arguments)}
+
+
+class OpenAIResponseFormatter(ToolResponseFormatter):
+    # https://platform.openai.com/docs/guides/function-calling
+    def format(self, tool_call: ChatCompletionMessageToolCall, result):
```
Kipok: is this different because it uses responses api? I imagine qwen and openai chat completions should use the same interface, right? As in, if this is an input to one of openai.X functions, then the interface should be model independent as long as the function is the same?

nemo_skills/mcp/servers/python_tool.py
gwarmstrong: Understood and working on it. The crux and decision points lie in actually passing the config to the tool.

nemo_skills/mcp/clients.py
gwarmstrong: This is the layer that helps seamlessly enables the `hide_args` functionality that allows a client to have to tool parameters without exposing them to the model. I will add a more informative docstring for this.

nemo_skills/mcp/adapters.py
gwarmstrong: good catch, yeah that seems to be the source of the difference. Inspecting both of their templates, they use the `role: tool` and `content: <content>` flags in the messages, so this difference from the docs is likely due to the different APIs.

I think we may still need some response adapter logic here (e.g., claude se? Especially if we want to have the ability to use different API functions. But I'll at least get that reflected in the naming.

nemo_skills/mcp/servers/plane_tickets.py
gwarmstrong: I think this comment was intended for the python tool?

Assuming that's the case, we can do it either way.

IMO there is actually a good argument for implementing it as stdio first rather than http--and that's basically that it's the more extensible and modular approach. We re-use the existing sandbox and keep it general purpose (e.g., execute a script). Then implementing a new tool is effectively just writing a simple client wrapper, rather than exposing a new endpoint, and this allows people to use different tool configurations more easily with lower impact on nemo-skills code.

E.g.,
assume we have
```
built-in:
* python tool
* lean tool
* exa web search tool
```
and alice wants to make a tool that queries financial data from the yahoo finance api, she could just write a new file with a tool:

```
# /nemo_run/code/finance_data/tool.py
# headers go here
@mcp.tool()
async def execute(
        ticker: Annotated[str, Field(description="Stock ticker")],
    ) -> ExecutionResult:
    """Executes the given python code"""
    language = "ipython"
    code = f"""import requests; requests.get("https://query2.finance.yahoo.com/v7/finance/quote?symbols={ticker}"""
    output, _ = await sandbox.execute_code(code, language=language)
    logger.info('Ran request with status: %s', output["process_status"])
    return output


if __name__ == "__main__":
    # Initialize and run the server
    mcp.run(transport='stdio')
```

and then extend a configuration file with something like:
```
{...
 "tools": ["finance": "python /nemo_run/code/finance_data/tool.py"]
 }
```

and then the tools are flexible rather than having to update the server and maintain a separate server code, or do more complex server setups.


that all being said, to have our basic set of tools implemented as endpoints should all be fine to do if that's what we want, and should still be compatible with these one-off stdio tools via the MCPClientManager.

nemo_skills/mcp/clients.py
gwarmstrong: okay I've added some more functionality here as well as a docstring that explains it

nemo_skills/mcp/adapters.py
gwarmstrong: I have corrected the naming and logic around this

nemo_skills/mcp/servers/exa_tool.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+mcp = FastMCP(name="exa_tool")
+
+# Populated from CLI args in main()
+EXA_API_KEY: str | None = None
+
+
+@mcp.tool()
+async def exa_websearch(
```
gwarmstrong, line 41: @activatedgeek 
here is a minimal example using the tool the way i wrote it, can you ensure it works in the way you need it to?

```python
import asyncio
import os

from nemo_skills.mcp.servers import exa_tool as exa


async def main():
    # Provide your Exa API key (env var preferred)
    exa.EXA_API_KEY = os.environ.get("EXA_API_KEY") or "YOUR_EXA_API_KEY"

    query = "What are the latest features in Python 3.12?"
    res = await exa.exa_websearch(query)

    print(res)

if __name__ == "__main__":
    asyncio.run(main())
```

nemo_skills/mcp/servers/python_tool.py
```diff
@@ -0,0 +1,83 @@
...
+
+mcp = FastMCP(name="python_tool")
+
+# Initialized from config in main()
+sandbox = None
+
+
+@mcp.tool()
+async def execute(
+    code: Annotated[str, Field(description="Code to run in python interpretter")],
```
Kipok:
```suggestion
    code: Annotated[str, Field(description="Code to run in python interpreter")],
```

nemo_skills/mcp/clients.py
```diff
@@ -0,0 +1,574 @@
...
+    - list_tools() fetches tool metadata from the running stdio server.
+    - call_tool() returns the server's structuredContent.
+
+    The following optional configurables can be supplied (injected by the
+    metaclass): hide_args, disabled_tools, enabled_tools, output_formatter,
+    init_hook.
+
+    Example (OmegaConf config, like in mcp_demo.py):
+    ```python
+    from omegaconf import OmegaConf
```
Kipok, line 401: do you want to put these examples as tests?

Kipok (approved):
it's a bit hard to judge whether the current abstractions are good or not without a full use-case, but I think it should be ok to merge as is and then refine later

nemo_skills/mcp/clients.py
gwarmstrong: Can I depend on the sandbox during pytests on CI? I wrote some that are agnostic to the Sandbox, but if I can use it, then I can test the integration too.

nemo_skills/mcp/servers/exa_tool.py
activatedgeek: I think the request looks equivalent to what I had, so that should be good. I'm not sure how to test it though?

PR #717, opened 2025-08-22
Tool support based on Chat Completion APIs

activatedgeek:
A JSON dump of example arguments.
```json
{
  "benchmarks": "gpqa",
  "split": "debug",
  "data_dir": "/store/data/nemo_skills",
  "server_type": "vllm",
  "model": "/store/models/hf_models/Qwen3-14B",
  "server_gpus": 1,
  "server_nodes": 1,
  "server_args": "--async-scheduling --enable-auto-tool-choice --tool-call-parser hermes --reasoning-parser deepseek_r1",
  "with_sandbox": true,
  "output_dir": "....",
  "expname": "...",
  "cluster": "....",
  "partition": "...",
  "time_min": 30,
  "exclusive": false,
  "ctx": "wrap_arguments('++prompt_config=/nemo_run/code/nemo_tir/prompt/config/gpt-oss/gpqa.yaml ++inference.tokens_to_generate=20000 ++inference.temperature=1.0 ++inference.top_p=1.0 ++max_concurrent_requests=1024 ++tool_config=/nemo_run/code/nemo_skills/mcp/config/exa.yaml ++skip_filled=True')"
}
```

TODO:
- [x] Support for MCP servers
- [x] Support for tool descriptions as input to the chat completion request.
- [ ] Add tests

Out of Scope:
- Support for custom parsing for models that do not support tool calling with Chat Completion API.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -136,6 +136,7 @@ class GenerateSolutionsConfig:
     # genselect config
     online_genselect_config: OnlineGenSelectConfig = field(default_factory=OnlineGenSelectConfig)
 
+    ## FIXME(sanyamk): Rethink the structure of this configuration.
     tool_config: str | None = None  # Path to tool configuration file.
 
     # if True, will move full generation to _full_generation key and keep cfg.generation_key without thinking tokens
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -136,9 +136,7 @@ class GenerateSolutionsConfig:
     # genselect config
     online_genselect_config: OnlineGenSelectConfig = field(default_factory=OnlineGenSelectConfig)
 
-    # TODO(sanyamk): tool configuration path
-    with_tools: bool = False
-    tool_config: str | None = None  # Path to tool configuration file. Required in `with_tools` is True
+    tool_config: str | None = None  # Path to tool configuration file.
 
     # if True, will move full generation to _full_generation key and keep cfg.generation_key without thinking tokens
     remove_thinking: bool = False
@@ -272,7 +270,7 @@ class GenerationTask:
 
         if self.cfg.code_execution:
             llm = get_code_execution_model(**self.cfg.server, sandbox=self.sandbox)
-        elif self.cfg.with_tools:
+        elif self.cfg.tool_config:
             llm = get_tool_calling_model(
                 **self.cfg.server, tool_config=self.cfg.tool_config, additional_config={"sandbox": self.cfg.sandbox}
             )
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -22,7 +22,7 @@ import time
 from copy import deepcopy
 from dataclasses import asdict, field, is_dataclass
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
 
 import hydra
 from omegaconf import ListConfig, OmegaConf
@@ -34,6 +34,7 @@ from nemo_skills.inference.model import (
     get_code_execution_model,
     get_model,
     get_online_genselect_model,
+    get_tool_calling_model,
     server_params,
 )
 from nemo_skills.prompt.utils import get_prompt
@@ -134,6 +135,9 @@ class GenerateSolutionsConfig:
     online_genselect: bool = False
     # genselect config
     online_genselect_config: OnlineGenSelectConfig = field(default_factory=OnlineGenSelectConfig)
+
+    # TODO(sanyamk): tool configuration path
+    with_tools: bool = False
 
     # if True, will move full generation to _full_generation key and keep cfg.generation_key without thinking tokens
     remove_thinking: bool = False
@@ -263,10 +267,16 @@ class GenerationTask:
         self.output_lock = None
 
     def setup_llm(self):
+        self.sandbox = get_sandbox(**self.cfg.sandbox) if self.cfg.sandbox is not None else None
+
         if self.cfg.code_execution:
-            sandbox = get_sandbox(**self.cfg.sandbox) if self.cfg.sandbox is not None else None
-            llm = get_code_execution_model(**self.cfg.server, sandbox=sandbox)
-        elif self.cfg.online_genselect:
+            llm = get_code_execution_model(**self.cfg.server, sandbox=self.sandbox)
+        elif self.cfg.with_tools:
+            llm = get_tool_calling_model(**self.cfg.server, sandbox=self.sandbox)
+        else:
+            llm = get_model(**self.cfg.server)
+
+        if self.cfg.online_genselect:
             # Use the same prompt parameters for genselect as the one used for generation
             self.cfg.online_genselect_config.use_completions_api = self.cfg.use_completions_api
             self.cfg.online_genselect_config.tokenizer = self.cfg.tokenizer
@@ -276,8 +286,6 @@ class GenerationTask:
             llm = get_online_genselect_model(
                 **self.cfg.server, online_genselect_config=self.cfg.online_genselect_config
             )
-        else:
-            llm = get_model(**self.cfg.server)
 
         return llm
 
@@ -445,10 +453,10 @@ class GenerationTask:
             inference_params = dict(self.cfg.inference)
 
         generation_params = {
-            "prompt": self.fill_prompt(data_point, all_data),
-            "stop_phrases": [self.cfg.stop_phrase] if self.cfg.stop_phrase else None,
             **inference_params,
             **self.extra_generate_params,
+            "prompt": self.fill_prompt(data_point, all_data),
+            "stop_phrases": [self.cfg.stop_phrase] if self.cfg.stop_phrase else None,
         }
 
         if self.cfg.code_execution:
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -284,7 +284,8 @@ class GenerationTask:
             self.cfg.online_genselect_config.thinking_begin = self.cfg.thinking_begin
             self.cfg.online_genselect_config.thinking_end = self.cfg.thinking_end
             llm = get_online_genselect_model(
-                **self.cfg.server, online_genselect_config=self.cfg.online_genselect_config
+                **{**self.cfg.server, "model": llm},
+                online_genselect_config=self.cfg.online_genselect_config,
             )
 
         return llm
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -138,6 +138,7 @@ class GenerateSolutionsConfig:
 
     # TODO(sanyamk): tool configuration path
     with_tools: bool = False
+    tool_config: str | None = None  # Path to tool configuration file. Required in `with_tools` is True
 
     # if True, will move full generation to _full_generation key and keep cfg.generation_key without thinking tokens
     remove_thinking: bool = False
@@ -272,7 +273,9 @@ class GenerationTask:
         if self.cfg.code_execution:
             llm = get_code_execution_model(**self.cfg.server, sandbox=self.sandbox)
         elif self.cfg.with_tools:
-            llm = get_tool_calling_model(**self.cfg.server, sandbox=self.sandbox)
+            llm = get_tool_calling_model(
+                **self.cfg.server, tool_config=self.cfg.tool_config, additional_config={"sandbox": self.cfg.sandbox}
+            )
         else:
             llm = get_model(**self.cfg.server)
 
--- a/nemo_skills/inference/model/__init__.py
+++ b/nemo_skills/inference/model/__init__.py
@@ -71,10 +71,10 @@ def get_online_genselect_model(model, online_genselect_config=None, **kwargs):
     return OnlineGenSelectWrapper(model=model, cfg=online_genselect_config or OnlineGenSelectConfig())
 
 
-def get_tool_calling_model(model, sandbox, **kwargs):
+def get_tool_calling_model(model=None, tool_config=None, additional_config=None, **kwargs):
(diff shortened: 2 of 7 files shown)


Kipok (commented):
can we move this into a separate model class, not generation module? Similar to online genselect here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/model/online_genselect.py? So this class would define a generate_async method and would be an in-place replacement for the self.llm in any generation module? That way this tool-calling support will be possible to use in any sdg / eval setup

nemo_skills/inference/generate.py
```diff
@@ -261,10 +265,16 @@ def __init__(self, cfg: GenerateSolutionsConfig):
...
     def setup_llm(self):
+        self.sandbox = get_sandbox(**self.cfg.sandbox) if self.cfg.sandbox is not None else None
+
         if self.cfg.code_execution:
-            sandbox = get_sandbox(**self.cfg.sandbox) if self.cfg.sandbox is not None else None
-            llm = get_code_execution_model(**self.cfg.server, sandbox=sandbox)
-        elif self.cfg.online_genselect:
+            llm = get_code_execution_model(**self.cfg.server, sandbox=self.sandbox)
+        elif self.cfg.with_tools:
+            llm = get_tool_calling_model(**self.cfg.server, sandbox=self.sandbox)
```
gwarmstrong: It would be better to pass the sandbox config here rather than the sandbox object, since it will need to be passed over a communication barrier

activatedgeek: I don't quite understand what you mean here. Can you elaborate please?

gwarmstrong: `self.sandbox` is an a concrete instance of `class Sandbox`. But I can't pass this to a tool--it has to be something I can pass to a command line call. So it would be better if you did something like:
```
get_tool_calling_model(**self.cfg.server, sandbox_cfg=self.cfg.sandbox, tool_cfg=self.cfg.tools)
```
or
```
get_tool_calling_model(cfg=self.cfg)
```

but also, feel free to defer and I can adjust the interface once I finish the tool calling implementation (probably later today or early tomorrow).

activatedgeek: Understood. Yes, I think once we have finalized the connecting interface to the MCP server, we can make changes to these methods as well.

nemo_skills/inference/model/__init__.py
```diff
@@ -61,12 +64,17 @@ def get_code_execution_model(server_type, code_execution=None, sandbox=None, **k
...
-    model = get_model(server_type=server_type, **kwargs)
-    if online_genselect_config is None:
-        online_genselect_config = OnlineGenSelectConfig()
-    return OnlineGenSelectWrapper(model=model, cfg=online_genselect_config)
+    if isinstance(model, str):
+        model = get_model(model=model, **kwargs)
+    return OnlineGenSelectWrapper(model=model, cfg=online_genselect_config or OnlineGenSelectConfig())
+
+
+def get_tool_calling_model(model=None, sandbox=None, **kwargs):
```
Kipok:
```suggestion
def get_tool_calling_model(model, sandbox, **kwargs):
```
both are not really optional

nemo_skills/inference/model/__init__.py
```diff
@@ -61,12 +64,17 @@ def get_code_execution_model(server_type, code_execution=None, sandbox=None, **k
     return CodeExecutionWrapper(model=model, sandbox=sandbox, config=code_execution_config)
 
 
-def get_online_genselect_model(server_type, online_genselect_config=None, **kwargs):
+def get_online_genselect_model(model=None, online_genselect_config=None, **kwargs):
```
Kipok:
```suggestion
def get_online_genselect_model(model, online_genselect_config=None, **kwargs):
```

nemo_skills/inference/model/__init__.py
```diff
@@ -61,12 +64,17 @@ def get_code_execution_model(server_type, code_execution=None, sandbox=None, **k
...
-    model = get_model(server_type=server_type, **kwargs)
-    if online_genselect_config is None:
-        online_genselect_config = OnlineGenSelectConfig()
-    return OnlineGenSelectWrapper(model=model, cfg=online_genselect_config)
+    if isinstance(model, str):
+        model = get_model(model=model, **kwargs)
+    return OnlineGenSelectWrapper(model=model, cfg=online_genselect_config or OnlineGenSelectConfig())
+
+
+def get_tool_calling_model(model=None, tool_config=None, additional_config=None, **kwargs):
```
Kipok:
```suggestion
def get_tool_calling_model(model, tool_config, additional_config=None, **kwargs):
```

nemo_skills/inference/model/tool_call.py
```diff
@@ -0,0 +1,130 @@
...
+    async def _execute_tool_call(self, tool_call):
+        ## TODO(sanyamk): The correct key format needs to be cohesive with other formatters.
+        tool_name = tool_call["function"]["name"]
+        tool_args = tool_call["function"]["arguments"]
+
+        try:
+            tool_args = json.loads(tool_args)
+        except json.decoder.JSONDecodeError as e:
+            LOG.exception(e)
+            return {"error": "Tool argument parsing failed."}
```
Kipok, line 65: is this being passed back to the model? If so, should we include the exception details in here?

nemo_skills/inference/model/tool_call.py
```diff
@@ -0,0 +1,130 @@
...
+
+        try:
+            result = await self.client_manager.execute_tool(tool_name, tool_args)
+        except Exception as e:
```
Kipok, line 70: in the future, we should probably have a separate exception class for tool call failure vs any other failure. E.g. if there is a bug in the client that's always failing, we should stop the job with exception, while if there is some intermittent failure because tool isn't used properly, we should catch it and get it back into the model.

nemo_skills/inference/model/tool_call.py
```diff
@@ -0,0 +1,130 @@
...
+        assert tools is None, "Specify ++tool_config=</path/to/file.yaml> only."
+
+        tools = self.schema_adapter.convert(await self.client_manager.list_all_tools())
+
+        ## Bookkeeping.
+        num_tool_calls = 0
+        generation_steps = []
+        reasoning_steps = []
+
+        conversation = list(prompt)
```
Kipok: .copy() is probably more explicit here as we don't really want to convert to list anything, but do want to copy it

nemo_skills/inference/model/tool_call.py
```diff
@@ -0,0 +1,130 @@
...
+
+        while True:
+            generation = await self.model.generate_async(prompt=conversation, tools=tools, **generation_kwargs)
+
+            content = generation.get("generation")
```
Kipok: what's the case when this key might not be there?

nemo_skills/inference/model/tool_call.py
```diff
@@ -0,0 +1,130 @@
...
+
+        ## Bookkeeping.
+        num_tool_calls = 0
+        generation_steps = []
+        reasoning_steps = []
+
+        conversation = list(prompt)
+
+        while True:
+            generation = await self.model.generate_async(prompt=conversation, tools=tools, **generation_kwargs)
```
Kipok: we need some adjustment for tokens_to_generate here, similar to how code execution model does it

nemo_skills/inference/generate.py
```diff
@@ -133,6 +134,8 @@ class GenerateSolutionsConfig:
     # genselect config
     online_genselect_config: OnlineGenSelectConfig = field(default_factory=OnlineGenSelectConfig)
 
+    tool_config: str | None = None  # Path to tool configuration file.
```
Kipok, line 140: imo a better structure here would be to pass in a list of tool files / names and then those will have all parameters defined within them. And extra user-defined config parameters (e.g. for sandbox timeout, etc.) can be passed in via a separate config dictionary in the form tool_config: {<tool name>: <arbitrary dict of params>}. This will make it simpler to pick different tools and re-configure them without always writing new configs for every change.

we don't need to do this in this pr, but if you agree that this is the structure we should try to aim for, maybe add a todo here

Kipok (commented):
is there an assumption currently that tool call has to be formatted as json for both llm generation and tool output? I think that's not always the case, e.g. some models use xml and it might actually make more sense for e.g. python code as it's tricky to escape it in a json string

activatedgeek:
> is there an assumption currently that tool call has to be formatted as json for both llm generation and tool output? I think that's not always the case, e.g. some models use xml and it might actually make more sense for e.g. python code as it's tricky to escape it in a json string

This assumption is actually configurable, and can be modified using the different MCP adapters - specifically the call interpreter, and the response formatter. As I've also noted in code as a TODO, we need to modify this adapter interface a bit and rather unify into a single class because they all need to be coherent in terms of formatting.

nemo_skills/inference/model/tool_call.py
activatedgeek: Yes, for now the default behavior I am keeping is that the error response goes back into context, but also don't want to send in the whole traceback to avoid uselessly filling up the context length.

Ideally, this logic of parsing the arguments must be with the tool function itself because the format may not be JSON for every model. I can perhaps make modifications to the tool function signatures such that the tool args are just the raw arguments string from the model (either parsed manually or from any model server). The tool function itself must now handle the raw string (decode as JSON, etc.), and subsequently may be also decide what kind of error message needs to be set.

nemo_skills/inference/model/tool_call.py
activatedgeek: Let me add this as a TODO note for the next revision, because error handling here has a few possibilities.

nemo_skills/inference/model/tool_call.py
activatedgeek: Usage of `.get` was mostly a consistency thing. I don't think there is a case when the `generation` key won't exist.

nemo_skills/inference/model/tool_call.py
activatedgeek: Actually no matter how we put it, it will have to be model dependent.

Kipok: actually the parsing of the model output is often handled by the server side, right? E.g. vllm can parse both xml and json depending on what you provide as `--tool-call-parser` parameter, so maybe for now we just assume it's set correctly and let the server handle this? But not sure how it works with getting the output back into the model, hopefully they also handle it properly

nemo_skills/inference/model/tool_call.py
Kipok: then I'd rather not use it to avoid cases when there is a mistake and generation isn't actually returned, but this line causes silent incorrect behavior instead of a clear error right away

nemo_skills/inference/model/tool_call.py
activatedgeek: As long as we follow the Chat Completion API i/o specification, we should be good with using the standard arguments. Because in the vLLM code, the parsing uses model's chat template to convert it into the final raw string (and we can override that chat template also via a CLI argument to the server). If the XML formatting is part of the template, then we will not have to worry about it. For all others, it will have to be handled on a case-by-case basis.

nemo_skills/inference/model/tool_call.py
activatedgeek: I've modified how all the keys for each generation step are collected.

nemo_skills/inference/model/tool_call.py
activatedgeek: This should be fixed now.

nemo_skills/inference/model/tool_call.py
activatedgeek: Using deepcopy now.

nemo_skills/inference/model/tool_call.py
Kipok: I think tool-call-parser is more than jinja. It's some custom logic for a particular way of handle tools going beyond what's in chat template, as far as I understand it

activatedgeek: Yes, the tool call parser simply has things like the regex parser of the raw string, and special tokens to look for when parsing the tool calls. And then it correctly sets the `tool_calls` and `content` appropriately in the response object. But ultimately, vLLM also just takes conversation messages list and pushes it through the model's chat template.

More generally, as long as the chat template supports the `tool` role, we likely don't have to worry about writing custom parsing code because vllm might already support the corresponding parser. For exceptions like `gpt-oss`, we will need to write custom code which will very closely resemble what vLLM does with its custom parser plugins.

I will need to see more examples when such a structure would break (e.g. the chat template does not support tools, or vLLM does not support the model itself, etc.), but once more people start using this code, I'm sure we'll discover other cases that need fixing for.

nemo_skills/inference/generate.py
activatedgeek: I agree, this configuration specification is currently very ugly. 

Here's a suggestion: May be in the PR, we can try rewriting the BFCL benchmark to use this new structure, and we'll know all the places where we need modifications for configurations?

Kipok: BFCL might be a bit too much of an endeavor, but if you want to give it a try, sure. If not, we can just make sure it's easy to add / remove tools from a particular setup and customize specific tools when multiple are used without writing new configs, through commands line / python args

activatedgeek: Ok fair enough. For the next PR, I can make changes as needed for a custom parsing loop and see how the configurations change, since it is higher priority to support `gpt-oss` as well. For now, I can try to generalize the configuration between these two cases (i.e. Chat Completions API support vs no support).

activatedgeek:
@Kipok I'll merge, and will add example/tests once the configuration format is more friendly in the next revision.

PR #718, opened 2025-08-22
Nano v2 tutorial

shtoshni:


docs/tutorials/posts/nemotron-nano-v2-evals.md
```diff
@@ -0,0 +1,376 @@
...
+- top-p to 1.0
+- system message to '/no_think'
+
+
+!!!note
+    The NVIDIA-Nemotron-Nano-9B-v2 is a hybrid model which uses mamba layers along with transformer layers.
+    The model requires the latest version of vllm:
+    ```
+    pip install -U "vllm>=0.10.1"
+    ```
```
Kipok: do we need to add this? Our default container should work, but if you want to keep this, maybe mention that it's only relevant if using non-default container

docs/tutorials/posts/nemotron-nano-v2-evals.md
```diff
@@ -0,0 +1,376 @@
...
+    ++inference.top_p=1.0 \
+    ++system_message='/no_think'
+```
+
+
+### Results
+
+The eval jobs also launch a dependent job to perform metrics calculation and store the result in a file called `metrics.json`.
+In our running example, for a benchmark such as aime25, the `metrics.json` would be located at `/workspace/nvidia_nemotron_nano_9b_v2/eval-results/aime25/metrics.json`.
+This metrics calculation is done typically by the `summarize_results` pipeline, except in the case of BFCL where the metrics are calculated by a BFCL specific script because BFCL has a specific way of combining subtask accuracy to obtain the overall accuracy.
```
Kipok: both BFCL and RULER

docs/tutorials/posts/nemotron-nano-v2-evals.md
```diff
@@ -0,0 +1,376 @@
...
+
+To print the results for these benchmarks (except for BFCL), we could rerun the `summarize_results` script manually as follows:
+```bash
+ns summarize_results --cluster=local /workspace/nvidia_nemotron_nano_9b_v2/eval-results/{BENCHMARK}
```
Kipok, line 253: does this actually work? Without --metrics_type parameter?

Kipok: probably need to remove {BENCHMARK}?

Kipok (changes requested):
thanks, just a few small things

docs/tutorials/posts/nemotron-nano-v2-evals.md
shtoshni: This was meant to be used for individual benchmarks. 
The summary can be quite detailed, so run it separately for each benchmark.

docs/tutorials/posts/nemotron-nano-v2-evals.md
shtoshni: Ok, removed this detail to avoid confusion.

PR #721, opened 2025-08-25
Add SFT data translation module

shuoyangd:
Recreate from #677 from a branch instead of a fork.


nemo_skills/inference/generate.py
```diff
@@ -435,10 +435,19 @@ def prefill_generation(self, data_point) -> dict | None:
...
+            inference_params = asdict(self.cfg.inference)
+        else:
+            # Already a dict from Hydra
+            inference_params = dict(self.cfg.inference)
+
         generation_params = {
             "prompt": self.fill_prompt(data_point, all_data),
-            "stop_phrases": [self.cfg.stop_phrase] if self.cfg.stop_phrase else None,
-            **asdict(self.cfg.inference),
+            "stop_phrases": combine_stop_phrases(
```
shtoshni: `nemo-run/0 NameError: name 'combine_stop_phrases' is not defined`

nemo_skills/pipeline/generate.py
```diff
@@ -207,8 +210,11 @@ def generate(
     if generation_module is None:
         generation_module = GENERATION_MODULE_MAP[generation_type or GenerationType.generate]
 
-    generation_task = importlib.import_module(generation_module)
-    if not hasattr(generation_task, 'GENERATION_TASK_CLASS'):
+    if os.sep in generation_module:
```
Kipok, line 222: Can you please do the same in here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/utils/eval.py#L365?

recipes/translation/translate_jsonl.py
```diff
@@ -0,0 +1,393 @@
+import json
```
Kipok, line 15: license header

nemo_skills/pipeline/utils/eval.py
```diff
@@ -362,7 +362,10 @@ def prepare_eval_commands(
             for chunk_id in benchmark_chunk_ids:
                 job_benchmarks.add(benchmark)
 
-                generation_task = importlib.import_module(generation_module or benchmark_args.generation_module)
+                if os.sep in generation_module:
```
Kipok: the check needs to be for the combined thing `generation_module or benchmark_args.generation_module`. So perhaps assign it to variable first and then do if and reassignment to generation_task after import?

shuoyangd: Right, didn't notice `generation_module` could be `None` there. This new commit should fix it.

PR #723, opened 2025-08-26
Soft failure 

shtoshni:
Enables soft failure and recovery logic as well
The default strategy is to just return an empty generation along with an error and a detailed error entry.  

Also supports three recovery strategies: 
- Reduce generation budget (`reduce_generation`)
- Reduce prompt from start (`reduce_prompt_from_start`) - Trim the prefix of the message list/prompt string to fit in the context window budget
- Reduce prompt from end (`reduce_prompt_from_end`) - Trim the suffix of the message list/prompt string

To enable soft failure for generation/eval jobs with the option of trimming the end of the prompt:
` ++server.enable_soft_fail=True   ++server.context_limit_retry_strategy=reduce_prompt_from_end`


nemo_skills/inference/model/base.py
```diff
@@ -138,10 +161,21 @@ def _build_chat_request_params(self, **kwargs) -> dict:
     def _build_completion_request_params(self, **kwargs) -> dict:
         pass
 
+    def _build_request_params(self, prompt: str | list[dict], stream: bool, **kwargs) -> dict:
+        if isinstance(prompt, str):
+            return self._build_completion_request_params(prompt=prompt, stream=stream, **kwargs)
+        elif isinstance(prompt, list):
+            request_params = self._build_chat_request_params(messages=prompt, stream=stream, **kwargs)
+            request_params["skip_special_tokens"] = False
```
```suggestion
```
this is done in a different place in latest main

nemo_skills/inference/model/base.py
```diff
@@ -212,6 +246,9 @@ async def generate_async(
                 self._maybe_apply_stop_phrase_removal(result, remove_stop_phrases, stop_phrases)
                 return result
 
+            except litellm.exceptions.ContextWindowExceededError as e:
```
Kipok: isn't this a no-op? Or that one is a subclass of BadRequestError? In any case, I think we should change the except part below to re-raise if that message isn't present as we only want to retry that particular error. Which should probably handle this case you need an except for

nemo_skills/inference/model/context_retry.py
```diff
@@ -0,0 +1,449 @@
...
+    Extract token information from LiteLLM context window error messages.
+
+    Returns:
+        Dict with keys: max_context_length, message_tokens, completion_tokens
+        None if parsing fails
+    """
+
+    error_str = str(error)
+
+    # "maximum context length is 40960 tokens and your request has 142 input tokens (1000000 > 40960 - 142)"
```
Kipok, line 43: why are there different messages? Completion vs chat completion? Or it's server dependent? Would be good to add some comment about that

nemo_skills/inference/model/context_retry.py
```diff
@@ -0,0 +1,449 @@
...
+    )
+
+    # Pattern 3: This model's maximum context length is 40960 tokens. However, your request has 3000009 input tokens.
+    pattern3 = re.compile(
+        r"maximum context length is (\d+) tokens.*?"
+        r"your request has (\d+) input tokens",
+        re.IGNORECASE | re.DOTALL,
+    )
+
+    # Try pattern 1 first (matches your example)
```
Kipok:
```suggestion
    # Try pattern 1 first
```

nemo_skills/inference/model/context_retry.py
```diff
@@ -0,0 +1,449 @@
...
+        else:
+            return False
+
+
+def with_context_retry(func: Callable) -> Callable:
+    """
+    Decorator to add context limit retry logic to generate functions.
+    Uses the model's context_limit_retry_config attribute.
+    """
+    default_config = ContextLimitRetryConfig()
```
Kipok: do we need a default config? Might lead to hard-to-find errors in case we e.g. rename the variable in base model and not update here and suddenly the parameters we pass there are ignored

tests/gpu-tests/test_context_retry.py
```diff
@@ -0,0 +1,274 @@
...
+        subprocess.run(cmd, shell=True, check=True)
+
+        assert self.validator.validate_generation_output(output_file), "Output file not found"
+
+
+# Initialize test suite
+test_suite = ContextRetryTestSuite()
+
+
+@pytest.mark.gpu
```
Kipok, line 242: would be good to test sglang / trtllm / api models as well. Also would be good to test with ++use_completions_api=True in at least some cases

nemo_skills/inference/model/base.py
shtoshni: this is to avoid retries, there's no point in doing a retry if we have the window exceed error.

nemo_skills/inference/model/context_retry.py
shtoshni: the different messages depend on the error. For example, if the prompt itself is too long, the error would be different, and if the generation asked for is long, there would be a different error

nemo_skills/inference/model/base.py
Kipok: right, so we should instead update the below except to only retry on that specific error it's checking for in the "if" and otherwise re-raise. To cover all cases, not just context exceeded, as we only want to retry one specific error. So the original logic needs to be changed as it's not quite correct

nemo_skills/inference/model/base.py
```diff
@@ -130,6 +148,11 @@ def _maybe_apply_stop_phrase_removal(
         if remove_stop_phrases:
             result["generation"] = trim_after_stop_phrases(result["generation"], stop_phrases)
 
+    @abc.abstractmethod
```
Kipok: this is not really abstract as by default it just disables the functionality and isn't required
```suggestion
```

nemo_skills/inference/model/base.py
```diff
@@ -138,10 +161,20 @@ def _build_chat_request_params(self, **kwargs) -> dict:
...
+            return request_params
+        else:
+            raise ValueError("Either prompt or messages must be provided")
+
+    @with_context_retry
     async def generate_async(
         self,
         prompt: str | list[dict],
-        tokens_to_generate: int = 2048,
+        tokens_to_generate: int | None = None,
```
Kipok, line 206: code execution model also needs this update for the default?

shtoshni: let me update it.

nemo_skills/inference/generate.py
```diff
@@ -265,7 +265,7 @@ def __init__(self, cfg: GenerateSolutionsConfig):
     def setup_llm(self):
         if self.cfg.code_execution:
             sandbox = get_sandbox(**self.cfg.sandbox) if self.cfg.sandbox is not None else None
-            llm = get_code_execution_model(**self.cfg.server, sandbox=sandbox)
+            llm = get_code_execution_model(**self.cfg.server, tokenizer=self.cfg.tokenizer, sandbox=sandbox)
```
Kipok: the full tokenizer should be like this. Maybe we can reassign this somewhere before? It's also used in setup_prompt this way

```python
self.cfg.tokenizer or self.cfg.server["model"]
```

nemo_skills/inference/model/base.py
```diff
@@ -130,6 +149,32 @@ def _maybe_apply_stop_phrase_removal(
         if remove_stop_phrases:
             result["generation"] = trim_after_stop_phrases(result["generation"], stop_phrases)
 
+    def _get_tokenizer(self, tokenizer: str | None) -> Union[ServerTokenizer, WrapperAutoTokenizer, None]:
+        """Get the tokenizer endpoint if available, otherwise initialize from tokenizer string"""
+        tokenizer_endpoint = self._get_tokenizer_endpoint()
```
Kipok, line 167: explicit tokenizer should always take precedence. If it's defined we should take it to avoid overriding default. If it's not defined, we can query the endpoint

nemo_skills/inference/model/utils.py
```diff
@@ -24,5 +33,54 @@ def trim_after_stop_phrases(text: str, stop_phrases: list[str]) -> str:
...
+    def __init__(self, model_name: str):
+        LOG.info(f"Initializing tokenizer from string: {model_name}")
+        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+    def encode(self, prompt: str | list[dict]) -> list[int]:
+        """Encode the prompt using the tokenizer."""
+        if isinstance(prompt, str):
+            return self.tokenizer.encode(prompt)
+        elif isinstance(prompt, list):
+            return self.tokenizer.apply_chat_template(prompt, add_generation_prompt=True)
```
Kipok, line 78: the `add_generation_prompt=True` is not always correct, maybe add a TODO or something. This function might be called in the middle of the generation (e.g. in TIR mode). But I guess it's not that big a deal as we'd just be subtracting a little more tokens than necessary

tests/gpu-tests/run_qwen.sh
```diff
@@ -1,16 +1,17 @@
 # will run all tests starting from only a HF checkpoint. Only requires 1 GPU.
 # also need to define HF_TOKEN for some of the tests
-# model needs to be inside /mnt/datadrive/nemo-skills-test-data/Qwen2.5-Math-7B-Instruct
+# model needs to be inside /mnt/datadrive/nemo-skills-test-data/Qwen3-4B
 # if you need to place it in a different location, modify test-local.yaml config
 # example: HF_TOKEN=<> ./tests/gpu-tests/run.sh
 set -e
 
-export NEMO_SKILLS_TEST_HF_MODEL=/mnt/datadrive/nemo-skills-test-data/Qwen2.5-Math-7B-Instruct
+export NEMO_SKILLS_TEST_HF_MODEL=/mnt/datadrive/nemo-skills-test-data/Qwen3-4B
```
Kipok: this won't be available on the CI machine by default, how about instead we switch to using the `Qwen/Qwen3-4B` here and not a local path? The current code always sets up HF_HOME, so it will only be downloaded one time and will make it simpler to switch models in the future.

tests/gpu-tests/test_context_retry.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def validate_environment(self):
+        """Validate required environment variables"""
+        if not self.model_path:
+            pytest.skip("Define NEMO_SKILLS_TEST_HF_MODEL to run this test", allow_module_level=True)
+        if not self.model_type:
+            pytest.skip("Define NEMO_SKILLS_TEST_MODEL_TYPE to run this test", allow_module_level=True)
+
+
+class CommandBuilder:
```
Kipok, line 54: No need to change now, but for the future, the same comment applies here as on slurm ci PR. I think this is a bit overengineered and makes it hard to add new tests and also understand what existing tests do. So generally I think for tests we should go for a simpler structure with just a command + check and duplication is not a big deal imo.

tests/gpu-tests/test_generate.py
```diff
@@ -26,11 +26,11 @@
 
 
 @pytest.mark.gpu
-def test_vllm_generate_greedy():
-    model_path = os.getenv('NEMO_SKILLS_TEST_HF_MODEL')
+def test_vllm_generate_greedy(server_type):
```
Kipok: is this accidental typo? I don't think this is being used

nemo_skills/inference/model/utils.py
shtoshni: Yeah, this won't matter except for maybe cost a few tokens. But better to be safe here.

tests/gpu-tests/run_qwen.sh
shtoshni: Good idea! I suggested Qwen3 because it's smaller and 4B should be faster than 7B

tests/gpu-tests/test_context_retry.py
shtoshni: Let me fix this in another PR. 
The test script was more vibe-coded, and vibe coding tends to overengineer :)

nemo_skills/inference/generate.py
shtoshni: Okay, reversed the order of setting prompt

PR #729, opened 2025-08-27
Add Formal Math Eval Docs

gwarmstrong:


Kipok (commented):
did some style fixes, so that docs are properly displayed.

I think our default evaluation setup here might actually need to be updated. The default prompt seems to be for non-reasoning models (asking to do lean code right away), we should probably change that. Also the **FINAL ANSWER** thing isn't even mentioned in that prompt, so we should probably either not use it by default or make consistent with the prompt.

We probably need to change the default setup and update the docs accordingly. Would be good to add an example evaluation command that can match results of DS-prover or Geodel prover

docs/evaluation/formal-math.md
```diff
@@ -38,6 +38,31 @@ Important configuration options and considerations for reliable Lean 4 evaluatio
 - Stateless execution for Lean 4
   - Evaluation is stateless; prior context is not persisted across items.
 
+## Sample launch command
+
+To reproduce [Goedel-Prover-V2](https://github.com/Goedel-LM/Goedel-Prover-V2) pass@32 results on minif2f, you can use the following command:
+
+```bash
+ns eval \
```
gwarmstrong, line 42: @stephencge are there any other benchmarks models it makes sense to expand on the appropriate parameter settings here? mostly with respect to the nemo-skills specific ones (like `extract_code_mode`)

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+Under the hood we assemble the final Lean 4 program and evaluate it in a sandboxed Lean 4 environment, capturing a proof status for each item and writing it back to the originating files.
+
+## How we execute and evaluate
+
+Evaluation constructs a complete Lean 4 file from the model output and dataset metadata, then uses a sandboxed Lean 4 checker to validate it.
+
+At a high level, for each JSONL line in your result manifests we do the following:
+
+1. Build the final Lean program to check.
+   All the parameters mentioned here are *evaluation* parameters which are controlled with `--extra_eval_args="++eval_config.<parameter_name>=<parameter_value>`
```
Kipok:
```suggestion
   All the parameters mentioned here are *evaluation* parameters which are controlled with `++eval_config.<parameter_name>=<parameter_value>`
```

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+      - Normalize the model output to keep only the intended Lean code. By default we extract the last Lean code block when multiple are present (controlled by `extract_code_mode` parameter, which defaults to `"last"`). You can change the split phrase with `final_answer_key` parameter. If the `final_answer_key` phrase isn't present, we use the full generation instead.
```
Kipok: should we disable final_answer_key logic by default? I think it's not relevant to latest models

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+
+Important configuration options and considerations for reliable Lean 4 evaluation:
+
+- Producing only the proof body
+  - Encourage the model to output only the proof body. If a theorem header is generated, it will be stripped and the dataset statement will be used.
+- Providing a preassembled proof
+  - If the line already includes a complete proof artifact, it will be used directly; otherwise the proof is assembled from the model’s generated text and dataset metadata.
+- `restate_formal_statement` (default: True)
+  - Controls whether the dataset’s `formal_statement` is inserted before the proof. Keeping this enabled enforces the canonical theorem; disabling it relies on the model’s emitted statement and is generally not recommended for benchmarking.
+- `timeout` (default: 30.0 seconds)
```
Kipok: should we increase default?

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+  - Per-item execution timeout. A timeout returns `proof_status='timeout'`.
+- Stateless execution for Lean 4
```
Kipok: I think this can be removed as it's assumed by everything above that there is just a single final block, so it's just a little confusing as to what prior context even means here

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+    --model=Goedel-LM/Goedel-Prover-V2-32B \
+    --server_gpus=8 \
+    --benchmarks=minif2f:32 \
+    --output_dir=/workspace/minif2f-pass32 \
+    --server_args="--max-model-len 40960" \
+    ++inference.temperature=1.0 \
+    ++inference.top_p=0.95 \
+    ++inference.tokens_to_generate=38912 \
+    --extra_eval_args="++eval_config.timeout=400" \
+    ++prompt_config=lean4/formal-proof-deepseek-prover-v2
```
Kipok: is this default? If not, we should probably make it default

Kipok (commented):
added a few comments - please also make sure to run `mkdocs serve` and check that the rendering on the website looks good

docs/evaluation/formal-math.md
```diff
@@ -1,6 +1,68 @@
...
+```
```
Kipok, line 65: Please also add the output of summarize-results table here so that it's very visible what users should expect to get

docs/evaluation/formal-math.md
stephencge: 30s is a good default (still low single digit % timeout on most benchmarks). Flagging it in this section good for user who wants to increase for finer evaluation environment control

docs/evaluation/formal-math.md
stephencge: extract_code_mode defaulting to last was added to match convention from DeepSeek/Goedel-Prover

docs/evaluation/formal-math.md
stephencge: yes it looks like that is default prompt_config in minif2f __init__.py

gwarmstrong (commented):
LGTM. Can't approve because I opened the PR, but I can help merge when ready.

PR #730, opened 2025-08-27
Fix SciCode evaluation

jubick1337:
Add missing imports, fix newlines, and update helper functions to use correct indexes


Kipok (approved):
thanks!

PR #733, opened 2025-08-27
HLE Fix

shtoshni:
Removing `no_answer` calculation for HLE.
Additionally, changing the answer key to the entire generation for HLE which will bring fix the counterintuitive majority numbers.

### DIFF
--- a/nemo_skills/evaluation/metrics/map_metrics.py
+++ b/nemo_skills/evaluation/metrics/map_metrics.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific lang
 
+import functools
+
 from nemo_skills.evaluation.metrics.answer_judgement_metrics import AnswerJudgementMetrics
 from nemo_skills.evaluation.metrics.arena_metrics import ArenaMetrics
 from nemo_skills.evaluation.metrics.bfcl_metrics import BFCLMetrics
@@ -29,7 +31,7 @@ from nemo_skills.evaluation.metrics.ruler_metrics import RulerMetrics
 
 METRICS_MAP = {
     "math": MathMetrics,
-    "hle": MathMetrics,  # Please see the `get_metrics` function where HLE is handled with specific parameters
+    "hle": functools.partial(MathMetrics, compute_no_answer=False, answer_key="generation"),
     "lean4-proof": Lean4Metrics,
     "lean4-statement": Lean4Metrics,
     "answer-judgement": AnswerJudgementMetrics,
@@ -50,8 +52,4 @@ METRICS_MAP = {
 def get_metrics(metric_type: str):
     if metric_type not in METRICS_MAP:
         raise ValueError(f"Metric {metric_type} not found.\nSupported types: {str(METRICS_MAP.keys())}")
-    if metric_type == "hle":
-        # HLE does not have "no answer" metric, so we set compute_no_answer to False
-        return METRICS_MAP[metric_type](compute_no_answer=False, answer_key="generation")
-    else:
-        return METRICS_MAP[metric_type]()
+    return METRICS_MAP[metric_type]()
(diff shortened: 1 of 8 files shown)


nemo_skills/evaluation/metrics/map_metrics.py
```diff
@@ -29,6 +29,7 @@
 
 METRICS_MAP = {
     "math": MathMetrics,
+    "hle": MathMetrics,  # Please see the `get_metrics` function where HLE is handled with specific parameters
```
Kipok: would functools.partial work here? I think it might support building a class with prefilled init arguments with later python versions

nemo_skills/evaluation/metrics/math_metrics.py
```diff
@@ -24,16 +24,22 @@
 
 class MathMetrics(BaseMetrics):
     # TODO: how can we ensure that user-defined aggregations have all the same metrics as in base?
+    answer_key: str = "predicted_answer"
```
Kipok: does it need to be defined here?

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
```diff
@@ -225,16 +225,13 @@ pass@1[avg-of-16] | 12032       | 4879       | 12516       | 81.44%           |
...
 
--------------------------------------------------- hle --------------------------------------------------
-evaluation_mode   | num_entries | avg_tokens | gen_seconds | judge_correct | symbolic_correct | no_answer
-pass@1[avg-of-16] | 2158        | 12111      | 7782        | 7.75%         | 2.40%            | 64.13%
-majority@16       | 2158        | 12111      | 7782        | 4.31%         | 3.43%            | 49.91%
-pass@16           | 2158        | 12111      | 7782        | 27.80%        | 10.10%           | 49.91%
+-------------------------------------------- hle --------------------------------------------
+evaluation_mode   | num_entries | avg_tokens | gen_seconds | judge_correct | symbolic_correct
+pass@1[avg-of-16] | 2158        | 12111      | 7782        | 7.75%         | 2.40%
+majority@16       | 2158        | 12111      | 7782        | 7.61%         | 2.34%
```
Kipok, line 231: so this majority is built on a string match of the full generation? Are there really some examples where they are fully identical? Or do I misunderstand something here?

Also if you're calculating majority, does it make sense to keep no_answer metric and just use the same thing you have for majority as the answer? Or maybe we remove both majority and no answer together?

nemo_skills/evaluation/metrics/map_metrics.py
shtoshni: oh interesting, didn't realize you could do it for class as well!

nemo_skills/evaluation/metrics/math_metrics.py
shtoshni: Because of this line - https://github.com/NVIDIA/NeMo-Skills/blob/50853feb2df67759859bad14777ebf1900b0d548/nemo_skills/evaluation/metrics/math_metrics.py#L89

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
shtoshni: Yes, basically generation string after the thinking part. I don't expect it to be the same almost ever, but it looked more fair than removing answers in the majority calculation because the `predicted_answer` field is null. 

For HLE, we can get rid of majority as well. In fact, I propose that we keep majority calculation a dynamic thing. So for any task, we should allow majority calculation. By default, for math, we can keep it on. For most other metrics, we can keep it off, but with the ability to override and turn it on if needed. Especially useful for GenSelect, where the same structure can be picked multiple times.

nemo_skills/evaluation/metrics/math_metrics.py
shtoshni: I think the `get_incorrect_sample` method doesn't have to be a class method. 
So, pushed a commit changing the class method to a regular method.
Then the L27 where we redefine `answer_key` is not needed as well.

docs/tutorials/posts/llama-nemotron-super-v1.5-evals.md
Kipok: ok, let's merge now, but I would probably disable majority here, I think it's confusing to have it as people would assume we are extracting the answer which we don't do. Probably good to add a parameter to disable it in math metrics and set that in hle case

Issue #577, opened 2025-07-11
Move server waiting logic inside generate task


Kipok:
We currently have a separate while loop waiting for the server to start before running a generation command https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/utils/server.py#L68. This is still useful for run_cmd with server args, but for our main generate / eval jobs we should move that logic inside GenerationTask https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/generate.py#L201. Main reason for doing that is to not waste time if we have multiple dependent jobs, but data is already fully finished. Current behavior will only check data and stop job after server is started, but with the new behavior it will finish immediately which is much better

Kipok:
done in #735

PR #735, opened 2025-08-28
Defer server wait for generation

smahdavi4:
Deferring server wait to right before generation to prevent server load time if the job is already done.
Resolving #577


Kipok (approved):
Thanks!

PR #739, opened 2025-08-28
Update evaluation numbers and a default split

Kipok:


Kipok:
@shtoshni do you want to re-generate the plots for openreasoning models with these new numbers? And push here as well as on HF?

shtoshni:
Looks good to me!
The SciCode numbers are in the AAI ballpark/even better now for both the models.

shtoshni (approved):

PR #747, opened 2025-08-29
Updates to file formatting

shtoshni:
This is just a quick file formatting part. The linter also identified a few errors. I'll fix those in a separate PR because they are a bit more time-consuming and require more testing. This is literally just the formatting part.


shtoshni:
The Lint and Format failures are due to deeper issues than formatting, so don't worry about it.

shtoshni:
yes, fixing those.

.pre-commit-config.yaml
```diff
@@ -39,7 +39,8 @@ repos:
     rev: v0.12.9
     hooks:
       - id: ruff-check
-        args: ["--fix"]
+      # F401 is the import error. nemo_skills/utils.py has certain imports that are not used in the file but used in files which import it.
+        args: ["--fix", "--per-file-ignores=__init__.py:F401,nemo_skills/utils.py:F401,nemo_skills/pipeline/cli.py:F401"]
```
activatedgeek: We should move these checks to `pyproject.toml`.

For instance: https://docs.astral.sh/ruff/settings/#lint_extend-per-file-ignores

activatedgeek (approved):

PR #748, opened 2025-08-29
Precommit fixes more complex

shtoshni:
The more complicated fixes for lint check


nemo_skills/training/start_grpo.py
```diff
@@ -98,7 +98,6 @@ def main(cfg) -> None:
     init_distributed(trainer, ptl_model, cfg.model.get("transformer_engine", False))
```
Kipok: Please remove this file completely, it's not used anymore

Kipok (approved):
thanks. Generally for files like those in conversion folder we don't necessarily want to enforce the codestyle as they mostly get copy-pasted from another place and it's harder to maintain when we change them like this. We can still do this now, but for the future, I think we might add exceptions to cases like these.

nemo_skills/pipeline/train.py
```diff
@@ -96,12 +96,15 @@ def get_cmd(params: TrainingParams) -> str:
     TrainingAlgo.rm: "rm_config",
 }
 
-rl_extra_args_fn = lambda params: (
-    f" ++model.data.data_prefix.train='[{params.training_data}]' "
-    f" ++model.data.data_prefix.validation='[{params.validation_data}]' "
-    f" ++model.data.data_prefix.test='[{params.validation_data}]' "
-    f" pretrained_checkpoint.restore_from_path={params.nemo_model} " + params.extra_arguments
-)
+
```
activatedgeek, line 99: String concatenation is a little too error prone with whitespaces, like here.

In such cases, I personally like using either multiline strings as:
```python
import textwrap

multiline_str = textwrap.dedent(f"""
# all my strings/variables as a single blob
""")
```

or, if we are working with shell arguments, then it is better to write them in the `execv` format which for our case is effectively just a list of strings, as

```python
multiarg_str = " ".join([
  "argA=valA",
  "argB=valB",
   ...
])
```

I think the `execv` format is much easier to maintain and less error prone in future changes.

pyproject.toml
```diff
@@ -74,3 +74,4 @@ extend-select = ["I"]
 "__init__.py" = ["F401"]
 "nemo_skills/utils.py" = ["F401"]
 "nemo_skills/pipeline/cli.py" = ["F401"]
+"nemo_skills/training/gpt_sft_dataset.py" = ["E402"]
```
activatedgeek: I think this list of ignore rules will start getting out of hand and make `pyproject.toml` cluttered, unless the rules have more wildcards instead of specific files.

Perhaps, we can move all these ignores to file level ignores by using such a comment at the top of a file:
```
# ruff: noqa: E402
```

shtoshni: Okay, that's a good idea. Let me do that.

nemo_skills/pipeline/train.py
shtoshni: That's a good point. Unfortunately, this format is followed throughout the codebase. But let's keep this in mind, and try to revert to the `execv` format whenever we have the time.

PR #751, opened 2025-08-29
Add MMLU-Pro-X

shuoyangd:
This PR adds MMLU-Pro-X dataset. It attempts to match the prompts used in [lm-evaluation-harness](https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/mmlu_prox) to the maximum extend possible such that we can mostly reproduce the numbers in the [paper](https://arxiv.org/pdf/2503.10497) and [public leaderboard](https://huggingface.co/spaces/TIGER-Lab/MMLU-Pro).

With GPT-OSS we got 79.6 on English while the public leaderboard shows 80.8.

Pending test on Mistral-Small-2506 to compare with the paper. Will remove draft label once it's available.


shuoyangd:
I did some tests. Here are the results:

| Model              | Type   |   en | de   | es   | fr   | it   | ja   | ru   |
|:-------------------|:-------|-----:|:-----|:-----|:-----|:-----|:-----|:-----|
| gpt-oss-120b       | Public | 80.8 | -    | -    | -    | -    | -    | -    |
| gpt-oss-120b       | Nemo-Skills  | 79.6 | 73.8 | 74.3 | 76.1 | 75.4 | 74.6 | 72.9 |
| mistral-3.1-small  | Public | 62   | 58.5 | 59.4 | 60.6 | 59.6 | 54.4 | 59.2 |
| mistral-3.1-small  | Nemo-Skills  | 68.4 | 59.7 | 61.6 | 62.9 | 63.7 | 61.1 | 59.2 |
| qwen3-32b-thinking | Public | 74.9 | 71.7 | 72.8 | 72.1 | 73.5 | 70.2 | 69.1 |
| qwen3-32b-thinking | Nemo-Skills  | 73.7 | 69.4 | 70.6 | 74.2 | 75.1 | 75.0 | 76.5 |

Numbers are mostly in the same range, though I found quite a bit variations goes into (1) non-format-conforming answers (2) inference parameters. A lot of information is not fully transparent, so I don't know how far we should go to make sure that we perfectly match the specific results in the paper.

I think this is ready for merging. Do we want to add a test specifically for this dataset?

nemo_skills/dataset/mmlu-prox/lang_libs.py
```diff
@@ -0,0 +1,704 @@
+# Ported from https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/tasks/mmlu_prox/lang_libs.py
```
Kipok: can we download this file on the fly if it's identical? So can download inside prepare.py and use, same as what we do for datasets

nemo_skills/evaluation/evaluator/mcq.py
```diff
@@ -45,6 +45,10 @@ def extract_letter(text):
             data = [json.loads(line) for line in fin]
         with open(file, "wt", encoding="utf-8") as fout:
             for sample in tqdm(data):
-                sample["predicted_answer"] = extract_letter(sample["generation"])
+                extract_from_boxed = sample.get("extract_from_boxed", "True") == "True"
```
Kipok: why don't we use True / False as actual values, not strings? json / omegaconf all support this and feels a bit less error-prone that way. So this applies both here and inside prepare.py

Kipok:
> Do we want to add a test specifically for this dataset?
I think it's a good idea to do that, otherwise hard to ensure things are always working correctly. Ideally we use some small model and / or a subset of the dataset (you can control that with `++max_samples`) to make test run a bit faster. We should add it as another slurm test in here https://github.com/NVIDIA/NeMo-Skills/tree/main/tests/slurm-tests similar to 49b eval. 

We can do it in another pr. And also we should add this to docs (ideally in this pr)

nemo_skills/dataset/mmlu-prox/lang_libs.py
shuoyangd: Updated to download from lm-evaluation-harness

nemo_skills/evaluation/evaluator/mcq.py
shuoyangd: Updated to use boolean literal instead of string

docs/evaluation/multilingual.md
```diff
@@ -0,0 +1,74 @@
...
+```
+ns eval \
+    --cluster=[cluster] \
+    --model=openai/gpt-oss-120b \
+    --benchmarks mmlu-prox \
+    --output_dir=[output dir] \
+    --num_chunks=16 \
+    --server_type=vllm \
+    --server_gpus=4 \
+    --server_args='--tensor-parallel-size 4 --async-scheduling' \
```
Kipok:
```suggestion
    --server_args='--async-scheduling' \
```
that's a default

docs/evaluation/multilingual.md
```diff
@@ -0,0 +1,74 @@
...
+```
+ns eval \
+    --cluster=[cluster] \
+    --model=mistralai/Mistral-Small-3.1-24B-Instruct-2503 \
+    --benchmarks mmlu-prox \
+    --output_dir=[output dir] \
+    --server_type=vllm \
+    --num_chunks=16 \
+    --server_gpus=2 \
+    --server_args='--tokenizer-mode mistral --config-format mistral --load-format mistral --tensor-parallel-size 2' \
```
Kipok:
```suggestion
    --server_args='--tokenizer-mode mistral --config-format mistral --load-format mistral' \
```

docs/evaluation/multilingual.md
```diff
@@ -0,0 +1,74 @@
...
+```
+ns eval \
+    --cluster=[cluster] \
+    --model=Qwen/Qwen3-32B \
+    --benchmarks mmlu-prox \
+    --output_dir=[output dir] \
+    --server_type=vllm \
+    --num_chunks=32 \
+    --server_gpus=2 \
+    --server_args='--tensor-parallel-size 2' \
```
Kipok:
```suggestion
```

Kipok (commented):
let me push a small update to the docs as currently this page isn't exposed in the left panel

Kipok (approved):

PR #752, opened 2025-08-29
update nemo-rl to latest main

wedu-nvidia:
Updated nemo-lr docker file


Kipok (approved):

Kipok:
updated container in-place everywhere

PR #757, opened 2025-09-01
Add standard deviation metrics for benchmark variance analysis

AdamRajfer:
# Add standard deviation metrics for benchmark variance analysis

This PR implements comprehensive standard deviation metrics to provide variance analysis directly within existing evaluation modes as additional columns, helping assess evaluation reliability and consistency across multiple benchmark runs.

### 📊 **New Integrated Metrics**
- **`{metric_name}_std_across_runs`**: Standard deviation of metric values across max_k benchmark runs
- **`{metric_name}_avg_sample_std`**: Average of per-sample standard deviations

### 🎯 **Universal Metric Support**
- Works with **all metric types**: `symbolic_correct`, `no_answer`, `false_positives`, `correct_judgements`, etc.
- Generic implementation supports any performance measure, not limited to correctness rates

### Integrated Column Structure
```json
{
  "pass@1[avg-of-4]": {
    "symbolic_correct": 80.44,
    "symbolic_correct_std_across_runs": 0.112,
    "symbolic_correct_avg_sample_std": 0.327,
    "no_answer": 6.31,
    "no_answer_std_across_runs": 0.038,
    "no_answer_avg_sample_std": 0.184
  }
}
```

### Core Implementation
- **`BaseMetrics._add_benchmark_std_metrics()`**: New method that adds std dev columns to existing evaluation modes
- **All metrics classes**: Enhanced with std dev column support and formatting
- **Test suite**: Comprehensive 35+ test cases covering all implementation scenarios  
- **Reference files**: Updated with new decimal format examples

## Testing

- **35+ comprehensive tests** covering edge cases and integration scenarios
- **End-to-end validation** with real metrics classes 
- **Reference file updates** for all test scenarios
- **Cross-compatibility testing** with all existing metrics implementations


nemo_skills/evaluation/metrics/base.py
```diff
@@ -273,7 +347,7 @@ def _compute_pass_at_k(
             is_binary_score = (max(scores_list) == 1) and (min(scores_list) == 0)
 
             if is_binary_score:
-                total_correct = sum(scores_list)
+                total_correct = int(sum(scores_list))
```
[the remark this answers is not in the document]
AdamRajfer: Applying this change introduces the following failures in the tests:

```
FAILED tests/test_benchmark_std_dev.py::TestPassAtKIntegration::test_score_collection - TypeError: 'float' object cannot be interpreted as an integer
FAILED tests/test_benchmark_std_dev.py::TestPassAtKIntegration::test_score_accumulation - TypeError: 'float' object cannot be interpreted as an integer
FAILED tests/test_benchmark_std_dev.py::TestEndToEndIntegration::test_complete_flow - TypeError: 'float' object cannot be interpreted as an integer
```

It is related to the fact that later in the code we run:

```python
prob_all_incorrect = math.comb(total_incorrect, k) / math.comb(total, k)
```

and `math.comb` requires integer as a first argument.

nemo_skills/evaluation/metrics/base.py
```diff
@@ -41,8 +42,80 @@ def get_metrics(self):
                     metrics_dict[agg_mode][metric_key] = 100.0 * metric_value / self.total
                 else:
                     metrics_dict[agg_mode][metric_key] = metric_value
+        self._add_benchmark_std_metrics(metrics_dict)
         return metrics_dict
 
+    def _add_benchmark_std_metrics(self, metrics_dict):
+        """Add standard deviation metrics as columns to existing evaluation modes.
```
Kipok: Thanks for the nice and clear docstring!

nemo_skills/evaluation/metrics/base.py
```diff
@@ -41,8 +42,80 @@ def get_metrics(self):
...
+        for score_method, k_dict in self.all_scores.items():
+            if sample_list := k_dict.get(self.max_k):
+                # Calculate benchmark run std dev
+                run_scores = [[] for _ in range(self.max_k)]
+                for sample_scores in sample_list:
+                    for run_idx, score in enumerate(sample_scores):
+                        run_scores[run_idx].append(score)
+
+                run_averages = [sum(scores) / len(scores) for scores in run_scores]
+                benchmark_std = statistics.stdev(run_averages) if len(run_averages) > 1 else 0.0
```
Kipok: you can also use numpy if you want, might make this a bit shorter and faster as well. Doesn't matter too much, but it's installed if you need it

nemo_skills/evaluation/metrics/base.py
```diff
@@ -41,8 +42,80 @@ def get_metrics(self):
...
+            - Run 4: [0,0,1] → avg 0.3333
+            → std_across_runs = stdev([0.6667, 0.6667, 0.6667, 0.3333]) ≈ 0.1925
+
+            Sample std devs:
+            - Sample 1: stdev([1,0,1,0]) ≈ 0.5773
+            - Sample 2: stdev([1,1,0,0]) ≈ 0.5773
+            - Sample 3: stdev([0,1,1,1]) ≈ 0.5000
+            → avg_sample_std = (0.5773 + 0.5773 + 0.5000) / 3 ≈ 0.5515
+        """
+        if self.max_k <= 1:
```
Kipok: should we do it for all k, not just for max? That would make it consistent with other reported metrics and allow to easily plot how things change as we increase k (e.g. we can plot mean and add error pars with std). Should hopefully be as easy as adding k as a parameter and moving call of this method inside the loop

nemo_skills/evaluation/metrics/base.py
```diff
@@ -41,8 +42,80 @@ def get_metrics(self):
...
+            return
+
+        for score_method, k_dict in self.all_scores.items():
+            if sample_list := k_dict.get(self.max_k):
```
Kipok: I think max_k should always be in that dict, if it's not it indicates some issue. So we can probably remove the if and just use it directly. Let me know if I'm missing some case when it expects not to have max_k in

nemo_skills/evaluation/metrics/base.py
AdamRajfer: Nice catch, switched to np!

nemo_skills/evaluation/metrics/base.py
AdamRajfer: Agreed, added for all ks!

nemo_skills/evaluation/metrics/base.py
AdamRajfer: You're right. I fixed it and now for all ks we are asserting the existence and length of sample scores.

nemo_skills/evaluation/metrics/base.py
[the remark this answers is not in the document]
Kipok: @shtoshni is there any problem making this int(...)? Not fully sure about this part, the is_binary_score check above is a little strange as it might accept things like `[0, 0.5, 1]`, but if the intention was to check only 0/1, I guess it should be ok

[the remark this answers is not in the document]
shtoshni: @Kipok @AdamRajfer - Actually, ideally the check should be if all entries are 0 or 1 to decide if it's a `is_binary_score`.  Then making it `int` is completely fine. I was having issues with the Instruction-Following tasks where the score entries can be higher than 1. I don't know the scoring for all the tasks, so we should just fix the loophole that @Kipok pointed out to be safe for future tasks.

Kipok (approved):
thanks!

nemo_skills/evaluation/metrics/base.py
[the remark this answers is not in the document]
Kipok: updated the check there to assert all are (0, 1, True, False)

PR #758, opened 2025-09-02
Server container now can be passed with CLI

vmendelev:
This commit allows to pass container image path or uri with CLI tool, thus enabling support for more flexible server configuration without touching cluster configs.


nemo_skills/pipeline/utils/exp.py
```diff
@@ -433,9 +433,13 @@ def add_task(
     executors = []
     # assuming server always has the largest resources request, so it needs to go first
     if server_config is not None and server_config["num_gpus"] > 0:
-        server_cmd, num_server_tasks = get_server_command(**server_config, cluster_config=cluster_config)
-        if "container" not in server_config:
-            server_container = cluster_config["containers"][server_config["server_type"]]
+        # do not pass container into the command builder
+        _server_cfg = {k: v for k, v in server_config.items() if k != "container"}
```
Kipok: would it be a bit simpler and more explicit to do
```suggestion
        server_container = server_config.pop('container', cluster_config["containers"][server_config["server_type"]])
```
and then remove the assignment below?

nemo_skills/pipeline/eval.py
```diff
@@ -92,6 +92,9 @@ def eval(
         help="Path to the entrypoint of the judge server. "
         "If not specified, will use the default entrypoint for the server type.",
     ),
+    server_container: str = typer.Option(
```
Kipok, line 95: Please also add to run_cmd and start_server

Kipok (commented):
thanks, just a few small changes

Kipok:
also tests are failing because this argument isn't passed somewhere

nemo_skills/pipeline/utils/exp.py
vmendelev: Good idea.

Kipok (approved):

PR #759, opened 2025-09-02
add lr scheduler for nemo-rl sft with fsdp as backend

wedu-nvidia:
add lr scheduler for Nemo-RL SFT with fsdp as backend


nemo_skills/training/nemo_rl/configs/sft.yaml
```diff
@@ -136,6 +136,18 @@ policy:
       foreach: False
       fused: False
 
+  scheduler:
```
Kipok, line 142: we should ideally set the default to be no warmup and no decay, but let it be easily configurable with command line arguments (ideally using the same parameters as in megatron, so they are always consistent and users can just switch a backend and all parameters will transfer).

also I guess there is no way to automatically set T_max here? So we would always have to explicitly be setting it to the same as steps value?

wedu-nvidia: Ok, I will combine these parameters for both  backend. 
I think I can set T_max = max_steps - warmup_steps, then users do not have to set it explicitly

Kipok: the only problem with that is if num_epochs is set instead of num_steps, then it doesn't work. But maybe it's still a good default

wedu-nvidia: yes, I agree.

nemo_skills/training/nemo_rl/configs/sft.yaml
```diff
@@ -127,15 +129,28 @@ policy:
...
+    - name: "torch.optim.lr_scheduler.LinearLR"
+      kwargs:
+        start_factor: 1.0
+        end_factor: 1.0
+        total_iters: 1  # must be >=1
+    - name: "torch.optim.lr_scheduler.ConstantLR"
```
Kipok: instead of using contstant, can we use cosine here and just set the min = initial for lr by default? That way it will be the same as constant, but we can still customize the schedule in the same way as for megatron? Ot does it not work?

Kipok: I know we discussed this, but after thinking some more, I guess I still don't fully get why we can't use something like this

```
- name: "torch.optim.lr_scheduler.LinearLR"
      kwargs:
        start_factor: 1.0
        end_factor: 1.0
        total_iters: 1  # must be >=1
    - name: "torch.optim.lr_scheduler.CosineLR"
       kwargs:
        < ... min_lr and any other required params here to match megatron setup>
        T_max: ${sft.max_num_steps} 
    - milestones: [0]
```
and by default set min_lr = initial_lr, so that it's a constant schedule unless user changes things

wedu-nvidia: yes, I think you are right. We can make min_lr = initial_lr so that it is always a constant schedule. I will fix it today

nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -49,6 +49,8 @@ policy:
   tensor_model_parallel_size: 1
   pipeline_model_parallel_size: 1
   context_parallel_size: 1
+  lr: 1e-6
+  weight_decay: 0.01
```
Kipok, line 53: Can you please also move min_lr over here? And change the commands in here https://github.com/NVIDIA/NeMo-Skills/blob/main/recipes/openmathreasoning/scripts/simplified_recipe.py#L159 and here https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/tutorials/posts/omr-simple-recipe.md?plain=1#L249

wedu-nvidia: Done

docs/tutorials/posts/omr-simple-recipe.md
```diff
@@ -246,7 +246,7 @@ ns nemo_rl sft \
     ++policy.train_global_batch_size=32 \
     ++policy.megatron_cfg.tensor_model_parallel_size=4 \
     ++policy.megatron_cfg.context_parallel_size=2 \
-    ++policy.megatron_cfg.optimizer.lr=1e-4 \
+    ++policy.lr=1e-4 \
```
Kipok:
```suggestion
    ++policy.lr=1e-5 \
```

PR #760, opened 2025-09-02
Add bootstrapping

smahdavi4:
* Adds bootstrapping (50X) for metric confidence.

We might need to run them multi-process depending on how large summaries are, but keeping it single-process for now.


Kipok:
this overlaps with https://github.com/NVIDIA/NeMo-Skills/pull/757

@smahdavi4 @AdamRajfer what do you think is the best way to unify the two versions? Should we do bootstrapping for across samples variance and simple std for in-sample one? Or support std with/without bootstrapping with some parameter?

smahdavi4:
Sorry I just saw this PR. I modified this PR a bit to remove the intra-sample bootstrapping. I think both of our PRs serve a similar purpose, @AdamRajfer 's implementation is more efficient, but it does not cover maj@k and  pass@k and non-score metrics (such as precision/recall/f1 scores for answer judgement). 

My use cases are mainly precision/recall/f1 in answer-judgement. I am ok with merge either pull requests as long we have support for those.

PR #766, opened 2025-09-03
disable validation for nemo-rl if no validation data is provided

wedu-nvidia:
Disable validation if no validation data is provided. 
Also add another para in config for grpo to avoid error with latest nemo-rl


Kipok (approved):
thanks

PR #767, opened 2025-09-03
Natural Language math docs

avem-nv:
- Rearranged the sections for better readability
- Example command for how to add judge parameters eval() command in python interface
- Example command for how to add judge parameters via ns eval command line interface
- Example command line directly in dataset definition


avem-nv:
Just reviewed the changes and looks good to me. We can merge. @Kipok

PR #772, opened 2025-09-04
MCP interface updates

gwarmstrong:
### Summary
- Replace YAML-based tool config with module-based providers loaded via `++tool_modules`.
- Replace client-centric `MCPClientManager` with a tool-centric `ToolManager` keyed by Tool class name.
- Add `++tool_overrides.<ToolClass>.*` CLI overrides with a minimal, consistent shape.
- Prefer double-colon locator syntax for both python modules and `.py` files; single-colon remains supported for backward compatibility.
- Provide `MCPClientTool` base to minimize boilerplate for new tools.
- Add `extra_args` path to send server-only arguments after sanitization (without disabling sanitization).
- Default `PythonTool` to Hydra-based init hook, send timeout via `extra_args`.

### Details

- **Module-based tools**
  - New `ToolManager` loads tools from `++tool_modules=[module.path:ClassName,/abs/path/file.py:ClassName]`.
  - Tools implement `Tool` interface; qualified tool names are `ClassName.tool`.
  - Overrides are keyed by class name: `++tool_overrides.<ClassName>.<key>=...`.

- **Manager change (client-centric → tool-centric)**
  - We moved from a client-centric manager (which grouped tools under clients) to a tool-centric manager (which registers and routes tools directly by class name).
  - Rationale:
    - Simplifies overrides: `++tool_overrides.<ToolClass>.*` maps directly to the tool users care about (e.g., `PythonTool.exec_timeout_s`).
    - Uniform composition: mix built-in and user-defined tools (including file-based) with consistent naming: `ToolClassName.tool`.
    - Cleaner sanitization: providers can pass server-only args via `extra_args` without affecting model-visible schemas.
  - YAML tool config and client-centric registration are not used in the current pipeline

- **Client boilerplate reduction**
  - `nemo_skills.mcp.tool_providers.MCPClientTool` handles:
    - client construction via `client` and `client_params`
    - output formatting and init hooks (including `init_hook="hydra"`)

- **Locator syntax**
  - `locate()` now accepts double-colon for both modules and files:
    - Module: `nemo_skills.mcp.servers.python_tool::PythonTool`
    - File: `/nemo-skills-recipes/mcp-generation/example_tool.py::ExampleTool`

- **Sanitization and extra_args**
  - `MCPStdioClient.call_tool(tool=..., args=..., extra_args=...)` merges `extra_args` AFTER sanitization.
  - Allows sending hidden/internal server-only args without disabling sanitization.

- **Structured output**
  - StdIO client returns `structuredContent` when present; otherwise falls back to JSON-parsed content text.
  - Example stdio server returns a dataclass type to populate `structuredContent`.

### Updated/Built-in Tools

- **PythonTool** (`nemo_skills.mcp.servers.python_tool:PythonTool`)
  - Introduced self-contained tool class for easy default configuration
  - Execution now passes timeout via `extra_args` (post-sanitize), not in main args.

- **ExaTool** (stdio) and **ExaMCPTool** (HTTP)
  - Introduced self-contained tool classes for easy default configuration
  - Both forward `extra_args` through to the client.

#### Minimal (Python tool only with overrides)
```bash
++tool_modules=[nemo_skills.mcp.servers.python_tool::PythonTool]
++tool_overrides.PythonTool.exec_timeout_s=30
```

#### Python + Example tool (file path via single colon)
```bash
++tool_modules=[nemo_skills.mcp.servers.python_tool::PythonTool,/nemo-skills-recipes/mcp-generation/example_tool.py::ExampleTool]
```

#### Exa hosted MCP
```bash
++tool_modules=[nemo_skills.mcp.servers.exa_tool::ExaMCPTool]
++tool_overrides.ExaMCPTool.enabled_tools=["exa_web_search"]
```

### Breaking Changes
- YAML-based tool configs and client-centric registration are no longer used in tool-calling pipeline paths.
- Overrides are now keyed by `Tool` class name (e.g., `PythonTool`, `ExaTool`) instead of provider IDs.

### Migration Guide
- Replace any `++tool_config=...` usage with `++tool_modules=[...]` and `++tool_overrides.<ToolClass>.*`.
- If you previously relied on hidden args (e.g., server-only params), provide them via `extra_args` (handled internally by providers) instead of disabling sanitization.


tests/test_mcp_clients.py
```diff
@@ -459,73 +481,4 @@ async def __aexit__(self, exc_type, exc, tb):
...
-
-    manager = build_client_manager(cfg)
-    client = manager.get_client("http")
-    assert isinstance(client, MCPStreamableHttpClient)
-
-    # init_hook should have modified base_url to include API key
-    assert client.base_url.endswith("?exaApiKey=KEY123")
-    # output formatter should be resolved callable
-    assert client.output_formatter is utils_mod.exa_output_formatter
+# Removed tests relying on build_client_manager; tool-centric flow supersedes client-centric config
```
gwarmstrong:
```suggestion
```

nemo_skills/inference/model/tool_call.py
```diff
@@ -66,7 +79,8 @@ async def _execute_tool_call(self, tool_call):
 
         ## TODO(sanyamk): Only exceptions related to tool execution here, all others must fail.
         try:
-            result = await self.client_manager.execute_tool(tool_name, tool_args)
+            # Allow providers to specify extra_args behavior internally if needed in the future
+            result = await self.tool_manager.execute_tool(tool_name, tool_args, extra_args=None)
```
activatedgeek, line 83: Perhaps we can extend the signature of `_execute_tool_call` to take in `extra_args`?

nemo_skills/inference/generate.py
```diff
@@ -136,8 +136,9 @@ class GenerateSolutionsConfig:
     # genselect config
     online_genselect_config: OnlineGenSelectConfig = field(default_factory=OnlineGenSelectConfig)
 
-    ## FIXME(sanyamk): Rethink the structure of this configuration.
-    tool_config: str | None = None  # Path to tool configuration file.
+    # Module-based tool configuration
+    tool_modules: list[str] | None = None  # List of tool module specs (e.g., module.path:ClassName)
```
activatedgeek: Can we add a longer comment here for all possible ways to define the tool modules, i.e. `module.path`, and absolute path?

nemo_skills/inference/generate.py
```diff
@@ -294,10 +295,11 @@ def setup_llm(self):
 
         if self.cfg.code_execution:
             llm = get_code_execution_model(**self.cfg.server, tokenizer=self.tokenizer, sandbox=self.sandbox)
-        elif self.cfg.tool_config:
+        elif getattr(self.cfg, "tool_modules", None):
```
activatedgeek: Isn't `self.cfg.tool_modules` always a valid attribute?

nemo_skills/inference/generate.py
```diff
@@ -294,10 +295,11 @@ def setup_llm(self):
...
             llm = get_tool_calling_model(
                 **self.cfg.server,
-                tool_config=self.cfg.tool_config,
+                tool_modules=getattr(self.cfg, "tool_modules", None),
```
activatedgeek: Same here. Wouldn't these always be accessible attributes?

nemo_skills/mcp/servers/exa_tool.py
```diff
@@ -76,5 +78,47 @@ def main():
...
+# ==============================
+# Module-based tool implementation
+# ==============================
+
+
+class ExaTool(MCPClientTool):
+    def __init__(self) -> None:
+        super().__init__()
+        # Defaults for stdio Exa server launch using explicit client class
+        self._config.update(
```
activatedgeek: I like this unified default config approach!

Nit: I think we should not directly update private variables. For now, `MCPClientTool` does not have much logic in the constructor, so not that big of a problem. 

In case the class grows, it might be better to have a separate method that updates the private variables, which can be called from the constructor. This setup will allow the flexibility in case the structure of this configuration changes since all the updated logic and intended side-effects can be handled within such a config update method. Direct access to the dictionary will require that all other side-effects also be taken care of by the inheriting class, and therefore might lead to repetitive code.

nemo_skills/inference/model/tool_call.py
gwarmstrong: I think the best way to pass extra_args is still not super clear (e.g., for managing session ids with stateful python)--so this is just to mark that this is where you would pass such things. I think we might need a slightly larger refactor when we get here, so I'm not sure it's worth passing it up the stack yet.

nemo_skills/inference/generate.py
gwarmstrong: Updated in 5e3f82160750fc96dacfbcc405b793e6e6b7101f

nemo_skills/inference/generate.py
gwarmstrong: Updated in 5e3f82160750fc96dacfbcc405b793e6e6b7101f

nemo_skills/inference/generate.py
gwarmstrong: added in https://github.com/NVIDIA/NeMo-Skills/pull/772/commits/a263f18d1d202e97de7af6a7ff591099e828d071

nemo_skills/mcp/servers/exa_tool.py
gwarmstrong: updated in https://github.com/NVIDIA/NeMo-Skills/pull/772/commits/53c2973e8961d1d89cef72f10498781cb3338bae

activatedgeek (approved):

PR #778, opened 2025-09-05
Fix OpenHands patch issue & add more info to SWE-bench docs

ludwig-n:
This fixes an issue where patches produced by OpenHands occasionally couldn't be applied during evaluation because they were missing a trailing newline. Also expands the SWE-bench docs with more info.

### DIFF
--- a/docs/evaluation/code.md
+++ b/docs/evaluation/code.md
@@ -138,8 +138,30 @@ ns eval \
 ```
 replacing <...> with your desired parameters.
 
+After all jobs are complete, you can check the results in `<OUTPUT_DIR>/eval-results/swe-bench/metrics.json`. They should look something like this:
+```
+{
+  "swe-bench": {
+    "pass@1": {
+      "num_entries": 500,
+      "gen_seconds": 7172,
+      "issues_resolved": 50.0,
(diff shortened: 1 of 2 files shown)


docs/evaluation/code.md
```diff
@@ -138,9 +138,31 @@ ns eval \
...
+      "num_entries": 500,
+      "gen_seconds": 7172,
+      "issues_resolved": 50.0,
+      "no_patch": 0.2,
+      "patch_cant_apply": 2.0
+    }
+  }
+}
+```
+Keep in mind there is some variance between runs, so we recommend running evaluation multiple times and averaging out the resolve rate.
```
Kipok, line 155:
```suggestion
Keep in mind there is some variance between runs, so we recommend running evaluation multiple times and averaging out the resolve rate. To do that automatically you can set `--benchmarks=swe-bench:N` where N is your desired number of repeats.
```

Kipok (approved):
thanks!

docs/evaluation/code.md
Kipok: @ludwig-n please update the docs after we fix running multiple in parallel

PR #781, opened 2025-09-05
Fix async tool registration

gwarmstrong:
* Fixes issue where multiple workers register the same tool resulting in an error
* Made tool listing thread safe, so only the first worker registers a tool
* Removed outdated comments in clients

### DIFF
--- a/nemo_skills/mcp/clients.py
+++ b/nemo_skills/mcp/clients.py
@@ -23,21 +23,24 @@ from mcp.client.streamable_http import streamablehttp_client
 
 
 def _process_hide_args(result, hide_args):
-    if hide_args:
-        output = []
-        for entry in result:
-            method_name = entry.get("name")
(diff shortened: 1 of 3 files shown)


activatedgeek (commented):
Should we also remove the folder `nemo_skills/mcp/config` now?

nemo_skills/mcp/clients.py
```diff
@@ -467,7 +385,7 @@ class MCPStdioClient(MCPClient):
     def __init__(self, command: str, args: list[str] | None = None):
         if args is None:
             args = []
-        # Default: inherit the caller's environment for all stdio-launched servers
+        # Inherit parent environment by default so servers can see test-set env vars
```
gwarmstrong:
```suggestion
        # Default: inherit the caller's environment for all stdio-launched servers
```

Issue #584, opened 2025-07-11
Remove duplication from log messages


Kipok:
The messages that our pipelines print are sometimes duplicated two times. Need to find the reason and fix it

PR #782, opened 2025-09-05
Potential solution to logging duplication

shtoshni:
https://github.com/NVIDIA/NeMo-Skills/issues/584


shtoshni:
So I didn't see the duplicate logs when I played around with this branch. 
But duplicate logs are somewhat random for me.

PR #785, opened 2025-09-05
updated config for sft data prep

wasiahmad:
This PR changes the SFT config yaml file for code sft data preparation.

### DIFF
--- a/nemo_skills/training/data_preparation_utils/config/code_sft.yaml
+++ b/nemo_skills/training/data_preparation_utils/config/code_sft.yaml
@@ -10,10 +10,8 @@ exclude_optional_keys: true # exclude keys other than input, output and expected
 metadata: null  # can provide additional metadata to store (e.g. dataset or generation_type)
 skip_first: 0  # useful for skipping validation set from train_full generation (it's always first)
 
-downsampling_method: null  # fair, random or null
 random_seed: 42
 do_shuffle: true
-num_output_samples: null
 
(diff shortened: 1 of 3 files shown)


nemo_skills/training/data_preparation_utils/config/code_sft.yaml
```diff
@@ -24,11 +24,22 @@ assistant_end: null  # can be used to not use jinja and just add this suffix
 
 code_tags: null
 
+# input params
+min_input_length: 0
+max_input_length: 1024
```
Kipok: probably a good idea to not set this by default and require to be set by user. Both here and in math config. Other than that, lgtm

Kipok (approved):

nemo_skills/training/data_preparation_utils/config/code_sft.yaml
titu1994: Agreed. Set it to none / null by default

wasiahmad: Done.

titu1994 (approved):
Thanks !

Kipok (commented):
please add `++do_shuffle=false` here https://github.com/NVIDIA/NeMo-Skills/blob/main/tests/test_data_preparation.py#L110

Kipok:
also can you please make the same change to default length parameters for the other two configs?

Issue #773, opened 2025-09-04
if server.enable_soft_fail=True is set for unsupported server (e.g. trtllm), job should fail with error


Kipok:

shtoshni:
Solved in - https://github.com/NVIDIA/NeMo-Skills/pull/786

PR #786, opened 2025-09-05
TRTLLM  + Soft Fail

shtoshni:
Adding the basic, no-retry version of soft fail for TRTLLM. 
Resolves this - https://github.com/NVIDIA/NeMo-Skills/issues/773

### DIFF
--- a/nemo_skills/inference/model/__init__.py
+++ b/nemo_skills/inference/model/__init__.py
@@ -53,6 +53,9 @@ models = {
 def get_model(server_type, tokenizer=None, **kwargs):
     """A helper function to make it easier to set server through cmd."""
     model_class = models[server_type.lower()]
+    if server_type == "trtllm" and kwargs.get("enable_soft_fail", False):
+        if kwargs.get("context_limit_retry_strategy", None) is not None:
+            raise ValueError("context_limit_retry_strategy is not supported for trtllm")
     return model_class(tokenizer=tokenizer, **kwargs)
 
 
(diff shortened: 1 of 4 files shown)


shtoshni:
Tests are passing in my local setup.

tests/gpu-tests/run_qwen.sh
```diff
@@ -2,13 +2,16 @@
 # also need to define HF_TOKEN for some of the tests
 set -e
 
-export NEMO_SKILLS_TEST_HF_MODEL=Qwen/Qwen3-4B
 export NEMO_SKILLS_TEST_MODEL_TYPE=qwen
+# TRTLLM still doesn't support Qwen3 models, using a smaller Qwen2.5 model for context retry tests
```
Kipok, line 6: it actually does, but you need to use `--backend pytorch` argument in server args

nemo_skills/inference/model/__init__.py
```diff
@@ -53,6 +53,9 @@
 def get_model(server_type, tokenizer=None, **kwargs):
     """A helper function to make it easier to set server through cmd."""
     model_class = models[server_type.lower()]
+    if server_type == "trtllm" and kwargs.get("enable_soft_fail", False):
+        if kwargs.get("context_limit_retry_strategy", None) is not None:
+            raise ValueError("context_limit_retry_strategy is not supported for trtllm")
```
Kipok, line 58: what about api models, are they supported or it's case-by-case there?

Kipok (approved):

tests/gpu-tests/run_qwen.sh
shtoshni: oh i see

nemo_skills/inference/model/__init__.py
shtoshni: I have not tested API models so far.
It's a manual process where you get the errors: 
- for exceeding the length with the prompt only and 
- for exceeding the length with the generation messages + prompt.
 
Add parts of those error messages that we can compare against. And then write a regex to extract the error info from that. 

Let me add it in another PR.

PR #790, opened 2025-09-07
Riva NIMs can be run as part of the generate pipeline

vmendelev:
This PR includes:
1. Generic server type.
2. A serve script which configures and starts Riva server within a NIM
3. Generation modules for ASR and TTS tasks.
4. Example scripts to run ns.generate for ASR and TTS.


nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,289 @@
...
+            ssl_cert,
+            use_ssl,
+            f"{host}:{port}",
+            metadata,
+            max_message_length=max_message_length,
+        )
+        self.service = riva.client.ASRService(auth)
+        
+        # Add thread pool for async execution to avoid blocking the event loop
+        self._executor = ThreadPoolExecutor(max_workers=max_workers)
```
Kipok, line 122: do they not have a proper asyncio api for the client? Or is there some other reason we want a thread pool here?

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,289 @@
...
+    
+    def generate_sync(self, prompt: str, **kwargs):
+        """Override to use gRPC instead of litellm."""
+        return self._generate_single(prompt, **kwargs)
+    
+    def _generate_single(
+        self,
+        prompt: str,
+        # Unused generation parameters – they are accepted so that the wrapper
+        # is API-compatible with ``BaseModel``.
```
Kipok, line 153: If there is almost nothing shared, we can also have a fully different class here without inheritance. We can then have an if-else or handle it through overrides on the inference side.

Kipok: If we still go with inheritance, it's great to have some checks that any unsupported parameters aren't set to non-default values, as we don't want users to set some things and expect the behavior to change, but our code just silently ignore that. But probably easier to just not inherit

Kipok: Or potentially we make a new BaseLLMModel and make the current BaseModel more abstract to cover both speech and text usecases

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,289 @@
...
+        extra_body = extra_body or {}
+
+        # ------------------------------------------------------------------
+        # Build the RecognitionConfig – we expose only the most common fields by
+        # default but allow the caller to supply *anything* supported by the
+        # proto via ``extra_body['config_overrides']``.
+        # ------------------------------------------------------------------
+
+        language_code = extra_body.get("language_code", self.language_code)
+        max_alternatives = int(extra_body.get("max_alternatives", 1))
```
Kipok: would be good to maybe have a dataclass with all supported parameters, so that users don't accidentally pass something not supported and we just ignore it. This makes our code more robust to users having typos in the arguments, expecting things to change and we just silently ignore those instead of raising an error

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,289 @@
...
+            "pred_text": pred_text,
+            "words": words,
+            "generation_time": generation_time,
+            "audio_file": str(audio_path),
+        }
+
+        if include_response:
+            result["response"] = response  # protobuf message
+
+        return dict(generation = result)
```
Kipok: we don't have to wrap everything in a generation. I think generation is a required key, but any other keys are also supported

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,289 @@
...
+        port: str = "50051",
+        model: str = "riva-asr",
+        *,
+        base_url: str | None = None,  # accepted but ignored – matches other models
+        use_ssl: bool = False,
+        ssl_cert: Optional[str] = None,
+        metadata: Optional[List[tuple]] = None,
+        language_code: str = "en-US",
+        max_message_length: int = -1,
+        max_workers: int = 64,  # Number of threads for concurrent requests
```
Kipok, line 81: is the default too low maybe? I'd set to something like 2k here

nemo_skills/inference/model/tts_nim.py
```diff
@@ -0,0 +1,273 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: most comments from asr nim apply here as well

nemo_skills/inference/server/serve_riva_nim.py
```diff
@@ -0,0 +1,116 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+import subprocess
+from typing import Dict, List, Tuple
+
+
+def _visible_devices_from_count(num_gpus: int) -> str:
```
Kipok: is this unused?

nemo_skills/pipeline/utils/server.py
```diff
@@ -199,6 +201,17 @@ def get_server_command(
             num_tasks = 1
         else:
             num_tasks = num_gpus
+    elif server_type == "generic":
+        if not server_entrypoint:
+            raise ValueError("For 'generic' server type, 'server_entrypoint' must be specified.")
+        server_start_cmd = (
+            f"{server_entrypoint} "
+            f"    --model {model_path} "
+            f"    --num_gpus {num_gpus} "
```
Kipok, line 212: maybe also accept num_nodes?

recipes/asr_tts/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```yaml
+env_vars:
+  NGC_API_KEY: "your-ngc-api-key-here"
+```
+
+### Dependencies
+The Riva client library will be installed automatically via the `installation_command` in the scripts:
+```bash
+pip install nvidia-riva-client==2.21.1
```
Kipok, line 19: is this heavy? We can also just add this to default requirements

recipes/asr_tts/README.md
```diff
@@ -0,0 +1,75 @@
+# ASR and TTS Generation with Riva NIMs
```
Kipok, line 1: Should we have this inside main docs instead of recipes? Or keep here for now and move to main docs when this is more broadly used?

recipes/asr_tts/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+## Prerequisites
+
+### NGC API Key
+The `NGC_API_KEY` environment variable must be present in your cluster configuration to pull Riva NIM containers from NGC. Add it to your cluster config file (e.g., `cluster_configs/dfw.yaml`):
+
+```yaml
+env_vars:
+  NGC_API_KEY: "your-ngc-api-key-here"
```
Kipok, line 13: Can we call it NVIDIA_API_KEY for consistency with LLM? We can re-map somewhere in the code. Or we could also rename the LLM one, is NGC_API_KEY a standard name? I think I saw NVIDIA_API_KEY used on some docs on build.nvidia.com, but maybe they updated it

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
+import logging
```
Kipok, line 15: license header

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+    language_code: str = 'en-US'
+    sample_rate_hz: int = 22050
+    
+    # ASR parameters  
+    automatic_punctuation: bool = True
+    speaker_diarization: bool = False
+    
+    generation_key: str = "result"
+    prompt_config: str | None = None
+    prompt_format: str = "openai"  # Bypass prompt validation for TTS/ASR
```
Kipok, line 44: again, similar to the model class, we can maybe refine the structure here to be able to inherit more naturally. E.g. we can add another prompt format here and we can maybe add a new LLMGenerationTask and make the original GenerationTask more general to naturally support asr/tts. We can do it in another pr to merge this one faster

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    def wait_for_server(self):
+        """Override to wait for gRPC server instead of HTTP."""
+        host = self.cfg.server.get('host', '127.0.0.1')
+        grpc_port = int(self.cfg.server.get('port', '8000')) + 1
+        
+        LOG.info(f"Waiting for Riva gRPC server at {host}:{grpc_port}")
+        
+        # Try to connect with grpc_health_probe or netcat
+        max_attempts = 60  # Wait up to 3 minutes
```
Kipok: not sure if 3 minutes is reasonable as a default setup. Maybe 10 or even 20 minutes is a safer option to allow for custom large models that could potentially be used here?

Kipok: We can also add this logic to the parent class as I guess it's useful there as well. But can be done later

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+    prompt_format: str = "openai"  # Bypass prompt validation for TTS/ASR
+
+
+cs = hydra.core.config_store.ConfigStore.instance()
+cs.store(name="base_riva_generate_config", node=RivaGenerateConfig)
+
+
+class RivaGenerationTask(GenerationTask):
+    def __init__(self, cfg: RivaGenerateConfig):
+        self.riva_cfg = cfg
```
Kipok: why renaming? The base class already saves this as .cfg

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+        LOG.warning(f"Riva gRPC server not responding after {max_attempts*3} seconds")
+    
+    def setup_llm(self):
+        host = self.cfg.server.get('host', '127.0.0.1')
+        grpc_port = str(int(self.cfg.server.get('port', '8000')) + 1)
+        
+        if self.riva_cfg.generation_type == 'tts':
+            from nemo_skills.inference.model.tts_nim import TTSNIMModel
+            Path(self.riva_cfg.tts_output_dir).mkdir(parents=True, exist_ok=True)
+            return TTSNIMModel(
```
Kipok, line 106: in the future, we should ideally handle this via a standard get_model class, but can keep current logic for now

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+            return data_point.get('audio_path', data_point.get('audio_file', ''))
+    
+    def log_example_prompt(self, data):
+        if data:
+            LOG.info(f"Example input: {self.fill_prompt(data[0], data)}")
+    
+    async def process_single_datapoint(self, data_point, all_data):
+        prompt = self.fill_prompt(data_point, all_data)
+        if not prompt:
+            return {self.cfg.generation_key: "", "error": "Empty input"}
```
Kipok: why hide this by default? I'd rather have it fail here as it's easy to run this on partially corrupted data files and not notice it. If you have a use-case where it's normal to have missing data points, we can enable it with a parameter

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,155 @@
...
+        
+        extra_body = dict(self.cfg.inference.extra_body) if self.cfg.inference.extra_body else {}
+        
+        if self.riva_cfg.generation_type == 'tts':
+            # Pass through all TTS and zero-shot parameters from data_point
+            for key in ['voice', 'language_code', 'zero_shot_audio_prompt_file',
```
Kipok: should we maybe put all of this in a subdict and then just use ** syntax to pass it through? Again with the motivation to avoid users having a typo and us silently ignoring it

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
Kipok, line 1: would be great to eventually add this as part of our slurm tests https://github.com/NVIDIA/NeMo-Skills/tree/main/tests/slurm-tests, so that we ensure this part is working after updates

nemo_skills/inference/model/asr_nim.py
vmendelev: I think the server may not handle more.

nemo_skills/inference/model/asr_nim.py
vmendelev: I think they don't have it.

nemo_skills/inference/model/asr_nim.py
vmendelev: I've gotten rid of the inheritance in the next revision and added validation for unused parameters value.

nemo_skills/inference/model/asr_nim.py
vmendelev: Will be available in the next revision

nemo_skills/inference/model/tts_nim.py
vmendelev: Yes, applied same fixes.

nemo_skills/inference/server/serve_riva_nim.py
vmendelev: TY. Will be fixed in the next revision

nemo_skills/pipeline/utils/server.py
vmendelev: We can, but not sure will work with current Riva server and client. How is client-server in general working in this case? Server has several nodes but then where is the client running?

recipes/asr_tts/README.md
vmendelev: I think adding a reference to the main documentation is a good idea. But I would still have md file here as well. It's nice when the repo is self contained documentation wise.

recipes/asr_tts/README.md
vmendelev: No. Very lightweight

recipes/asr_tts/riva_generate.py
vmendelev: TY. Will be fixed in the next revision

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
vmendelev: Will be available in the next revision

recipes/asr_tts/README.md
vmendelev: Not sure here. The NIMs require specifically NGC_API_KEY. I've even read there are some products which require both :)

vmendelev: It is at least clear to me where to get NGC_API_KEY while not NVIDIA_API_KEY. Maybe keep NGC_API_KEY for NIMs for now until there is clear guidance? Or whom can we ask?

nemo_skills/pipeline/utils/server.py
Kipok: client always running on the master node, but server is distributed (with master node ip getting all requests)

recipes/asr_tts/README.md
Kipok: How about we accept both, but then in the code (probably here) https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/pipeline/utils/cluster.py#L108 we can duplicate them so that they are both defined and to the same value. Unless user explicitly defined both with different values in which case we don't want to interfere

nemo_skills/pipeline/utils/server.py
vmendelev: I see. Will add num_nodes.

recipes/asr_tts/riva_generate.py
vmendelev: > We can do it in another pr to merge this one faster

That would be the best option.

recipes/asr_tts/riva_generate.py
vmendelev: Good point. Fixed.

recipes/asr_tts/riva_generate.py
vmendelev: The check here is flawed. I've fixed it in the new revision. We can put it into the parent class with a separate PR.

recipes/asr_tts/riva_generate.py
vmendelev: TY. Don't see such use-cases. Will raise.

recipes/asr_tts/riva_generate.py
vmendelev: Will fix this.

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,314 @@
...
+        validate_unsupported_params(kwargs, "ASRNIMModel")
+        
+        loop = asyncio.get_event_loop()
+        # Run the blocking call in a thread pool to avoid blocking the event loop
+        return await loop.run_in_executor(
+            self._executor,
+            lambda: self._generate_single(prompt, **kwargs)
+        )
+    
+    def generate_sync(self, prompt: str, **kwargs):
```
Kipok: we actually removed sync api recently, so this is no longer necessary

nemo_skills/inference/model/asr_nim.py
```diff
@@ -0,0 +1,314 @@
...
+            transcript_parts = []
+            for result in response.results:
+                if result.alternatives:
+                    transcript_parts.append(result.alternatives[0].transcript)
+                    # Collect words from this chunk
+                    for w in result.alternatives[0].words:
+                        words.append({
+                            "start_time": w.start_time / 1000,  # s -> ms
+                            "end_time": w.end_time / 1000,  # s -> ms
+                            "word": w.word,
```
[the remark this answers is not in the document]
Kipok: @vmendelev ^

nemo_skills/inference/model/tts_nim.py
```diff
@@ -0,0 +1,301 @@
...
+        validate_unsupported_params(kwargs, "TTSNIMModel")
+        
+        loop = asyncio.get_event_loop()
+        # Run the blocking call in a thread pool to avoid blocking the event loop
+        return await loop.run_in_executor(
+            self._executor, 
+            lambda: self._generate_single(prompt, **kwargs)
+        )
+    
+    def generate_sync(self, prompt: str, **kwargs):
```
Kipok: could be removed

nemo_skills/inference/model/nim_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        'reasoning_effort': (kwargs.get('reasoning_effort'), None),
+    }
+    
+    non_default = []
+    for param, (val, default) in unsupported_checks.items():
+        if val is not None and val != default:
+            non_default.append(f"{param}={val}")
+    
+    if non_default:
+        LOG.warning(
```
Kipok, line 169: should we fail with error in this case? Feels like if users try to specify one of the unsupported parameters, something might not be right in the setup. But if you think warning is better, I guess we can keep it

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -198,18 +199,14 @@ def get_env_variables(cluster_config):
...
-            else:
-                raise ValueError(f"Cannot resolve environment variable {key} inside the placeholder value: {value}")
-
+    # Unless NGC_API_KEY is explicitly set we will populate it to be equal to NVIDIA_API_KEY
+    if "NGC_API_KEY" not in env_vars:
+        if "NVIDIA_API_KEY" in env_vars:
+            env_vars["NGC_API_KEY"] = env_vars["NVIDIA_API_KEY"]
+            LOG.info("Populating NGC_API_KEY to be equal to NVIDIA_API_KEY")
+        else:
+            LOG.warning("NGC_API_KEY is not set, but NVIDIA_API_KEY is not set either. "
```
Kipok: I don't think we need a warning here. It's perfectly common situation not to have it set if people are self-hosting models

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -198,18 +199,14 @@ def get_env_variables(cluster_config):
...
-                    f"Resolved environment variable {key} inside the placeholder value: {value} with {env_vars[key]}"
-                )
-            else:
-                raise ValueError(f"Cannot resolve environment variable {key} inside the placeholder value: {value}")
-
+    # Unless NGC_API_KEY is explicitly set we will populate it to be equal to NVIDIA_API_KEY
+    if "NGC_API_KEY" not in env_vars:
+        if "NVIDIA_API_KEY" in env_vars:
+            env_vars["NGC_API_KEY"] = env_vars["NVIDIA_API_KEY"]
+            LOG.info("Populating NGC_API_KEY to be equal to NVIDIA_API_KEY")
```
Kipok: can we add the same check for logged env vars as above to avoid this message being printed many times over if large pipelines are submitted

nemo_skills/pipeline/utils/exp.py
```diff
@@ -241,6 +235,11 @@ def get_executor(
         additional_parameters["mail_type"] = cluster_config["mail_type"]
     if cluster_config.get("mail_user") is not None:
         additional_parameters["mail_user"] = cluster_config["mail_user"]
+    
+    # Merge slurm_kwargs into additional_parameters
```
Kipok: are you sure this is working properly? From first glance slurm_kwargs are applied as main arguments, while additional_parameters are being passed with a separate keyword, but could be that eventually they are the same. Just want to make sure you double checked it works like that

nemo_skills/pipeline/run_cmd.py
```diff
@@ -107,15 +106,16 @@ def run_cmd(
...
+    skip_hf_home_check: bool = typer.Option(
+        False,
         help="If True, skip checking that HF_HOME env var is defined in the cluster config.",
     ),
     dry_run: bool = typer.Option(False, help="If True, will not run the job, but will validate all arguments."),
     _reuse_exp: str = typer.Option(None, help="Internal option to reuse an experiment object.", hidden=True),
     _task_dependencies: List[str] = typer.Option(
         None, help="Internal option to specify task dependencies.", hidden=True
     ),
+    _slurm_kwargs_nodelist: str = typer.Option(None, help="Internal option to specify nodelist.", hidden=True),
```
Kipok: what's the use-case of this parameter? Generally we want to try to keep thing consistent in terms of arguments like this across all pipelines, so if we specify it here, we should probably specify it in all other pipelines as well. Another thing we can consider is to just add a blank `slurm_kwargs` dictionary object that's only specifiable from python with arbitrary arguments, so that we don't have to keep adding them one-by-one

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
Kipok, line 1: license

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
```diff
@@ -0,0 +1,32 @@
...
+SERVER_ARGS='--CONTAINER_ID parakeet-tdt-0.6b-v2 --NIM_TAGS_SELECTOR name=parakeet-tdt-0.6b-v2,mode=ofl --nim-disable-model-download false'
+HF_HOME=?
+INPUT_FILE=/workspace/asr.jsonl
+OUTPUT_DIR=/workspace/asr_a1
+WORKSPACE=?
+
+MOUNTS="$HF_HOME:$HF_HOME,$WORKSPACE:/workspace"
+
+ns generate \
+    --cluster dfw \
```
Kipok: accept from command line instead of hardcoding?

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
```diff
@@ -0,0 +1,32 @@
...
+    --generation_module recipes.asr_tts.riva_generate \
+    --input_file "$INPUT_FILE" \
+    --output_dir "$OUTPUT_DIR" \
+    --server_type generic \
+    --server_entrypoint "python3 -m nemo_skills.inference.server.serve_riva_nim" \
+    --mount_paths "$MOUNTS" \
+    --server_gpus 1 \
+    --installation_command "pip install nvidia-riva-client==2.21.1" \
+    --num_chunks 1 \
+    --partition interactive \
```
Kipok: also doesn't look like a good default. We generally should try to structure scripts such that they would work on any cluster, not necessarily the ones we have internally

recipes/asr_tts/scripts/run_tts_nim_cluster.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
Kipok, line 1: license

recipes/asr_tts/nim_configurations.py
```diff
@@ -0,0 +1,32 @@
...
+        "server_entrypoint": "DISABLE_RIVA_REALTIME_SERVER=True python3 -m nemo_skills.inference.server.serve_riva_nim",
+        "installation_command": "pip install nvidia-riva-client==2.21.1",
+        "local_images": {} # items shouldbe "cluster_name": "path_to_image.sqsh"
+    },
+    "parakeet-tdt-0.6b-v2:1.0.0": {
+        "image_uri": "nvcr.io/nim/nvidia/parakeet-tdt-0.6b-v2:1.0.0",
+        "server_args": "--CONTAINER_ID parakeet-tdt-0.6b-v2 --nim-tags-selector name=parakeet-tdt-0.6b-v2,mode=ofl  --nim-disable-model-download false",
+        "server_entrypoint": "python3 -m nemo_skills.inference.server.serve_riva_nim",
+        "installation_command": "pip install nvidia-riva-client==2.21.1",
+        "local_images": {} # items shouldbe "cluster_name": "path_to_image.sqsh"
```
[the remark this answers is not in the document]
Kipok: @vmendelev ^

recipes/asr_tts/README.md
Kipok: Consider adding to main requirements then, up to you, but feels more convenient not to always install on-the-fly

recipes/asr_tts/README.md
```diff
@@ -0,0 +1,75 @@
...
+The recipe supports:
+- `--num_chunks`: Split input into chunks for parallel processing
+- `--skip_filled`: Skip already processed samples
+- Async processing for improved throughput
+
+## Support
+
+For issues or questions, refer to:
+- [NVIDIA Riva Documentation](https://docs.nvidia.com/deeplearning/riva/user-guide/docs/index.html)
+- [NeMo Skills Documentation](https://github.com/NVIDIA/NeMo-Skills)
```
Kipok:
```suggestion
- [NeMo Skills Documentation](https://github.com/NVIDIA-NeMo/Skills)
```

tests/slurm-tests/asr_nim/run_test.py
```diff
@@ -0,0 +1,420 @@
...
+    # Extract HF_HOME from env_vars
+    hf_home = None
+    env_vars = cluster_config.get("env_vars", [])
+    for env_var in env_vars:
+        if isinstance(env_var, str) and "HF_HOME=" in env_var:
+            hf_home = env_var.split("=", 1)[1]
+            break
+    
+    if not hf_home:
+        hf_home = "/lustre/fsw/portfolios/convai/users/vmendelev/.cache/huggingface"
```
Kipok: why do we need this? Everyone should just have this defined in a config

tests/slurm-tests/asr_nim/run_test.py
```diff
@@ -0,0 +1,420 @@
...
+    cluster_config = get_cluster_config(cluster)
+    
+    # Ensure workspace exists
+    create_remote_directory(workspace, cluster_config)
+    
+    # Get tunnel and establish connection
+    tunnel = get_tunnel(cluster_config)
+    tunnel.run("echo 'Establishing connection...'", hide=True)
+    
+    # Upload test data JSONL
```
Kipok: I don't think we need an explicit upload here. Anything inside this repo will be packaged and uploaded automatically when job is submitted so you just need to reference those as `/nemo_run/code/tests/slurm-tests/asr_nim/wavs/...`

tests/slurm-tests/tts_nim/run_test.py
```diff
@@ -0,0 +1,354 @@
...
+    # Upload test data from dev desktop to cluster
+    # Need to ensure tunnel connection is established first
+    tunnel = get_tunnel(cluster_config)
+    tunnel.run("echo 'Establishing connection...'", hide=True)
+    cluster_upload(cluster_config, test_data_path, input_file, verbose=True)
+    
+    # Set up partition and GPU requirements based on whether we need to run on specific node
+    if server_node:
+        # When running on same node as server, use interactive partition with GPU
+        partition = "interactive"
```
Kipok: probably not the best default (not available on all clusters)

tests/slurm-tests/tts_nim/run_test.py
```diff
@@ -0,0 +1,354 @@
...
+        num_gpus = 1
+    else:
+        # When running on any node, use CPU partition
+        partition = "cpu"
```
Kipok: this should automatically happen as long as you have `cpu_partition: cpu` defined in the config, so again best not to reference explicitly as names might be different

tests/slurm-tests/tts_nim/run_test.py
```diff
@@ -0,0 +1,354 @@
...
+        output_dir=output_dir,
+        model="tts_nim",  # Dummy model name for TTS NIM
+        server_type="generic",
+        num_chunks=1,
+        server_gpus=1,
+        server_entrypoint=nim_config["server_entrypoint"],
+        server_container=container,
+        server_args=nim_config["server_args"],
+        installation_command=nim_config["installation_command"],
+        partition="interactive",
```
Kipok: best not to set

tests/slurm-tests/tts_nim/run_test.py
```diff
@@ -0,0 +1,354 @@
...
+        expname=expname_prefix,
+        mount_paths=mount_paths,
+        check_mounted_paths=True,
+        exclusive=True,  # Use exclusive mode to get random ports
```
Kipok: why is this needed?  This is wasting 7 gpus out of 8 on the node

Kipok:
@vmendelev let's try to fix as much as possible by the end of the week and aim to merge in whatever state this is and then we can refine in further prs? Otherwise it might take too long to merge this and it's already very large

nemo_skills/inference/model/asr_nim.py
vmendelev: Will get rid of it.

nemo_skills/inference/model/asr_nim.py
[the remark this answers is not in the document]
vmendelev: fixed

recipes/asr_tts/nim_configurations.py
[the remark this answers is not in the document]
vmendelev: fixed

recipes/asr_tts/riva_generate.py
```diff
@@ -0,0 +1,186 @@
...
+                output_dir=self.cfg.tts_output_dir
+            )
+        elif self.cfg.generation_type == 'asr':
+            from nemo_skills.inference.model.asr_nim import ASRNIMModel
+            return ASRNIMModel(
+                host=host, 
+                port=grpc_port, 
+                model='riva-asr',  # Required by BaseModel
+                language_code=self.cfg.language_code
+            )
```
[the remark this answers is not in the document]
vmendelev: This is impossible for now.

nemo_skills/inference/model/nim_utils.py
vmendelev: I think some parameters may be supported in the future, like temperature for TTS generations. So I would keep it as a warning

nemo_skills/pipeline/utils/exp.py
vmendelev: Good catch. TY. Reworked this.

nemo_skills/pipeline/run_cmd.py
vmendelev: This is needed for debugging. Slurm tests has a regime where you can start a server, and then a client, but in this case we need to pass the exact node with the server because otherwise they don't see each other. Are there better ways to achieve this?

vmendelev: I think we do need slurm_kwargs available here.

vmendelev: Supporting slurm_kwargs means again changes in many files, and this PR is already big. So I will disable for now server + client debugging and so will remove this property. The usual end2end slurm test will work. I will create a separate CR to support slurm_kwargs in all the calls and then reenable the partial mode in the slurm tests .

recipes/asr_tts/scripts/run_asr_nim_cluster.sh
vmendelev: Makes sense. Will remove

tests/slurm-tests/tts_nim/run_test.py
vmendelev: This branching is not needed. server_node is assumed to always be set, so the partition should same as for the server. So I will just omit partition everywhere. Is this a good idea or would suggest something different?

tests/slurm-tests/tts_nim/run_test.py
vmendelev: This was a mistake.

Kipok (approved):
Thanks for all the hard work to do this integration @vmendelev !

PR #791, opened 2025-09-08
GenSelect -> GenEvolution

shtoshni:
This PR adds support for GenSynthesis.

```      
f"++genevolution=True "
f"++genevolution_config.mode=gensynthesis "
```

For GenSelect:
```      
f"++genevolution=True "
f"++genevolution_config.mode=genselect "
```

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -133,10 +133,9 @@ class GenerateSolutionsConfig:
 
     # stop phrase for llms
     stop_phrase: str | None = None  # if None, will not add any extra stop phrase
-    # set to True if parallel_thinking is used
-    parallel_thinking: bool = False
+
     # parallel_thinking config
-    parallel_thinking_config: ParallelThinkingConfig = field(default_factory=ParallelThinkingConfig)
+    parallel_thinking: ParallelThinkingConfig = field(default_factory=ParallelThinkingConfig)
 
     # Module-based tool configuration
     #   List of tool provider locators using double-colon syntax for the tool class.
@@ -298,7 +297,7 @@ class GenerationTask:
         )
 
         # Initialize semaphore for controlling concurrent requests
-        if self.cfg.parallel_thinking:
+        if self.cfg.parallel_thinking.mode is not None:
             # Each request will generate multiple solutions, so we need to divide the semaphore by the parallel requests
             self.semaphore = asyncio.Semaphore(
                 self.cfg.max_concurrent_requests // self.llm.cfg.max_concurrent_requests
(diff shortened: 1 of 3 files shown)


nemo_skills/inference/model/genevolution.py
```diff
@@ -132,20 +172,48 @@ async def _run_genselect(
...
+        genselect_result = await self._generate_genimprovement(prompt=genselect_prompt, **kwargs)
 
         # Step 3: Extract the judgment from the GenSelect result
-        judgment = self._extract_judgment(genselect_result["generation"], max_idx)
-        if judgment is None:
-            LOG.warning("GenSelect failed to produce valid judgment, falling back to random selection")
-            judgment = local_random.randint(0, max_idx)
+        chosen_solution_idx = self._extract_chosen_solution(genselect_result["generation"], max_idx)
+        if chosen_solution_idx is None:
+            LOG.warning("GenSelect failed to produce valid solution index, falling back to random selection")
```
smahdavi4: We probably want to log this failure in the generation output as well. Also, more broadly, it would be useful to add a config key such as `verbose_generation`, to log all genselect/gensynthesis judgements to the output if someone wants to debug how the model is reasoning.

nemo_skills/inference/model/genevolution.py
```diff
@@ -32,45 +32,68 @@
...
-    chat_template_kwargs: dict | None = None  # extra parameters to pass to the tokenizer's apply_chat_template method
+class GenSelectSpecificConfig:
+    prompt_config: str = "generic/genselect"
+    regex: str = r"Judg[e]?ment: (\d+)"
+
+
+@nested_dataclass(kw_only=True)
+class GenSynthesisSpecificConfig:
+    prompt_config: str = "generic/gensynthesis"
+    regex: str = r"<NEW_SOLUTION>\n(.*?)\n</NEW_SOLUTION>"
```
smahdavi4: Minor, but we can probably remove \n for regex and strip the solution to cover more model failures (in case the model does not add new line to the xml tag).

nemo_skills/inference/model/genevolution.py
shtoshni: Logging it here:
- https://github.com/NVIDIA/NeMo-Skills/blob/shtoshni/gensynthesis/nemo_skills/inference/model/genevolution.py#L352
- https://github.com/NVIDIA/NeMo-Skills/blob/shtoshni/gensynthesis/nemo_skills/inference/model/genevolution.py#L362

Let me add a boolean for valid solution idx vs not.

shtoshni: Added a boolean entry - `synthesis_successful` / `selection_successful` to mark valid synthesis/selection

nemo_skills/inference/generate.py
```diff
@@ -133,10 +133,10 @@ class GenerateSolutionsConfig:
...
     # stop phrase for llms
     stop_phrase: str | None = None  # if None, will not add any extra stop phrase
-    # set to True if genselect is used
-    genselect: bool = False
-    # genselect config
-    genselect_config: GenSelectConfig = field(default_factory=GenSelectConfig)
+    # set to True if parallel_thinking is used
+    parallel_thinking: bool = False
+    # parallel_thinking config
+    parallel_thinking_config: ParallelThinkingConfig = field(default_factory=ParallelThinkingConfig)
```
Kipok: would be a bit cleaner and shorter to just use a dictionary with name `parallel_thinking` here. Similar to how we do for inference, so you can set `inference.temperature=X` instead of `inference_config.temperature=X`. So you can either set to None by default (not sure if that's going to work) or just set `mode=disable` by default or something like this. If it turns out to be too involved to do that, we can also keep the current way

shtoshni: hmm yeah that's a good point. let me do that.

PR #792, opened 2025-09-08
GenEvolution docs

shtoshni:
Added docs for running `GenSelect` and `GenSynthesis`

### DIFF
--- a/docs/agentic_inference/parallel_thinking.md
+++ b/docs/agentic_inference/parallel_thinking.md
@@ -64,7 +64,7 @@ ns eval \
 ```
 
 The evaluation pipeline would first generate `window_size` solutions (8 by default), and then run GenSynthesis with these solutions in the prompt to synthesize a new solution.
-Note that the same model is being used for both solution generation and selection, which we refer to as **Self-GenSynthesis**.
+Note that the same model is being used for both solution generation and synthesis, which we refer to as **Self-GenSynthesis**.
 
 !!!tip
     Parallel Thinking inputs can consume a lot of tokens, especially for large `window_size` values.


docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
+# GenEvolution
+
+GenEvolution encompasses methods that scale inference time via parallel sampling. The approach entails primarily two methods:
+
+
+- **GenSelect** is a generative Best-of-N method we introduced in the [OpenReasoning paper](https://arxiv.org/abs/2504.16891, followed by a more focused paper -- [GenSelect: A Generative Approach to Best-of-N](https://arxiv.org/abs/2507.17797). The method essentially uses an LLM to reason over and select the best candidate solution among the N candidates, leveraging LLMs' comparative strengths while scaling efficiently across parallel sampling budgets.
```
Kipok:
```suggestion
- **GenSelect** is a generative Best-of-N method we introduced in the [OpenReasoning paper](https://arxiv.org/abs/2504.16891), followed by a more focused paper -- [GenSelect: A Generative Approach to Best-of-N](https://arxiv.org/abs/2507.17797). The method essentially uses an LLM to reason over and select the best candidate solution among the N candidates, leveraging LLMs' comparative strengths while scaling efficiently across parallel sampling budgets.
```

mkdocs.yml
```diff
@@ -81,6 +81,9 @@ nav:
     - evaluation/tool-calling.md
     - evaluation/multilingual.md
     - evaluation/other-benchmarks.md
+  - Agentic Generation:
```
Kipok:
```suggestion
  - Agentic Inference:
```
maybe generation -> inference? Not sure on this one, so if you prefer generation, feel free to keep it

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+!!!note
+    The GenEvolution pipeline uses the same inference parameters as the generate pipeline. We allow overriding of two key inference config params:
+
+    - `temperature` via `++genevolution_config.temperature=<>`
+    - `tokens_to_generate` via `++genevolution_config.tokens_to_generate=<>`
+
+
+
+### Common Parameters
+- `window_size`: Number of solutions compared in a single GenSelect comparison (typically set to 8). Consider your model's context window size when setting this value (or allow for soft failure via `++server.enable_soft_fail=True`).
```
Kipok: link to soft-failure docs would be helpful here

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Offline GenEvolution Parameters
+
+These parameters only need to be passed when running offline GenEvolution.
+
+- `generation_dir`: The directory where the *offline* generated solutions are stored. We assume the solutions to be in `output-rs*.jsonl` files.
+- `num_initial_solutions`: Number of solutions from the offline generated solutions that are used for GenEvolution.
+
+
+To specify these variables, say `window_size=16`, pass `++genevolution_config.window_size=16` to the generate/eval pipelines.
```
Kipok: window_size isn't mentioned above, is this intentional?

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+## Sample Illustrations
```
Kipok: maybe "Examples"?

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+### Online GenEvolution (GenSelect)
```
Kipok: maybe via GenSelect? Otherwise it feels like online is always genselect and offline is always gensynthesis

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+
+In this example, we show how to use GenSelect for [aime25](https://nvidia.github.io/NeMo-Skills/evaluation/natural-math/) with [Qwen/Qwen3-8B](https://huggingface.co/Qwen/Qwen3-8B).
+
+```bash hl_lines="9-10"
+ns eval \
```
Kipok: not necessary now, but I think we should generally try to specify all examples with both python and cli interface. We generally prefer people to use python as it's more efficient and easier to manage, but since most of our examples are standalone commands, they tend to be cli, which creates wrong impression that this is how things should be used

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+
+Offline genevolution breaks down the generation and evolution (selection/synthetis) part into two separate steps. There are two use cases which are currently only supported with offline genselect:
+
+- Using a different model for generation and *evolution*.
+- Using a processed version of generated output for *evolution*.
+
+In the following example, we use [Qwen/Qwen3-8B](https://huggingface.co/Qwen/Qwen3-8B) to perform GenSelect over solutions generated by [Qwen/Qwen3-4B](https://huggingface.co/Qwen/Qwen3-4B) for `livecodebench`.
+
+```bash hl_lines="9-12"
+ns eval \
```
Kipok: this could be a nice example of python api as well where you can show the original eval command and chain it with the genselect one

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
...
+There are three things we want to highlight in the above example:
+
+- We run the GenSynthesis pipeline a total of 8 times (`livecodebench:8`) over the same set of solutions
+- The pre-generated solutions are specified via: `++genevolution_config.generation_dir=/experiments/qwen3_4b/eval-results/livecodebench`
+- Instead of the usual `generation` key for comparing the different solutions, we use `++genevolution_config.comparison_key=completion`
+
+The `completion` key in `livecodebench` outputs contains just the extracted code from the generated solutions. For coding tasks, we empirically find that representing the candidate solution with just the extracted code to perform better than representing it with the text around it.
+
+
+## Papers
```
Kipok: maybe not needed here, we already have a section for papers. If you want, you can maybe just link from here to there, but I don't think we need to add all references in multiple places

docs/agentic_generation/genevolution.md
shtoshni: Good point. Adding an example for offline gensynthesis.

docs/agentic_generation/genevolution.md
shtoshni: It's mentioned above in the `Common Parameters` section

shtoshni: Screenshot 2025-09-08 at 6 19 41 PM

Kipok: I see, was just confusing because it looked like "above" refers to the offline subsection

docs/agentic_generation/genevolution.md
```diff
@@ -0,0 +1,139 @@
+# GenEvolution
```
Kipok: should this be called Parallel Thinking now?

PR #794, opened 2025-09-08
Fix SWE-bench parallel runs and other issues

ludwig-n:
Fixes the following issues for SWE-bench:

- Parallel runs with `--benchmark=swe-bench:N` now work correctly by separating each run's output files into a folder called `rs{seed}`. Previously, parallel runs would overwrite each other's output files, e.g. in the `trajectories` folder.
- When rerunning an Apptainer command, log files for the different attempts are now separated. Previously, each new attempt would overwrite the last one, which was annoying when debugging errors.
- Jobs no longer fail [here](https://github.com/NVIDIA/NeMo-Skills/blob/50f3747a73e62fa8ff22b1484b47c25b770eb7e4/nemo_skills/inference/generate.py#L590) because of `enable_litellm_cache` missing in `SweBenchGenerationConfig`. I think this issue was introduced by https://github.com/NVIDIA/NeMo-Skills/pull/789. I just added a `getattr` check but I'm not sure it's the cleanest solution.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -271,7 +271,7 @@ class GenerationTask:
             self.tokenizer = None
 
         # Setup litellm cache
-        if self.cfg.enable_litellm_cache:
+        if getattr(self.cfg, "enable_litellm_cache", False):
             # One cache per (output_file_name, chunk_id) pair
             output_file_name = Path(self.cfg.output_file).name
             self.litellm_cache_dir = (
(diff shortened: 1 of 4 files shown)


nemo_skills/inference/generate.py
```diff
@@ -271,7 +271,7 @@ def __init__(self, cfg: GenerateSolutionsConfig):
             self.tokenizer = None
 
         # Setup litellm cache
-        if self.cfg.enable_litellm_cache:
+        if getattr(self.cfg, "enable_litellm_cache", False):
```
Kipok: maybe instead we add `setup_litellm_cache` and `cleanup_litellm_cache` methods and put this logic inside? And in the swebench we override those to be empty functions. I think that's a bit less error-prone than the current getattr approach

ludwig-n: done

PR #798, opened 2025-09-09
Add Long context benchmark AA-LCR

fayejf:


README.md
```diff
@@ -15,7 +15,7 @@ Here are some of the features we support:
...
     - [**Instruction following**](https://nvidia.github.io/NeMo-Skills/evaluation/instruction-following): e.g. [ifbench](https://nvidia.github.io/NeMo-Skills/evaluation/instruction-following/#ifbench), [ifeval](https://nvidia.github.io/NeMo-Skills/evaluation/instruction-following/#ifeval)
-    - [**Long-context**](https://nvidia.github.io/NeMo-Skills/evaluation/long-context): e.g. [ruler](https://nvidia.github.io/NeMo-Skills/evaluation/long-context/#ruler), [mrcr](https://nvidia.github.io/NeMo-Skills/evaluation/long-context/#mrcr)
+    - [**Long-context**](https://nvidia.github.io/NeMo-Skills/evaluation/long-context): e.g. [ruler](https://nvidia.github.io/NeMo-Skills/evaluation/long-context/#ruler), [mrcr](https://nvidia.github.io/NeMo-Skills/evaluation/long-context/#mrcr), [aalcr](https://nvidia.github.io/NeMo-Skills/evaluation/long-context/#aalcr)
```
Kipok, line 18: maybe we don't add it here if it's not possible for most people to run this

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
 
 - Benchmark is defined in [`nemo_skills/dataset/mrcr/__init__.py`](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/mrcr/__init__.py)
 - Original benchmark source is [here](https://huggingface.co/datasets/openai/mrcr).
+
+### [aalcr](https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR)
+- Benchmark is defined in [`nemo_skills/dataset/aalcr/__init__.py`](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/aalcr/__init__.py)
+- Original benchmark source is [here](https://artificialanalysis.ai/evaluations/artificial-analysis-long-context-reasoning).
+
+Data preparation. You will need to get txt files using data_source_url or consult with AA.
```
Kipok: I think we should either add more details on how to get the files, so others can reasonably follow this or just say that it's only runnable if you get files from AA

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
 
 - Benchmark is defined in [`nemo_skills/dataset/mrcr/__init__.py`](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/mrcr/__init__.py)
 - Original benchmark source is [here](https://huggingface.co/datasets/openai/mrcr).
+
+### [aalcr](https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR)
+- Benchmark is defined in [`nemo_skills/dataset/aalcr/__init__.py`](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/aalcr/__init__.py)
+- Original benchmark source is [here](https://artificialanalysis.ai/evaluations/artificial-analysis-long-context-reasoning).
```
Kipok: does this page have details on the benchmark? Should we point to HF or some other page instead?

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
...
+
+Data preparation. You will need to get txt files using data_source_url or consult with AA.
+```bash
+ns prepare_data \
+    --data_dir=/workspace/ns-data \
+    --cluster=<cluster_config> \
+    aalcr --txt_file_folder=/workspace/do_not_share_data/lcr
```
Kipok:
```suggestion
    aalcr --txt_file_folder=<path to lcr data>
```

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
...
+```
+
+You can also prepare a subset of the data with limited context window.
+```bash
+    --max_context_window 100000 --setup aalcr_100k
+```
+
+Example command for running evaluation. It follows official AA-LCR implementation. Qwen3-235B-A22B-Instruct-2507 is served as judge.
```
Kipok: Please put judge arguments here in the below command and update init.py to use gpt-4.1 similar to https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/omni-math/__init__.py#L25

The reason is we want benchmarks to work out of the box, so we should have a default setup by calling some api model as a judge, not hosting locally. But please add a comment here that default judge is different and if users want to match aai setup, they can use it like in your current way

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
...
+
+```bash
+model=Qwen2.5-7B-Instruct-1M
+ns eval \
+    --cluster=<cluster_config> \
+    --data_dir=/workspace/ns-data \
+    --server_gpus=8 \
+    --server_type=sglang \
+    --model=/hf_models/$model \
+    --benchmarks=aalcr:0 \
```
Kipok:
```suggestion
    --benchmarks=aalcr \
```

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
...
+    --split=aalcr \
+    --output_dir=/workspace/aalcr/$split/$model \
+    (--server_args='--disable-cuda-graph')
```
Kipok:
```suggestion
    --server_args='--disable-cuda-graph'
```
is this required?

nemo_skills/prompt/config/judge/aalcr.yaml
```diff
@@ -0,0 +1,14 @@
+# AA-LCR (Artificial Analysis Long Context Reading) judge prompt
+# from https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR
+# Based on the official AA-LCR scoring approach using Qwen3 235B A22B 2507 Non-reasoning
+
+system: ""
```
Kipok:
```suggestion
```

nemo_skills/dataset/aalcr/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+PROMPT_CONFIG = "generic/default"
+DATASET_GROUP = "long-context"
+METRICS_TYPE = "answer-judgement"
```
Kipok: I don't think that's the right type. This one is used for comparing the judge performance itself. I think the right one would be to use the same setup as for HLE here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/evaluation/metrics/map_metrics.py#L34. We should probably add a separate metric type for such cases of judge-only results in the future.

nemo_skills/dataset/aalcr/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+EVAL_ARGS = "++eval_type=math"
```
Kipok:
```suggestion
EVAL_ARGS = "++eval_type=no-op"  # using judgement directly in metrics, no need for special evaluation
```

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,158 @@
...
+    --cluster=fei-ord \
+    aalcr --txt_file_folder=/workspace/do_not_share_data/lcr
+"""
+
+
+def construct_prompt(docs, question):
+    documents_text = "\n\n".join(f"BEGIN DOCUMENT {i + 1}:\n{doc}\nEND DOCUMENT {i + 1}" for i, doc in enumerate(docs))
+    prompt = """BEGIN INPUT DOCUMENTS
+
+    {documents_text}
```
Kipok: this is going to have a bunch of spaces, is this intentional? If not you should remove indentation or use a different string formatting

Kipok: why don't we have it as a normal yaml prompt similar to other benchmarks?

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,158 @@
...
+                    with open(
+                        f"{txt_file_folder}/{document_category}/{document_set_id}/{data_source_filename}",
+                        "rt",
+                        encoding="utf-8",
+                    ) as fin:
+                        document = fin.read()
+                        documents.append(document)
+                except FileNotFoundError:
+                    print(
+                        f"File {txt_file_folder}/{document_category}/{document_set_id}/{data_source_filename} is missing"
```
Kipok: why skipping this, should we fail instead?

nemo_skills/evaluation/metrics/utils.py
```diff
@@ -40,6 +40,12 @@ def is_correct_judgement(judgement, return_none=False) -> Union[bool, None]:
             return True
         elif verdict.lower().startswith("no"):
             return False
+    else:
```
Kipok: Could you maybe add a new metric type for now that will use this logic? We probably want to generalize things a bit in regards to judgement-only benchmarks, but for now I guess we can just support a custom logic just for this one and fix it properly later. I think your current change might introduce unintended consequences in other places relying on this function.

nemo_skills/dataset/aalcr/__init__.py
Kipok: I guess we'd need a new metric type here, added another comment about that

Kipok (changes requested):

docs/evaluation/long-context.md
shtoshni: @fayejf - I think the data_source_urls from huggingface issue is not resolved. If it is, then you should state that the `data_source_urls` field is coming from the [huggingface dataset](https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR).  Otherwise, we can just say that get the data from Artificial Analysis Index team.

docs/evaluation/long-context.md
```diff
@@ -13,3 +13,36 @@ More details are coming soon!
...
+model=Qwen2.5-7B-Instruct-1M
+ns eval \
+    --cluster=<cluster_config> \
+    --data_dir=/workspace/ns-data \
+    --server_gpus=8 \
+    --server_type=sglang \
+    --model=/hf_models/$model \
+    --benchmarks=aalcr:0 \
+    --split=aalcr \
+    --output_dir=/workspace/aalcr/$split/$model \
```
shtoshni: I'm not sure if the `$split` variable is defined above

docs/evaluation/long-context.md
fayejf: They just released the txt fils on hf today! yay! I will update accordingly

README.md
fayejf: They just released the txt fils on hf today! yay! I will update accordingly

docs/evaluation/long-context.md
shtoshni: Also, we can skip specifying the split of the benchmark `aalcr` to be `aalcr` by simply renaming the processed output file to `test.jsonl`

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,158 @@
...
+    parser.add_argument(
+        "--tokenizer_name",
+        type=str,
+        default="cl100k_base",
+        help="tokenizer name",
+    )
+
+    args = parser.parse_args()
+
+    print(f"Preparing AA-LCR dataset with sadditional arguments: {args}")
```
shtoshni:
```suggestion
    print(f"Preparing AA-LCR dataset with additional arguments: {args}")
```

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,158 @@
...
+    parser.add_argument(
+        "--txt_file_folder",
+        type=str,
+        default="lcr",
+        help="txt file folder to process",
+    )
+    parser.add_argument(
+        "--setup",
+        type=str,
+        default="aalcr",
```
shtoshni: I would suggest it to name `test`. Weird to have a benchmark with the same split name :) Plus `test` would just save passing an additional argument during eval

docs/evaluation/long-context.md
fayejf: I think we should keep it as a split similar to livecodebench and other long context benchmark instead renaming to be test.jsonl because for long context we need to test different seq len for models. Some models doesns't allow > max_seq_len and nemo-skills doesn't have a feature to handle those failure yet. aalcr_64k kind of thing is a workaround.

nemo_skills/dataset/aalcr/prepare.py
fayejf: yes we can have default to be test.

docs/evaluation/long-context.md
fayejf: we can make default to be test while keep split so that it can be aalcr_64k or so.

shtoshni: or just `test_64k` for consistency?

docs/evaluation/long-context.md
shtoshni: This URL would be more appropriate - https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR

README.md
fayejf: resolved.

docs/evaluation/long-context.md
fayejf: resolved

docs/evaluation/long-context.md
fayejf: resolved

docs/evaluation/long-context.md
fayejf: resolved

docs/evaluation/long-context.md
fayejf: resolved

nemo_skills/dataset/aalcr/__init__.py
fayejf: Added a new metric aalcr here which also has detailed analysis. I prefer to have a more general metric type and move analysis elsewhere to keep it clean.

nemo_skills/dataset/aalcr/__init__.py
fayejf: resolved

nemo_skills/dataset/aalcr/prepare.py
fayejf: what do you mean
>  normal yaml prompt similar to other benchmarks?

I did not find yaml prompt in other benchmarks. Can you point me to one example please?

nemo_skills/dataset/aalcr/prepare.py
fayejf: resolved

nemo_skills/evaluation/metrics/utils.py
fayejf: resolved

docs/evaluation/long-context.md
fayejf: resolved

docs/evaluation/long-context.md
fayejf: resolved

nemo_skills/dataset/aalcr/prepare.py
fayejf: resolved

nemo_skills/dataset/aalcr/prepare.py
fayejf: resolved

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,253 @@
...
+
+or
+ns prepare_data \
+    --data_dir=/workspace/ns-data \
+    --cluster=fei-ord \
+    aalcr --max_context_window 100000 --setup test_100k
+"""
+
+
+prompt_template = """BEGIN INPUT DOCUMENTS
```
shtoshni, line 44: Can we make this a prompt config? I think it will help in making this be used for other parts of the pipeline as well, say SDG.

docs/evaluation/long-context.md
shtoshni: I'm not sure if the `$split` variable is defined in the eval command block

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,253 @@
...
+    # download the provied extracted text files
+    # https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR/resolve/main/extracted_text/AA-LCR_extracted-text.zip
+
+    if not os.path.exists(Path(__file__).absolute().parent / "lcr"):
+        import zipfile
+
+        import wget
+
+        wget.download(
+            "https://huggingface.co/datasets/ArtificialAnalysis/AA-LCR/resolve/main/extracted_text/AA-LCR_extracted-text.zip"
```
shtoshni: Can you make this URL a constant at the top of the file. Something like "DOWNLOAD_URL"

shtoshni: we can move the imports to the top as well, right?

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,253 @@
...
+
+            # Check if they match after normalization
+            if normalized_target == actual_file or target_filename == normalized_actual:
+                if debug:
+                    print(f"DEBUG find_actual_file: found directory match {repr(actual_file)}")
+                return actual_file
+    except OSError:
+        pass
+
+    if debug:
```
shtoshni: Ideally we should use the [LOG](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/bfcl_v3/prepare.py#L38) to indicate the level of messages.  `LOG.debug` can be used for debugging messages.

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,253 @@
...
+            entry["expected_answer"] = entry.pop("answer")
+            entry["expected_judgement"] = "correct"  # for judgement metric
+            # remove unused columns
+            entry.pop("data_source_urls")
+
+            json.dump(entry, fout)
+            fout.write("\n")
+
+
+def get_aalcr_data(max_context_window, setup, tokenizer_name):
```
shtoshni: maybe a more appropriate name is `prepare_aalcr_data`?

nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -0,0 +1,253 @@
...
+
+            entry[f"n_tokens_{tokenizer_name}"] = n_tokens
+            entry["question"] = question
+            entry["expected_answer"] = entry.pop("answer")
+            entry["expected_judgement"] = "correct"  # for judgement metric
+            # remove unused columns
+            entry.pop("data_source_urls")
+
+            json.dump(entry, fout)
+            fout.write("\n")
```
shtoshni:
```suggestion
            fout.write(json.dumps(entry) + "\n")
```

nemo_skills/dataset/aalcr/prepare.py
fayejf: they are only used here and only when lcr does not exist. i would prefer putting here incase some missing package issue (very unlikely issue though)

PR #801, opened 2025-09-09
lean eval last code block and has sorry logic fix

stephencge:
extract last code block instead of first

"proof_status" = "has_sorry" when sorry explicitly found in stdout/err, allows for warnings to pass

### DIFF
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -245,14 +245,14 @@ class Sandbox(abc.ABC):
             output = await self._send_request(request, timeout)
         except httpx.TimeoutException:
             return "timeout"
-        if output.get("process_status") == "completed":
-            stdout = (output.get("stdout") or "").lower()
-            stderr = (output.get("stderr") or "").lower()
+        if output["process_status"] == "completed":
+            stdout = output["stdout"].lower()
+            stderr = output["stderr"].lower()
             combined = stdout + "\n" + stderr
             if re.search(r"\bsorry\b", combined) is not None:
                 return "has_sorry"
             return "completed"
-        return output.get("process_status", "error")
+        return output["process_status"]
 
     async def batch_evaluate_results(
         self,
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -245,9 +245,14 @@ class Sandbox(abc.ABC):
             output = await self._send_request(request, timeout)
         except httpx.TimeoutException:
             return "timeout"
-        if output["process_status"] == "completed" and output["stdout"] != "":
-            return "has_sorry"
-        return output["process_status"]
+        if output.get("process_status") == "completed":
+            stdout = (output.get("stdout") or "").lower()
+            stderr = (output.get("stderr") or "").lower()
+            combined = stdout + "\n" + stderr
+            if re.search(r"\bsorry\b", combined) is not None:
+                return "has_sorry"
+            return "completed"
+        return output.get("process_status", "error")
 
     async def batch_evaluate_results(
         self,
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -264,7 +264,7 @@ class Sandbox(abc.ABC):
         final_answer_key: str = "**FINAL ANSWER**",
         restate_formal_statement: bool = True,
         strip_theorem_from_proof: bool = True,
-        code_block_from: str = "last",
+        extract_code_mode: str = "last",
     ):
         """Evaluate results and write back to original files."""
 
@@ -283,7 +283,7 @@ class Sandbox(abc.ABC):
             if answer_format == "lean4-proof":
                 if not use_predicted_proof_key:
                     generation = clean_formal_generation(
-                        line_dict["generation"], final_answer_key=final_answer_key, code_block_from=code_block_from
+                        line_dict["generation"], final_answer_key=final_answer_key, extract_code_mode=extract_code_mode
                     )
                     line_dict["predicted_proof"] = (
                         line_dict["header"]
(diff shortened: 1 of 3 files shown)


nemo_skills/code_execution/utils.py
```diff
@@ -91,9 +91,9 @@ def extract_code_block(text: str, languages=None) -> str:
     if languages is None:
         languages = [""]
     for language in languages:
-        match = re.search(rf"```{language}\s*\n?(.*?)\n?```", text, re.DOTALL)
```
gwarmstrong, line 94: please make the forward/backward search into a parameter since we may want have different constraints in different settings (e.g., sometime LLM will have a sloppy end with extra code blocks etc.)

gwarmstrong (changes requested):
Small request to make search method parameterizable

nemo_skills/code_execution/utils.py
stephencge: parametrized to preserve default behavior, extracting last only for lean

nemo_skills/code_execution/sandbox.py
```diff
@@ -245,9 +245,14 @@ async def is_proof_correct(self, pred_output, timeout=30.0):
             output = await self._send_request(request, timeout)
         except httpx.TimeoutException:
             return "timeout"
-        if output["process_status"] == "completed" and output["stdout"] != "":
-            return "has_sorry"
-        return output["process_status"]
+        if output.get("process_status") == "completed":
```
Kipok: why .get here and below? We always expect these fields to be populated, right?

nemo_skills/code_execution/utils.py
```diff
@@ -87,23 +87,28 @@ def extract_code_output(generation: str, code_output_begin: str, code_output_end
     return _extract_between_separators(generation, [code_output_begin, code_output_end], extract_all)
 
 
-def extract_code_block(text: str, languages=None) -> str:
+def extract_code_block(text: str, languages=None, which: str = "last") -> str:
```
Kipok: maybe extract_mode instead of which?

nemo_skills/inference/model/parallel_thinking.py
```diff
@@ -58,8 +59,8 @@ class ParallelThinkingConfig:
     # GenSelect vs GenSynthesis
     mode: str | None = None  # genselect or gensynthesis
 
-    genselect: GenSelectSpecificConfig = GenSelectSpecificConfig()
-    gensynthesis: GenSynthesisSpecificConfig = GenSynthesisSpecificConfig()
+    genselect: GenSelectSpecificConfig = field(default_factory=GenSelectSpecificConfig)
```
Kipok: isn't this merged in another PR?

nemo_skills/code_execution/sandbox.py
stephencge: yes expect them to be populated, changed below

nemo_skills/inference/model/parallel_thinking.py
stephencge: yes didn't git properly

nemo_skills/code_execution/utils.py
stephencge: changed below

nemo_skills/code_execution/utils.py
```diff
@@ -87,23 +87,28 @@ def extract_code_output(generation: str, code_output_begin: str, code_output_end
...
+            idx = 0 if extract_mode == "first" else -1
+            return matches[idx].strip()
     return ""
 
 
-def clean_formal_generation(generation: str, final_answer_key: str = "**FINAL ANSWER**") -> str:
+def clean_formal_generation(
+    generation: str,
+    final_answer_key: str = "**FINAL ANSWER**",
+    code_block_from: str = "first",
```
Kipok:
```suggestion
    code_block_from: str = "last",
```
using last by default makes more sense I guess?

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -53,6 +53,8 @@ class LeanEvaluatorConfig:
     timeout: float = 30.0
     final_answer_key: str = "**FINAL ANSWER**"
     restate_formal_statement: bool = True
+    # Which code block to extract when multiple are present: "first" or "last"
+    code_block_from: str = "last"
```
Kipok: maybe also rename this one for consistency? Maybe we use the same name for this argument in all places, I guess something like extract_code_mode. Not sure what's the best name, but keeping it consistent seems like a good idea

Kipok (approved):
just a small suggestion to change the name and can merge after that

nemo_skills/code_execution/utils.py
stephencge: I went with first to match the previous re.search behavior and passed through last just for lean

Kipok: I think that's the only place where this is used, so I'd still change the default as using last seems like a safer default option

nemo_skills/code_execution/utils.py
```diff
@@ -87,23 +87,28 @@ def extract_code_output(generation: str, code_output_begin: str, code_output_end
...
+            idx = 0 if extract_code_mode == "first" else -1
+            return matches[idx].strip()
     return ""
 
 
-def clean_formal_generation(generation: str, final_answer_key: str = "**FINAL ANSWER**") -> str:
+def clean_formal_generation(
+    generation: str,
+    final_answer_key: str = "**FINAL ANSWER**",
+    extract_code_mode: str = "first",
```
Kipok:
```suggestion
    extract_code_mode: str = "last",
```

nemo_skills/code_execution/utils.py
Kipok: Merging with the change, but if you or @gwarmstrong think the default should be first, let's update it in another pr

PR #802, opened 2025-09-09
Check for existing config in custom config dir before reading

activatedgeek:


Kipok (commented):
what's the motivation for this change? I think the current behavior of failing if file not found is correct as if users specify either env var or explicit dir, we want to take config from there and not from other places. And if it's not there, we should fail instead of falling back, I think. Users can remove those parameters if they want to use other locations. Otherwise it's easy to have a silently incorrect behavior if they have a typo in config name and it falls back to another folder they aren't expecting

PR #803, opened 2025-09-09
FIX Cleanup Sessions on timeout

gwarmstrong:
The current sandbox has issues with python code that causes timeouts--the timed out cells are continued until completion, which can block a worker indefinitely. This PR fixes this so sessions that timeout are killed and restarted (with no previous session information/memory) on the next execution. This will help prevent long-running jobs from stalling out many other runs.

We plan to add a more robust resume-on-timeout mechanism, but this should at least reduce the error rate for the time-being.

### DIFF
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -238,6 +238,8 @@ class Sandbox(abc.ABC):
         if output.get("process_status") == "completed":
             self.session_histories[request_session_id].append(generated_code)
 
+        output.pop("new_session_created", None)
+
         return output, request_session_id
 
     async def is_proof_correct(self, pred_output, timeout=30.0):


tests/test_session_affinity.py
```diff
@@ -501,6 +501,168 @@ def test_session_list_endpoint(self):
                 assert "last_used" in session_info
                 assert "alive" in session_info
 
+    def test_infinite_loop_timeout_then_simple_job(self, tester):
```
Kipok, line 504: can you please add a test that checks that state is restored. So something like 

first cell:
a = 2

second cell:
something that timeouts

third cell:
print(a + 4)

should work correctly as `a` from first cell should persist in a session

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -117,8 +117,20 @@ def strip_ansi_codes(text):
...
-def execute_ipython_session(generated_code, session_id, traceback_verbosity="Plain"):
-    """Execute Python code in a persistent IPython session"""
+def cleanup_session(session_id):
+    """Clean up and remove a specific session"""
+    with session_lock:
+        if session_id in sessions:
+            try:
+                del sessions[session_id]
+                logging.info(f"Cleaned up session after timeout: {session_id}")
+            except Exception as e:
```
Kipok: should we instead fail if it goes here? This would indicate something is very broken, so feels safer to just fail with error?

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -129,31 +141,77 @@ def execute_ipython_session(generated_code, session_id, traceback_verbosity="Pla
...
+                    captured.stderr if "captured" in locals() else ""
+                ) + f"\n{type(e).__name__}: {e}"
+
+        # Start execution in a separate thread
+        execution_thread = threading.Thread(target=execute_with_capture)
+        execution_thread.daemon = True  # Allow main thread to exit even if this is still running
+        execution_thread.start()
+
+        # Wait for completion or timeout
+        execution_thread.join(timeout=timeout)
```
Kipok: what happens to this thread after we return from this function? Does it keep running or nginx cleans it up somehow?

gwarmstrong: yeah this is a good call out, ti will keep running until it completes. I am working on a fix to better manage canceling the process running the cell.

tests/test_session_affinity.py
gwarmstrong: okay we have tests like this now for the updated functionality that handles this case

gwarmstrong: they are separate from this test though because it's really two different units of functionality

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
gwarmstrong: okay, this one was tricky 😅 had to switch to multiprocessing to make it work, which requires a lot of lot more management, but it seems to be working okay on the unit tests at this point

nemo_skills/code_execution/sandbox.py
```diff
@@ -216,6 +216,21 @@ async def execute_code(
             output = await self._send_request(request, timeout)
         except httpx.TimeoutException:
             output = {"process_status": "timeout", "stdout": "", "stderr": "Timed out\n"}
+            if session_id is not None:
+                history = self.session_histories.get(session_id, [])
+                combined_code = "\n".join(history) + ("\n" if history else "") + generated_code
```
Kipok: should generated_code not be here? Otherwise it's guaranteed to hit timeout again?

Kipok: if this is indeed a bug, how do the tests pass with that?

gwarmstrong: yeah it should be removed--in fact what was happening was the error was being silence because the ShellManager is able to salvage the job itself and pick back up with the existing memory. We can actually remove the whole re-execution here because the ShellManager seems pretty robust at handling it.

Kipok: I think it should still be there as a fall-back? I think the original intention was to add this when the whole worker dies in nginx, which just happens sometimes, so for those cases this can still be used to restore the state and actually it does make sense to keep the final code there as long as the status is ok, so we should have a check for that. But maybe for timeout this isn't required

Kipok: so, the newly added if, can be removed, but this existing one should be updated I guess to check the status before adding generated_code and not add it in case of errors https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/code_execution/sandbox.py#L224

gwarmstrong: like this? https://github.com/NVIDIA/NeMo-Skills/pull/803/commits/7282778aeed5f5fc1fb859209cdf03a2f9b124a1

nemo_skills/code_execution/sandbox.py
```diff
@@ -221,7 +221,10 @@ async def execute_code(
         # Rebuild state by executing concatenated history
         if session_id is not None and new_session_created:
             history = self.session_histories.get(session_id, [])
-            combined_code = "\n".join(history) + ("\n" if history else "") + generated_code
+            add_generated_code = output.get("process_status") not in ["timeout", "interrupted"]
```
Kipok: actually, I'm not sure now. If we do this, we will return the output of the previous code, which is also not correct. So probably it's ok to keep the original logic? It only fails if the worker crashes two times in a row, which is hopefully not that likely? Alternatively, we should rerun the history without final code always, but return the output from the previous run

gwarmstrong: I don't think you can ensure you're properly restoring state in this manner anyways. It kind of assumes everything is valid and runs properly? E.g.,

```python
# Block 1
x = 1
print(x)
```
```python
# Block 2
y = 2
z = 1 / 0  # raises ZeroDivisionError
print(z)
```
```python
# Block 3
w = x + y
print(w)
```

vs.

```python
x = 1
print(x)
y = 2
z = 1 / 0  # raises ZeroDivisionError
print(z)
w = x + y
print(w)
```

you'll get two different results.

gwarmstrong: So that's to say, for now in this PR, why don't we minimize the change and next week we can revisit the replays/history  with @i-vainn

Kipok: but errors are never saved in history. We only save successful completions. But the current code is incorrect anyway because it will re-print all previous outputs, but we only want the last one. So we should indeed probably switch to rerun history, but reuse output of previous run or rerun in two calls, one for history and then another for generated_code, right?

Kipok: ok, sounds good. Can you please add a TODO there and point to this thread?

gwarmstrong: added https://github.com/NVIDIA/NeMo-Skills/pull/803/commits/91a287ceab18eac9a38896bef635ee3aee26215b

PR #805, opened 2025-09-10
Add build stage for docker images

gwarmstrong:
Specifically builds `nemo-skills` and `nemo-skills-sandbox` containers on pull requests


.github/workflows/tests.yml
```diff
@@ -29,13 +29,16 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -e .[dev]
+    - name: Build Images
+      run: |
+        # these tags need to match the ones in tests/gpu-tests/test-local.yaml
+        docker build -t igitman/nemo-skills:0.7.0 -f dockerfiles/Dockerfile.nemo-skills .
```
Kipok, line 46: could we only do this if certain files are changed to avoid rebuilding everytime unnecessarily? I think it's only needed if either requirements, dockerfile or sandbox file changes

gwarmstrong: Yeah I think that's a good enough trigger at the moment. I spent some time trying to see if I could use a build cache, which is ideal in principle because it essentially let's docker figure out what needs to be rebuilt. But there are some practical hurdles and it appears to slow down builds a bit still, so yeah I think an approximate trigger on the files changed should be fine.

The other downside will be that jobs triggered by other workflows after this is merged will not be rebuilt--we will just have to rely on pushing the new sandbox shortly thereafter.

.github/workflows/tests.yml
```diff
@@ -25,17 +25,35 @@ jobs:
       with:
         python-version: "3.10"
         cache: pip
+    - name: Detect Docker changes
+      id: changes
+      uses: dorny/paths-filter@v3
+      with:
+        filters: |
+          docker:
+            - 'dockerfiles/**'
```
Kipok: should we be more specific to only list two dockerfiles that we are using here?

.github/workflows/tests.yml
Kipok: Not fully sure I understand this. If main gets merged to those, they will get these changes, right?

gwarmstrong: Are the build and push from `main` automated or do we trigger it manually?

Kipok: which build and push are you referring to? The containers are always built and pushed manually currently

gwarmstrong: yeah so,
1. PR 1 makes changes to sandbox
  a. Sandbox is rebuilt during PR
  b. PR 1 is merged
2. PR 2 doesn't change sandbox, but depends on changes from PR 1
  a. Sandbox is not rebuilt because it's sandbox/docker/requirements don't differ from `main`
  b. `main` hasn't been updated, so PR 2 get's old sandbox
3. `main` nemo-skills and sandbox are `built` and pushed
4. Now new PR 2 builds will pull new sandbox container

Kipok: makes sense. Yeah, the idea is that we will not merge until we update and re-push the container manually, so hopefully this won't be an issue.

gwarmstrong: cool, yeah it's not really a regression on current behavior, just wanted to call it out

.github/workflows/tests.yml
gwarmstrong: updated in a6083709

PR #811, opened 2025-09-16
Upgrade vllm container to v0.10.2

Kipok:


Kipok:
running into problems with qwen2.5 generations in slurm tests, will not merge for now, need more time to debug or we can skip this vllm version

PR #813, opened 2025-09-16
update miniF2F dataset

stephencge:
update miniF2F dataset to kimina error corrected version (14 problems), default prompt to lean4/formal-proof-deepseek-prover-v2, and preparation to match prompt (expect formal_statement end in := by\n with no sorry as that is in prompt)

### DIFF
--- a/nemo_skills/dataset/minif2f/prepare.py
+++ b/nemo_skills/dataset/minif2f/prepare.py
@@ -18,12 +18,79 @@ import os
 import urllib.request
 from pathlib import Path
 
-URL = "https://raw.githubusercontent.com/deepseek-ai/DeepSeek-Prover-V1.5/main/datasets/minif2f.jsonl"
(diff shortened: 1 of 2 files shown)


nemo_skills/dataset/minif2f/prepare.py
```diff
@@ -18,24 +18,94 @@
...
     test_data = []
 
     with open(input_file, "r", encoding="utf-8") as file:
         for line in file:
+            if not line.strip():
+                continue
             entry = json.loads(line)
-            if entry["split"] == "valid":
+            entry = process_entry(entry)
+            if entry.get("split") == "valid":
```
Kipok: why .get? Feels like it might lead to silent data loss if something changes in the format. Better to explicitly error out?

Kipok (commented):
thanks! Just one small comment

PR #814, opened 2025-09-16
add MathOlympiadBench

stephencge:
add MathOlympiadBench dataset from Goedel-Prover-V2 paper

### DIFF
--- a/nemo_skills/dataset/mobench/prepare.py
+++ b/nemo_skills/dataset/mobench/prepare.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
 import json
 import os
 import re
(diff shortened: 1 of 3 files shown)


nemo_skills/dataset/mobench/prepare.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Clean up original
+    try:
+        os.remove(original_file)
+    except OSError:
+        pass
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
Kipok: why empty parser? If no arguments, we can remove this and next line

Kipok (commented):
Thanks!

nemo_skills/dataset/mobench/prepare.py
stephencge: removing these 2 lines and import in following commit

Kipok:
one more thing, can you please add this new benchmark to https://github.com/NVIDIA/NeMo-Skills/blob/main/docs/evaluation/formal-math.md?plain=1#L20?

Kipok (approved):

PR #815, opened 2025-09-16
Fix tool calling

Kipok:
A few fixes for tool-calling
1. Fix a bug where tool calls were always added as a new assistant turn, instead of being appended to the previous turn
2. Add reasoning_content back into conversation if it's available
3. Add request_id parameter that's always passed in extra_args to support stateful tools
4. Update python tool to use request id as well as to return the string output of stdout/stderr instead of dict
5. Remove title from tool definition as it's not meant to be shown to the model
6. Remove python class from tool name. Otherwise, it's not possible to match tool names as the model might be trained with.
7. Rename python tool and update description

### DIFF
--- a/nemo_skills/inference/model/tool_call.py
+++ b/nemo_skills/inference/model/tool_call.py
@@ -142,7 +142,6 @@ class ToolCallingWrapper:
                 tool_calls_message = self.call_interpreter.parse(tool_calls)
                 conversation[-1].update(tool_calls_message)
 
-                ## TODO(sanyamk): refactor to not rely on hardcoded dict keys.
                 tool_calls_output_messages = await self._execute_tool_calls(
                     tool_calls_message["tool_calls"], request_id=request_id
                 )
--- a/nemo_skills/inference/model/tool_call.py
+++ b/nemo_skills/inference/model/tool_call.py
@@ -120,29 +120,12 @@ class ToolCallingWrapper:
         while True:
             if isinstance(tokens_to_generate, int) and tokens_to_generate <= 0:
                 break
-            print("!!", conversation)
-            from transformers import AutoTokenizer
-
-            tok = AutoTokenizer.from_pretrained(
-                "/home/igitman/workspace/openreasoning-toolcall-hf/", trust_remote_code=True
-            )
-            conv_to_print = copy.deepcopy(conversation)
-            for conv in conv_to_print:
-                if "tool_calls" in conv:
-                    conv["tool_calls"][0]["function"]["arguments"] = json.loads(
-                        conv["tool_calls"][0]["function"]["arguments"]
-                    )
-            prompt_str = tok.apply_chat_template(
-                conv_to_print, tools=tools, tokenize=False, add_generation_prompt=False
-            )
-            print("@@", prompt_str)
             generation = await self.model.generate_async(
                 prompt=conversation,
                 tools=tools,
-                tokens_to_generate=16000,
+                tokens_to_generate=tokens_to_generate,
                 **generation_kwargs,
             )
-            print("##", generation)
             if isinstance(tokens_to_generate, int):
                 tokens_to_generate -= generation["num_generated_tokens"]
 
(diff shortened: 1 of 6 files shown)


nemo_skills/inference/model/tool_call.py
```diff
@@ -113,33 +114,38 @@ async def generate_async(
...
-            conversation.append({"role": "assistant", "content": result_steps["generation"][-1]})
+            conversation.append({"role": "assistant", "content": generation["generation"]})
+            if "reasoning_content" in generation:
+                conversation[-1]["reasoning_content"] = generation["reasoning_content"]
 
             tool_calls = generation.get("tool_calls", [])
             if tool_calls:
                 tool_calls_message = self.call_interpreter.parse(tool_calls)
-                conversation.append(tool_calls_message)
+                conversation[-1].update(tool_calls_message)
```
gwarmstrong, line 143: In Qwen3 at least, these are supposed to be a separate message, with "role": "tool", and "content" with the response of the tool. Is it different for some other model, do we need to encapsulate that?

Kipok: this is not the output of the tool, this is tool calls part. So before the conversation was formatted like
```
{role: assistant, content: <>}
{role: assistant, content: '', tool_calls: <>}
{role: tool, content: <>}
```
after this change it will be
```
{role: assistant, content: <>, tool_calls: <>}
{role: tool, content: <>}
```
which I believe is the right way, otherwise there is an extra closing/opening tag there

gwarmstrong: ah I see, yeah that's right

nemo_skills/inference/model/tool_call.py
```diff
@@ -113,33 +114,38 @@ async def generate_async(
...
             if isinstance(tokens_to_generate, int):
                 tokens_to_generate -= generation["num_generated_tokens"]
 
             for k in ["generation", "num_generated_tokens", "reasoning_content", "finish_reason"]:
                 if k in generation:
                     result_steps[k].append(generation[k])
 
-            conversation.append({"role": "assistant", "content": result_steps["generation"][-1]})
+            conversation.append({"role": "assistant", "content": generation["generation"]})
+            if "reasoning_content" in generation:
```
activatedgeek, line 137: Just confirming, does this line align with the typical expectation, e.g. from [Qwen Models](https://qwen.readthedocs.io/en/latest/getting_started/concepts.html#thinking):

> The thinking block should only be included in the final round except for multi-step tool calls.

nemo_skills/inference/model/tool_call.py
```diff
@@ -113,33 +114,38 @@ async def generate_async(
...
+                conversation[-1]["reasoning_content"] = generation["reasoning_content"]
 
             tool_calls = generation.get("tool_calls", [])
             if tool_calls:
                 tool_calls_message = self.call_interpreter.parse(tool_calls)
-                conversation.append(tool_calls_message)
+                conversation[-1].update(tool_calls_message)
 
                 ## TODO(sanyamk): refactor to not rely on hardcoded dict keys.
```
activatedgeek: We can remove this TODO because this block of code should always rely on the `ChatCompletionMessageToolCall` object. For other models, it will be a separate block of code.

nemo_skills/inference/model/tool_call.py
Kipok: I believe this is handled by their chat template. So we should always give full conversation history, but the model will strip it away in jinja

nemo_skills/inference/model/tool_call.py
```diff
@@ -63,7 +64,7 @@ def __init__(
         self.call_interpreter = ChatCompletionCallInterpreter()
         self.response_formatter = ChatCompletionResponseFormatter()
 
-    async def _execute_tool_call(self, tool_call):
+    async def _execute_tool_call(self, tool_call, request_id: str):
         ## TODO(sanyamk): The correct key format needs to be cohesive with other formatters.
```
activatedgeek, line 68: We can remove this TODO and we should standardize always using the tool calls format, whether it comes from manual parsing on native support from model servers.

activatedgeek (approved):
Just using this opportunity to remove some dangling TODOs that are not required. Looks fine to me otherwise.

PR #817, opened 2025-09-18
Update filters.trim_solutions=false to be default

Kipok:
Remove trim_solutions=true by default as that's no longer the right setting for most models. We needed it before for models that didn't finish well, but now this can just cut generations mid-sentence and force the final models to always stop after boxed.


Kipok:
fyi @ekmb - not sure if you're using that script for your data preparation, but if you do, the default behavior will be changed. But the previous default was actually very bad, so you might need to rerun data prep if you had it set to true.

PR #825, opened 2025-09-18
Asynchronous eval in Generation Loop

gwarmstrong:
# Add Evaluation Support During Generation

## Summary

This PR introduces a new evaluation system that enables evaluation during generation for supported evaluators, while maintaining backward compatibility with existing batch evaluation workflows. This is helpful for things like Lean evaluation which are cpu intesnive and can affect the wall time significantly if run altogether after generation is complete.

### Core Architecture
- Converted evaluation functions to classes with `BaseEvaluator` interface
- Auto-detection of single eval support using method overriding pattern
- Dual evaluation modes: Single evaluation during generation + batch evaluation after generation
- Moved Lean evaluation machinery in the sandbox to shared utilities

### New Components

**BaseEvaluator Class** (`nemo_skills/evaluation/evaluator/base.py`)
- Abstract base class with `eval_full` (required) and `eval_single` (optional) methods
- Auto-detects single evaluation support via `supports_single_eval()` method

**Converted Evaluators** (`nemo_skills/evaluation/evaluator/math.py`)
- `MathEvaluator`: Batch-only evaluation
- `Lean4ProofEvaluator`: Supports evaluation during generation
- `Lean4StatementEvaluator`: Batch-only evaluation

**Generation Integration** (`nemo_skills/inference/generate.py`)
- Added `eval_type` and `eval_config` parameters to `GenerateSolutionsConfig`
    - They are optional and will run single data point evaluation when specified
- Evaluation hook in `process_single_datapoint()` for evaluation during generation loop
- Only single-eval-capable evaluators can be used during generation

**Pipeline** (`nemo_skills/pipeline/utils/eval.py`)
- Automatic strategy selection between single and batch evaluation--deferring to single if it is available
- Command construction that integrates eval into generation or adds separate eval command depending on available methods

## Usage Examples

```bash
# Direct generation with evaluation
python -m nemo_skills.inference.generate \
    input_file=problems.jsonl \
    output_file=solutions.jsonl \
    eval_type=lean4-proof \
    eval_config.timeout=30
```

### Pipeline
```python
# eval pipeline will work seamlessly while using new generation-loop evaluation
from nemo_skills.pipeline.cli import eval
eval(
    ctx=wrap_arguments(
        f"++inference.tokens_to_generate={38 * 1024} "
        "++inference.top_p=0.95 "
        "++inference.temperature=1.0 "
    ),
    benchmarks="minif2f:32",
    cluster=cluster,
    model=model,
    server_type='vllm',
    server_gpus=8,
    server_nodes=1,
    server_args='--max-model-len 40960 ',
    output_dir=output_dir,
    with_sandbox=True,
    partition=partition,
)
```

## Backward Compatibility

- All existing evaluation functions preserved and work unchanged
- Existing pipelines and scripts continue to work without modification
- Function-based evaluators remain in `EVALUATOR_MAP` for batch-only evaluation

## Migration Guide

To add generation-loop evaluation support to an existing evaluator:

1. Convert function to class inheriting from `BaseEvaluator`
2. Move logic from function to `eval_full` method
3. Implement `eval_single` for real-time evaluation (optional)
4. Add to `EVALUATOR_CLASS_MAP` and remove from `EVALUATOR_MAP`
5. Update legacy function to delegate to class

## Summary by CodeRabbit

- New Features
  - Shared Lean4 proof utilities for building/extracting/formatting proofs and inferring proof status.
  - Class-based evaluator framework with a common BaseEvaluator and concrete Math / Lean4-proof / Lean4-statement evaluators.
  - Generation tasks can accept eval_type/eval_config and optionally run per-item evaluation during generation.
  - Eval-args parsing to support in-generation vs. batch evaluation.

- Refactor
  - Centralized proof handling and concurrency improvements; evaluator flow unified around class-based APIs.

- Breaking Changes
  - Math grading API changed from batch file-based processing to a per-item evaluate_result API.

### DIFF
--- a/nemo_skills/evaluation/evaluator/base.py
+++ b/nemo_skills/evaluation/evaluator/base.py
@@ -37,7 +37,6 @@ class BaseEvaluator(ABC):
 
         Args:
             input_files: List of input files to evaluate
-            **kwargs: Additional evaluation parameters
         """
         semaphore = asyncio.Semaphore(self.num_parallel_requests)
         for input_file in tqdm.tqdm(unroll_files(input_files), desc="Processing files"):


nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -60,9 +63,28 @@ def dummy_eval(cfg):
     "bigcodebench": eval_bigcodebench,
 }
 
+# Evaluator class mapping
+EVALUATOR_CLASS_MAP = {
+    "math": MathEvaluator,
+    "lean4-proof": Lean4ProofEvaluator,
+    "lean4-statement": Lean4StatementEvaluator,
+    # Other evaluators can be added here as they're converted to classes
```
Kipok, line 68: can you please create an issue to track that?

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -72,9 +94,41 @@ def register_evaluator(eval_type: str, eval_fn: Callable[[Dict[str, Any]], None]
     EVALUATOR_MAP[eval_type] = eval_fn
 
 
-def evaluate(cfg):
-    if cfg.eval_type not in EVALUATOR_MAP:
+def get_evaluator(eval_type: str, config: Dict[str, Any]) -> BaseEvaluator:
```
Kipok: should this be called get_evaluator_class?

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -72,9 +94,41 @@ def register_evaluator(eval_type: str, eval_fn: Callable[[Dict[str, Any]], None]
...
+            f"All supported types: {list(EVALUATOR_MAP.keys())}"
         )
-    return EVALUATOR_MAP[cfg.eval_type](cfg)
+
+    evaluator_class = EVALUATOR_CLASS_MAP[eval_type]
+    return evaluator_class(config)
+
+
+def supports_single_eval(eval_type: str, config: Dict[str, Any]) -> bool:
+    """Check if evaluator supports single evaluation during generation."""
```
Kipok:
```suggestion
    """Check if evaluator supports single data point evaluation during generation."""
```

nemo_skills/evaluation/evaluator/base.py
```diff
@@ -0,0 +1,54 @@
...
+
+class BaseEvaluator(ABC):
+    """Base class for all evaluators."""
+
+    def __init__(self, config: Dict[str, Any]):
+        """Initialize evaluator with configuration."""
+        self.config = config
+
+    @abstractmethod
+    async def eval_full(self, input_files: List[str], **kwargs) -> None:
```
Kipok: should this not be abstract and have a default implementation of calling eval_single in parallel on all samples? If some classes don't support eval_single, they will override

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -58,30 +60,108 @@ class LeanEvaluatorConfig:
...
+    """Legacy function - delegates to Lean4StatementEvaluator class."""
+    evaluator = Lean4StatementEvaluator(cfg.eval_config)
+    asyncio.run(evaluator.eval_full(cfg.input_files))
+
+
+# Evaluator Classes
+
+
+class MathEvaluator(BaseEvaluator):
+    """Math evaluator - only supports batch evaluation."""
```
Kipok: could you add an issue for this? No reason to only support batch eval here

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -58,30 +60,108 @@ class LeanEvaluatorConfig:
...
+
+    async def eval_full(self, input_files: list[str], **kwargs) -> None:
+        """Batch evaluate math problems."""
+        eval_config = MathEvaluatorConfig(**self.config)
+        eval_config_dict = asdict(eval_config)
+        batch_evaluate_results(
```
Kipok: Generally, with your new interface I think we should completely remove this function and move it inside the class. But can be done later

nemo_skills/evaluation/evaluator/base.py
Kipok: I guess we could also do this later

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -58,30 +60,108 @@ class LeanEvaluatorConfig:
...
 def eval_lean4_proof(cfg):
-    eval_config = LeanEvaluatorConfig(**cfg.eval_config)
-
-    sandbox = get_sandbox(**eval_config.sandbox)
-    eval_config_dict = asdict(eval_config)
-    eval_config_dict.pop("sandbox")
-    asyncio.run(
-        sandbox.batch_evaluate_results(
-            input_files=cfg.input_files,
+    """Legacy function - delegates to Lean4ProofEvaluator class."""
```
Kipok: why do we need these two if they are already defined as classes?

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -58,30 +60,108 @@ class LeanEvaluatorConfig:
...
+class Lean4ProofEvaluator(BaseEvaluator):
+    """Lean4 proof evaluator - supports both single and batch evaluation."""
+
+    async def eval_full(self, input_files: list[str], **kwargs) -> None:
+        """Batch evaluate Lean4 proofs."""
+        eval_config = LeanEvaluatorConfig(**self.config)
+        sandbox = get_sandbox(**eval_config.sandbox)
+        eval_config_dict = asdict(eval_config)
+        eval_config_dict.pop("sandbox")
+        await sandbox.batch_evaluate_results(
```
Kipok: ideally we don't use this but instead directly call eval_single in parallel. This will minimize code duplication

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -58,30 +60,108 @@ class LeanEvaluatorConfig:
...
+            # Determine proof status using shared utility
+            proof_status = determine_proof_status(output)
+
+            return {
+                "predicted_proof": predicted_proof,
+                "proof_status": proof_status,
+                "lean_evaluation": {**output, "timeout": eval_config.timeout},
+            }
+
+        except Exception as e:
```
Kipok: I'm not sure having blank exception is good here. I'd rather let it fail, so that users quickly see what's wrong. Otherwise maybe there is 10% of samples where we have some bug in eval that would be incorrectly evaluated and it's hard to notice unless you really look into the generation file and spot those

nemo_skills/inference/generate.py
```diff
@@ -502,7 +519,14 @@ async def process_single_datapoint(self, data_point, all_data):
             if self.cfg.override_max_code_executions and self.cfg.total_code_executions_in_prompt is not None:
                 generation_params["max_code_executions"] = data_point["total_code_executions"]
 
-        return await self.llm.generate_async(**generation_params)
+        result = await self.llm.generate_async(**generation_params)
+
+        # Apply evaluation hook if configured
```
Kipok: could you put this into another method and call right after process_dingle_datapoint? This will make it easier to reuse as process_single_datapoint is most commonly overriden

nemo_skills/inference/generate.py
```diff
@@ -502,7 +519,14 @@ async def process_single_datapoint(self, data_point, all_data):
...
+        if self.evaluator:
+            eval_results = await self.evaluator.eval_single({**data_point, **result})
```
Kipok: this might be a bit tricky as e.g. it ignores the generation_key so it might be hard to reuse logic between batch and single evaluation. We can leave this as is I guess but might need to refactor things a bit so that we can ensure the structure of the final data points is identical between batch and single evals

nemo_skills/pipeline/utils/eval.py
```diff
@@ -28,6 +28,66 @@
...
+    eval_type, eval_config = parse_eval_args(eval_args)
+
+    if not eval_type:
+        return False
+
+    try:
+        from nemo_skills.evaluation.evaluator import supports_single_eval
+
+        return supports_single_eval(eval_type, eval_config)
+    except Exception:
```
Kipok: let's not do this, just let it fail, it's at job submission time so best to just report errors and not hide them. And we generally don't expect this to fail, so it's abnormal

nemo_skills/pipeline/utils/eval.py
```diff
@@ -28,6 +28,66 @@
 LOG = logging.getLogger(get_logger_name(__file__))
 
 
+def parse_eval_args(eval_args: str) -> tuple[str | None, dict]:
```
Kipok, line 34: this feels a bit brittle and hacky, but I can't currently think of a better way.. Maybe add a TODO to say that we ideally don't want to rely on custom parsing of the command, but some major refactoring or clever ideas might be needed :)

nemo_skills/pipeline/utils/eval.py
```diff
@@ -28,6 +28,66 @@
...
+
+    Handles Hydra argument formats:
+    - ++eval_type=value (override)
+    - +eval_type=value (new)
+    - eval_type=value (config)
+    """
+    if not eval_args:
+        return None, {}
+
+    import re
```
Kipok: move to top?

Kipok (commented):
Thanks, looks really nice and clean overall! And should give a really great speed boost to heavy evaluations, so very good feature to have. 

Just a few minor comments to address.

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: Just in case any users of NeMo-Skills imports them somewhere it doesn't break their code. But we can also remove, they're not needed internally afaik

Kipok: let's remove, these shouldn't be used anywhere

nemo_skills/inference/generate.py
Kipok: can you add a TODO above about this?

nemo_skills/evaluation/evaluator/__init__.py
gwarmstrong: updated in 9db3e4e7

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: It's a little tricky--I don't love the idea that we would be writing loop that has to account for different structures of input file and writing the eval_single results back to the input files without knowing the structure of the input file or eval_single. And without regard of the file sizes involved or the amount of work in eval_single. There's mechanisms to help with this for sure (like having a max workers param, adapters for the input/output data or assuming it's always dict), but I'd rather us get some examples of different evaluators doing this first before trying to generalize.

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: Removed in 56e32464

nemo_skills/code_execution/sandbox.py
```diff
@@ -284,38 +231,20 @@ async def process_line(line_data):
...
+                restate_formal_statement=restate_formal_statement,
+                strip_theorem_from_proof=strip_theorem_from_proof,
+            )
+
+            line_dict["predicted_proof"] = prepare_predicted_proof_from_line_dict(
+                line_dict=line_dict,
+                config=config,
+                answer_format=answer_format,
+                use_predicted_proof_key=use_predicted_proof_key,
+            )
```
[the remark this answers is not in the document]
gwarmstrong: This is a good find. Our processing code was previously running in series rather than concurrently--so only one proof was getting evaluated at a time.

I tried out this suggestion with the batch eval for proofs and it was faster than the current batch code but slower than single. I'll add it in since we might use this functionality elsewhere, even though we'll do single eval for lean since it still had a much smaller delta on the overall runtime.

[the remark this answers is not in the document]
gwarmstrong: added in 07dae3db24176c8b0ea3bfffe7607e7663f1ccad

nemo_skills/pipeline/utils/eval.py
gwarmstrong: added in 5f56c4ca

nemo_skills/pipeline/utils/eval.py
gwarmstrong: done in 5f56c4ca

nemo_skills/pipeline/utils/eval.py
gwarmstrong: fixed in 5f56c4ca

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: okay I did a simple version in 1220850fa5bf57c9ba804a8808772b4b5b3d9f13
I didn't want to be opinionated on how requests are run concurrently yet since it requires a decision about how to control the number of concurrent requests, so presently they are executed serially, but the code should at least reduce duplication of the read/modify/write loop.

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: Added the single support in 1220850fa5bf57c9ba804a8808772b4b5b3d9f13
And moved the full_eval to use the base batched method there as well

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: done in 1220850fa5bf57c9ba804a8808772b4b5b3d9f13

nemo_skills/inference/generate.py
gwarmstrong: Sorry, I actually don't understand your concern, can you help me understand what you mean by ignoring the generation_key?

nemo_skills/inference/generate.py
gwarmstrong: done

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: Removed in 7385739e

nemo_skills/evaluation/evaluator/math.py
gwarmstrong: okay I removed it here: d99060a67a2f06348cec8f2a54fc2c11f93cee76

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: Okay so I decided to take the parallelization mechanism from the sandbox now, added it here: d99060a67a2f06348cec8f2a54fc2c11f93cee76

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -57,31 +52,67 @@ class LeanEvaluatorConfig:
...
+        )
+
+        # Determine proof status using shared utility
+        proof_status = determine_proof_status(output)
+
+        return {
+            "predicted_proof": predicted_proof,
+            "proof_status": proof_status,
+            "lean_evaluation": {**output, "timeout": eval_config.timeout},
+        }
```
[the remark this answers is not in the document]
gwarmstrong: addressed with single sandbox creation init: af6a5e789afab6806d7764dba8f956dd6017f09e

nemo_skills/evaluation/evaluator/base.py
```diff
@@ -0,0 +1,91 @@
...
+                # TODO we could possibly make this more efficient by allowing concurrent processing, but this is an okay base impl
+                tasks = []
+                for file_line in tqdm.tqdm(fin, total=num_lines, desc=f"Evaluating {os.path.basename(input_file)}"):
+                    line_dict = json.loads(file_line)
+                    task = asyncio.create_task(process_line(line_dict))
+                    tasks.append(task)
+
+                for coro in tqdm.tqdm(asyncio.gather(tasks), total=len(tasks)):
+                    input_data.append(await coro)
+
```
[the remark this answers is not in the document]
gwarmstrong: The order needs to be maintained, so we can't use as_completed, so I should probably use  gather still. Perhaps I just need to await it before iterating?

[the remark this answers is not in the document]
gwarmstrong: wait so if I have a list of tasks, do I even need gather, or can I just await them in sequence I submitted them?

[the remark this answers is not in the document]
gwarmstrong: here I have a fix that does something like this, but also streams the output to file (though whole file is still read into memory): 701592c8310bb7db31130f7d070a08e955a89c03

nemo_skills/inference/generate.py
Kipok: So the evaluation logic might depend on a specific generation_key being used. E.g. for ifeval here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/ifeval/__init__.py#L19. And the reassignment of "generation" to "{generation_key}" happens somewhere after this call. So in that case (and generally in other cases where some postprocessing is being done after this function), the eval_single logic will not match the batch evaluation and we ideally want them to be fully matching

gwarmstrong: thanks for the explanation! TODO added in 0a64101a

nemo_skills/code_execution/sandbox.py
```diff
@@ -284,38 +231,20 @@ async def process_line(line_data):
...
-                    line_dict["predicted_proof"] = header + generation + "\n sorry"
-                else:
-                    if "predicted_proof" not in line_dict:
-                        raise ValueError(
-                            "predicted_proof key not found in the line_dict. "
-                            "Set use_predicted_proof_key=False to re-combine"
-                        )
-            else:
-                raise ValueError(f"Unknown answer_format: {answer_format}")
+            # Prepare predicted_proof using shared utility
```
Kipok, line 234: should we remove this logic completely? This should be handled by the lean eval class now, right?

nemo_skills/evaluation/evaluator/base.py
```diff
@@ -0,0 +1,91 @@
...
+class BaseEvaluator(ABC):
+    """Base class for all evaluators."""
+
+    def __init__(self, config: Dict[str, Any], num_parallel_requests=10):
+        """Initialize evaluator with configuration."""
+        self.config = config
+        self.num_parallel_requests = num_parallel_requests
+
+    async def eval_full(self, input_files: List[str], **kwargs) -> None:
```
Kipok: why do we need **kwargs if they are not used?

nemo_skills/evaluation/evaluator/base.py
```diff
@@ -0,0 +1,91 @@
...
+        Evaluate full dataset in batch mode.
+
+        Args:
+            input_files: List of input files to evaluate
+            **kwargs: Additional evaluation parameters
+        """
+        semaphore = asyncio.Semaphore(self.num_parallel_requests)
+        for input_file in tqdm.tqdm(unroll_files(input_files), desc="Processing files"):
+            # assume that input_file is small enough to entirely fit in the memory
+            with open(input_file, "rt", encoding="utf-8") as f:
```
Kipok: reading full input just to get line count for tqdm feels a little excessive, but I guess it's ok if you think this makes progress bar much better

nemo_skills/code_execution/sandbox.py
Kipok: or I guess it's used for statement check logic. Still could consider removing it completely as we aren't using that logic anymore I think, but it can be done in another PR

nemo_skills/inference/generate.py
```diff
@@ -513,6 +541,12 @@ async def _process_single_datapoint_with_semaphore(self, data_point, all_data, f
 
             # Generate output for this single data point
             output = await self.process_single_datapoint(data_point, all_data)
+            # Apply evaluation hook if configured
+            # TODO: note that this currently only evaluates independently--if there
+            # is any post-processing that needs to be done on the full set of
+            # generations, this will not work correctly, and we might need another
+            # hook at the end of generation to make it work properly
+            output = await self.apply_evaluation_hook({**data_point, **output})
```
Kipok, line 549: actually an even bigger issue than generation_key is that we aren't removing thinking part here, but this is done in batch evaluation here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/evaluation/evaluate_results.py#L74

should we move that logic into the base eval class and apply automatically there? I guess we can do it in a follow up pr, just to merge this sooner

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: I was  thinking forwards compatibility with the eval_full method having parameters  not used by all evaluators, but now that I think about it, it's probably better to encapsulate those in the eval config/evaluator class.

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: I was trying to maintain the existing behavior, but looking at it again, we can just calculate this from the number of tasks created. We still read the whole input file--but we do it once before starting to collect results. There might be merit to only reading in as many lines as you can queue, in which case yeah you would lose the total for the progress bar, but that's not being done anywhere yet, so I'm happy to move that out of scope for this PR.

gwarmstrong: simplified in 740511ee25c809e74b488192d88e868714bea6d8

nemo_skills/evaluation/evaluator/base.py
gwarmstrong: removed in ff28e7ad988a1640e1318cfba046aee765875aa2

nemo_skills/code_execution/sandbox.py
gwarmstrong: Issue created #840

nemo_skills/inference/generate.py
gwarmstrong: Created #841 to track

PR #828, opened 2025-09-19
Add new benchmark SimpleQA to nemo_skills

jiacheng-xu:
Add simpleqa dataset:
Naming: standard named as 'test' and verified as 'verified

```
(py310nemo) (base) NeMo-Skills|jcxu/simpleqa ⇒ ns prepare_data simpleqa
[16:10:09] WARNING  Cluster config is not specified. Running locally without     cluster.py:215
                    containers. Only a subset of features is supported and                     
                    you're responsible for installing any required dependencies.               
                    It's recommended to run `ns setup` to define appropriate                   
                    configs!                                                                   
[16:10:09] INFO     Optional environment variable WANDB_API_KEY not found in     cluster.py:141
                    user environment; skipping.                                                
           INFO     Adding optional environment variable NVIDIA_API_KEY from     cluster.py:132
                    environment                                                                
           INFO     Adding optional environment variable AZURE_OPENAI_API_KEY    cluster.py:132
                    from environment                                                           
           INFO     Optional environment variable OPENAI_API_KEY not found in    cluster.py:141
                    user environment; skipping.                                                
           INFO     Adding optional environment variable HF_TOKEN from           cluster.py:132
                    environment                                                                
           INFO     Adding a task with commands:                                     exp.py:430
           INFO     Adding optional environment variable                         cluster.py:127
                    NEMO_SKILLS_SANDBOX_PORT from config                                       
           INFO     Main command(s): export HYDRA_FULL_ERROR=1 && export             exp.py:507
                    PYTHONPATH=$PYTHONPATH:/nemo_run/code && cd /nemo_run/code &&              
                    python -m nemo_skills.dataset.prepare simpleqa                             
nemo-run/0 Preparing simpleqa
nemo-run/0 Writing verified.jsonl: 100%|██████████| 1000/1000 [00:00<00:00, 84243.27it/s]
```

### DIFF
--- a/nemo_skills/dataset/simpleqa/__init__.py
+++ b/nemo_skills/dataset/simpleqa/__init__.py
@@ -1,0 +1,35 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+DATASET_GROUP = "math"
+METRICS_TYPE = "simpleqa"
+EVAL_ARGS = "++eval_type=math "
+GENERATION_ARGS = "++prompt_config=generic/simpleqa "
+EVAL_SPLIT = "test"
+
+# SimpleQA requires judge model for evaluating factual accuracy
+# Setting openai judge by default, but can be overridden from command line for a locally hosted model
+# Using GPT-4 as recommended for factual evaluation tasks
+
+JUDGE_PIPELINE_ARGS = {
+    "model": "/hf_models/Qwen2.5-32B-Instruct",
+    "server_type": "sglang",
+    "server_gpus": 8,
+    "server_nodes": 1,
+    "server_args": "--context-length 8400",
+}
+
+
+JUDGE_ARGS = "++prompt_config=judge/simpleqa ++generation_key=judgement ++add_generation_stats=False"
--- a/nemo_skills/evaluation/metrics/map_metrics.py
+++ b/nemo_skills/evaluation/metrics/map_metrics.py
@@ -30,10 +30,12 @@ from nemo_skills.evaluation.metrics.lean4_metrics import Lean4Metrics
 from nemo_skills.evaluation.metrics.math_metrics import MathMetrics
 from nemo_skills.evaluation.metrics.mrcr_metrics import MRCRMetrics
 from nemo_skills.evaluation.metrics.ruler_metrics import RulerMetrics
+from nemo_skills.evaluation.metrics.simpleqa_metrics import SimpleQAMetrics
 
 METRICS_MAP = {
     "math": MathMetrics,
     "hle": functools.partial(MathMetrics, compute_no_answer=False, answer_key="generation"),
+    "simpleqa": functools.partial(SimpleQAMetrics, compute_no_answer=False),
     "lean4-proof": Lean4Metrics,
     "lean4-statement": Lean4Metrics,
     "answer-judgement": AnswerJudgementMetrics,
--- a/nemo_skills/evaluation/metrics/simpleqa_metrics.py
+++ b/nemo_skills/evaluation/metrics/simpleqa_metrics.py
@@ -1,0 +1,117 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nemo_skills.evaluation.metrics.base import BaseMetrics
+
+CORRECT_LABEL = "A"
+INCORRECT_LABEL = "B"
+NOT_ATTEMPTED_LABEL = "C"
+
+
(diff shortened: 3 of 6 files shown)


nemo_skills/dataset/simpleqa/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+METRICS_TYPE = "simpleqa"
+EVAL_ARGS = "++eval_type=math "
+GENERATION_ARGS = "++prompt_config=generic/simpleqa "
+EVAL_SPLIT = "test"
+
+# SimpleQA requires judge model for evaluating factual accuracy
+# Setting openai judge by default, but can be overridden from command line for a locally hosted model
+# Using GPT-4 as recommended for factual evaluation tasks
+
+JUDGE_PIPELINE_ARGS = {
```
Kipok, line 26: please set openai as a judge by default as we want this command to work out-of-the-box irrespective of mounts in user's configs

nemo_skills/dataset/simpleqa/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+DATASET_GROUP = "math"
+METRICS_TYPE = "simpleqa"
+EVAL_ARGS = "++eval_type=math "
+GENERATION_ARGS = "++prompt_config=generic/simpleqa "
+EVAL_SPLIT = "test"
```
Kipok: this doesn't correspond to default in prepare.py. Either make them the same or change prepare.py to prepare both test and verified by default (so with "all" argument for split or something like this)

nemo_skills/evaluation/metrics/map_metrics.py
```diff
@@ -30,10 +30,12 @@
 from nemo_skills.evaluation.metrics.math_metrics import MathMetrics
 from nemo_skills.evaluation.metrics.mrcr_metrics import MRCRMetrics
 from nemo_skills.evaluation.metrics.ruler_metrics import RulerMetrics
+from nemo_skills.evaluation.metrics.simpleqa_metrics import SimpleQAMetrics
 
 METRICS_MAP = {
     "math": MathMetrics,
     "hle": functools.partial(MathMetrics, compute_no_answer=False, answer_key="generation"),
+    "simpleqa": functools.partial(SimpleQAMetrics, compute_no_answer=False),
```
Kipok: if you're using a custom class anyway, let's just set compute_no_answer to False there by default and use this without functools.partial

nemo_skills/evaluation/metrics/simpleqa_metrics.py
```diff
@@ -0,0 +1,117 @@
...
+        if j is None:
+            out["not_attempted"] = True
+        else:
+            is_corr = is_correct_judgement_label_matching(j, CORRECT_LABEL)
+            is_inc = is_correct_judgement_label_matching(j, INCORRECT_LABEL)
+            is_na = is_correct_judgement_label_matching(j, NOT_ATTEMPTED_LABEL)
+            out["correct"] = bool(is_corr)
+            out["incorrect"] = bool(is_inc)
+            out["not_attempted"] = bool(is_na)
+
```
Kipok, line 85: should we have a separate metric for missing judgement (when it doesn't match any)?

Kipok:
please also update evaluation docs (not fully sure which category this fits in) https://github.com/NVIDIA/NeMo-Skills/tree/main/docs/evaluation

nemo_skills/dataset/simpleqa/__init__.py
jiacheng-xu:
```
DATASET_GROUP = "science"
METRICS_TYPE = "simpleqa"
EVAL_ARGS = "++eval_type=simpleqa "
GENERATION_ARGS = "++prompt_config=generic/simpleqa "
EVAL_SPLIT = "verified"
```

I changed the `eval_type` to `simpleqa` because I customized a Metric here. Metrics will be loaded from `METRICS_MAP` in `map_metrics.py` so I guess `eval_type` shouldn't matter for loading the metrics.

nemo_skills/evaluation/metrics/simpleqa_metrics.py
jiacheng-xu: Good catch. Following the official implementation, if not parseable, it's considered NOT_ATTEMPTED. Fixed.

nemo_skills/dataset/simpleqa/__init__.py
jiacheng-xu: done.

jiacheng-xu (commented):
Fixed issues @Kipok raised.

nemo_skills/prompt/config/generic/simpleqa.yaml
```diff
@@ -0,0 +1,7 @@
+# Default prompt that does not contain any extra instructions. Can be used for instruction following evals (e.g. ifeval)
+
+system: |-
+  Your response should be in the following format:
```
Kipok: is this some default prompt used for this benchmark? Looks a little strange to me, as in I don't really understand what this means. Also putting instruction in user message is typically better than in system

nemo_skills/prompt/config/generic/simpleqa.yaml
```diff
@@ -0,0 +1,7 @@
+# Default prompt that does not contain any extra instructions. Can be used for instruction following evals (e.g. ifeval)
```
Kipok: please update comment

nemo_skills/evaluation/metrics/simpleqa_metrics.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Returns correctness channels; "correct" is the primary pass@k signal.
+        """
+        d = prediction
+        out = {
+            "correct": False,
+            "incorrect": False,
+            "not_attempted": False,
+        }
+        j = d.get("judgement", None)
```
Kipok: maybe call it `judgement` instead of `j` for better readability?

nemo_skills/prompt/config/generic/simpleqa.yaml
jiacheng-xu: I am looking for a prompt which 1) only yields the answer, not the reasoning process, 2) does not yield bbox. 
I can go ahead and  use `default.yaml`.

nemo_skills/evaluation/metrics/simpleqa_metrics.py
jiacheng-xu: Done.

jiacheng-xu: Sorry I did not see your comments and requested a review 12 mins ago.

Kipok (approved):
thanks!

PR #830, opened 2025-09-20
Update default sandbox parameters

Kipok:
Mainly for better Lean compilation experience. Sometimes we need higher timeouts and default 10gb of memory isn't always enough


Kipok:
let me merge this, but please comment if any concerns @gwarmstrong

PR #833, opened 2025-09-21
NeMo-RL SFT sample printing (to verify if template is applied)

wasiahmad:
Since we can pass SFT data in different format, it is a good suggestion from @Kipok to print a few samples so that we can verify if the templated data is aligned with our expectation. For example, when we are passing data in `messages` format, we need to set:
```
++policy.tokenizer.chat_template="default"
```
Currently, in nemo-rl SFT config, we set tokenizer.chat_template to NULL. This PR simply prints 1 sample per worker.

## Summary by CodeRabbit

* **New Features**
  * Auto-detect training data format and auto-select the appropriate chat/input template; validation data is checked to match.

* **Chores**
  * Default tokenizer template set to enable auto-detection; cleaned log/checkpoint messages (removed emoji).
  * Nemo-rl container image references rolled back to an earlier patch version.

* **Tests**
  * Added message-format data fixture and a GPU test exercising the messages-format SFT flow.


nemo_skills/training/nemo_rl/start_sft.py
```diff
@@ -152,6 +153,19 @@ def sft_preprocessor(
...
 
+    # ==================== START: BLOCK FOR DEBUGGING ====================
+    global _call_counter
+    if _call_counter < 1:  # Only print for the first 3 samples
+        print(f"\n--- 🐛 Debugging invocation #{_call_counter + 1} (for original sample idx: {idx}) ---")
+        # Loop through up to the first 3 messages in the log
+        for i, message in enumerate(message_log[:3]):
+            print(f"  Message [{i}]:")
+            print(f"    Role    : {message['role']}")
+            print(f"    Content : {message['content']}")
```
Kipok, line 202: Could you please add print for token ids as well? Those are also very important to inspect

wasiahmad: @Kipok I think the issue is not in message format or other format. Someone can provide message format data with template applied, in that case, we don't need to have the chat_template. I think for now, we can write a Warning message if `tokenizer.chat_template` is `NULL`, we warn the user to make sure template is applied to data if needed. It should be the user who should set `tokenizer.chat_template` while submitting the training run, we should not do it programmatically.

Kipok: people can always override settings to configure it as they want. The main question is what we want to have for a default behavior and the current default definitely looks incorrect to me. So I'd suggest we change default to apply chat template on messages as this is what most people would expect to happen, right?

wasiahmad: @Kipok need your suggestion. We load tokenizer as:
```
tokenizer = get_tokenizer(config["policy"]["tokenizer"])
```

So, we need to set `config["policy"]["tokenizer"]["chat_template"]="default"` if we see the data is in messages format. But we load the data later in `setup_data` method. So, how should we check if the data in messages format.

Given people may have data in different formats:
- input/output with template applied
- raw messages format
- messages format with template applied

I think it is best to just leave on the user to decide the template during submitting job as:
```
f'++policy.tokenizer.chat_template="default" '
```
?

Kipok: @coderabbitai, do you have suggestions for @wasiahmad on this? Assuming we still want to change the default to work as expected for both input and messages format (and ignore the case when messages are with template applied, assuming there people will provide appropriate parameters as this is not commonly used)

Kipok: that's not going to work. Let me push a change for this and @wasiahmad you can test on your side

PR #838, opened 2025-09-23
Sandbox history restoration fix

i-vainn:
Following the discussion in [PR #803](https://github.com/NVIDIA/NeMo-Skills/pull/803), this PR:
* Restores IPython state via two-step replay (history then new code) so output reflects only the new code.
* Adds test_state_restoration verifying no history output prints.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable session state restoration after resume or restart, reducing errors and ensuring correct execution.
  * Eliminates duplicated/concatenated outputs; only the latest execution’s output is shown.
  * Handles session resets gracefully with proper cleanup and consistent session tracking.

* **Refactor**
  * Adopted per-step state replay before running new code for improved stability and correctness.

* **Tests**
  * Added an end-to-end test validating state restoration and output correctness after a session reset.


tests/test_code_execution.py
```diff
@@ -376,6 +376,28 @@ def add (a b : Nat) : Nat :=
     )
 
 
+@pytest.mark.asyncio
+async def test_state_restoration():
```
gwarmstrong, line 381: please add some code block(s) that error to this test into the session before deletion, so the restoring after error behavior is explicit

nemo_skills/code_execution/sandbox.py
```diff
@@ -218,21 +218,41 @@ async def execute_code(
             output = {"process_status": "timeout", "stdout": "", "stderr": "Timed out\n"}
         new_session_created = output.pop("new_session_created", False)
 
-        # Rebuild state by executing concatenated history
+        # Rebuild state by re-executing history first, then execute the new code
```
gwarmstrong: I have some concerns about the guarantees/lack there-of of this replay-ability.

If I understand correctly, only blocks that don't error or timeout are added to the session histories, right?
- Adding to the history: https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/code_execution/sandbox.py#L239
- Process status handling: https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/code_execution/local_sandbox/local_sandbox_server.py#L384-L427

 If that's the case isn't it true that this cannot always be a perfect reconstruction, right?

Consider if we have code block A:
```
foo =  'bar'
```
Then block B:
```
foo = 'baz'
raise Exception()
```

block C:
```
print(foo)
```

block D:
```
while True: pass
```

What if instead of `raise Exception()`, the block was `while True: pass`

If I'm reading it right, on the first pass or exception case, C would print `baz`, but on the second pass it would print `bar`.

In fact, I don't think there's any way to guarantee the state is fully reproduced.

So It might be that this method is good enough (at the end of the day, how likely is it that you are going to have such a construction?), but if so, let's add a comment here describing this decision.

nemo_skills/code_execution/sandbox.py
```diff
@@ -218,21 +218,41 @@ async def execute_code(
...
+                )
+                try:
+                    restore_output = await self._send_request(restore_request, timeout)
+                except httpx.TimeoutException:
+                    restore_output = {"process_status": "timeout", "stdout": "", "stderr": "Timed out\n"}
+
+                # If restoration didn't complete successfully, return that result and skip executing new code
+                if restore_output.get("process_status") != "completed":
+                    output = restore_output
+                    restoration_failed = True
```
Kipok: what do you think about just failing with error here? I'm not sure there is much we can do to help if things start failing after retry and we should probably just fail the whole job as things might be very unexpected in this state?

nemo_skills/code_execution/sandbox.py
Kipok: yeah, we won't have perfect restoration here, but I think it's ok. @i-vainn please add a comment about this, but we can just ignore this thing, too hard to handle

tests/test_code_execution.py
i-vainn: Do you mean to add a check to ensure that error block is not replayed?

nemo_skills/code_execution/sandbox.py
i-vainn: I agree, the current approach should be good for the most cases. Added the comment explaining this behavior

tests/test_code_execution.py
gwarmstrong: More or less. Check that the error block is not replayed, but also (based on conversation below) that any session state changes (like changing a variable value) that are contained in the error block are also not replayed.

i-vainn: Added this part to the test, let me know if you think we need to consider some different case as well

gwarmstrong: This looks good to me now, thanks!

gwarmstrong (approved):

PR #839, opened 2025-09-23
IOI Improvements

SeanNaren:
This PR has a few things:

1. Improve IOI Evaluations speeds by flattening tests which ensures high sandbox utilization
2. Exposes CPU Processes/Affinity for the sandbox, setting to the correct number for evaluation standard
3. Expose CPU only parameters for evaluation on CPU only SLURM machines
4. Set default sandbox environment variables for IOI

## Summary by CodeRabbit

* New Features
  * Configure sandbox CPU affinity and process count via environment variables; automatically propagated from benchmarks.
  * IOI evaluation now supports multi-worker, batched execution with precompiled graders for faster runs; added config options: num_workers, test_batch_size, overwrite.
  * CLI: added cpus_per_task option for job submissions; can be combined with exclusive mode.

* Improvements
  * Enhanced evaluation readiness checks, clearer error diagnostics, progress/status tracking, and timing logs.

### DIFF
--- a/nemo_skills/pipeline/utils/eval.py
+++ b/nemo_skills/pipeline/utils/eval.py
@@ -15,17 +15,73 @@
 import importlib
 import logging
 import os
+import re
+import shlex
 from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
 
 import nemo_skills.pipeline.utils as pipeline_utils
 from nemo_skills.dataset.utils import get_dataset_module, import_from_path
+from nemo_skills.evaluation.evaluator import supports_single_eval
 from nemo_skills.inference import GENERATION_MODULE_MAP
 from nemo_skills.inference.generate import GenerationTask
 from nemo_skills.utils import compute_chunk_ids, get_logger_name
 
 LOG = logging.getLogger(get_logger_name(__file__))
+
+
+def parse_eval_args(eval_args: str) -> tuple[str | None, dict]:
+    # TODO we ideally don't want to rely on custom parsing of the command, but
+    # some major refactoring or clever ideas might be needed
+    """Parse eval_args string to extract eval_type and eval_config.
+
+    Handles Hydra argument formats:
+    - ++eval_type=value (override)
+    - +eval_type=value (new)
+    - eval_type=value (config)
+    """
+    if not eval_args:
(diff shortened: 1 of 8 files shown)


nemo_skills/pipeline/utils/exp.py
```diff
@@ -517,7 +517,11 @@ def add_task(
             if "PYTHONPATH" in override:
                 if override.startswith("PYTHONPATH="):
                     override = override[11:]
-                sandbox_env_updates["PYTHONPATH"] = override + ":/app"
+                    sandbox_env_updates["PYTHONPATH"] = override + ":/app"
+            elif override.startswith("NEMO_SKILLS_SANDBOX_CPU_AFFINITY="):
```
Kipok, line 521: why do we need this renaming? Why not just define the original env vars as UWSGI_CPU_AFFINITY and UWSGI_PROCESSES?

nemo_skills/pipeline/utils/eval.py
```diff
@@ -125,6 +125,14 @@ def get_benchmark_args_from_module(
         generation_args = f"++prompt_config={prompt_config} {generation_args}"
     requires_sandbox = get_arg_from_module_or_dict(benchmark_module, "REQUIRES_SANDBOX", False, override_dict)
 
+    # Add any benchmark-specific environment variables to cluster config
+    env_vars_from_module = getattr(benchmark_module, "ENV_VARS", getattr(benchmark_module, "env_vars", []))
+    if env_vars_from_module:
+        cluster_config.setdefault("env_vars", [])
```
Kipok, line 187: this is a little dangerous as cluster_config might be reused across different evals or other pipelines. Would it be possible to instead carry this through some other variable and then use `with temporary_env_update` context manager?

nemo_skills/evaluation/evaluator/ioi.py
```diff
@@ -312,3 +431,5 @@ def eval_ioi(cfg):
 
     pool.close()
     pool.join()
+
+    print(f"Total evaluation time: {(time.monotonic() - start_time) / 60:.2f} minutes")
```
Kipok, line 435: you can also report this in metrics by the way, similar to how generation time is being reported

PR #844, opened 2025-09-24
Add Responses API

gwarmstrong:
# Add Responses API Support for GPT-OSS Generation

This PR adds support for the OpenAI Responses API in nemo-skills, enabling generation with `gpt-oss`  models that use the responses format instead of the standard chat completions API.

### 1. ResponsesModel Implementation
- **New `ResponsesModel` class** in `nemo_skills/inference/model/responses.py`
- Implements the Responses API endpoint using OpenAI client directly
- Supports both sync and async generation with streaming capabilities
- Handles response parsing for tool calls, reasoning content, and text generation

### 2. Tool Calling Adapter System
- **New adapter classes** in `nemo_skills/mcp/adapters.py`:
  - `ResponsesSchemaAdapter`: Converts MCP tools to flat responses format
  - `ResponsesCallInterpreter`: Parses tool calls from responses output
  - `ResponsesResponseFormatter`: Formats tool results for responses API
  - `ResponsesConversationManager`: Manages conversation history with serialized output
- `ToolCallingWrapper` now detects the Adapter set to use in `nemo_skills/inference/model/tool_call.py`

## Usage Example

```python
from nemo_skills.pipeline.cli import generate, wrap_arguments

generate(
    ctx=wrap_arguments(
        "++prompt_config=your/prompt/config "
        "++inference.tokens_to_generate=8192 "
        "++tool_modules=[nemo_skills.mcp.servers.python_tool::PythonTool] "
        "++inference.temperature=0.6 "
    ),
    input_file='input.jsonl',
    cluster='local',
    model='openai/gpt-oss-20b',
    server_type='responses',  # Key: use responses server type
    server_gpus=2,
    output_dir='outputs/',
    num_random_seeds=1,
    with_sandbox=True,
)
```

## Example Output Structure

```json
{
  "system": "You are a helpful assistant.",
  "question": "Calculate 2+2 using Python",
  "num_generated_tokens": 156,
  "reasoning_content": ["I need to use Python to calculate 2+2."],
  "finish_reason": ["completed", "completed"],
  "num_tool_calls": 1,
  "conversation": [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "Calculate 2+2 using Python"},
    {
      "id": "rs_abc123",
      "type": "reasoning",
      "content": [{"text": "I need to use Python to calculate 2+2.", "type": "reasoning_text"}]
    },
    {
      "arguments": "{\"code\":\"2+2\"}",
      "call_id": "call_xyz789",
      "name": "stateful_python_code_exec",
      "type": "function_call"
    },
    {
      "type": "function_call_output",
      "call_id": "call_xyz789",
      "output": "4"
    },
    {
      "id": "msg_def456",
      "content": [{"text": "The result is 4.", "type": "output_text"}],
      "role": "assistant",
      "type": "message"
    }
  ],
  "generation": "The result is 4."
}
```

### DIFF
--- a/nemo_skills/inference/model/__init__.py
+++ b/nemo_skills/inference/model/__init__.py
@@ -28,6 +28,7 @@ from .gemini import GeminiModel
 from .megatron import MegatronModel
 from .openai import OpenAIModel
 from .parallel_thinking import ParallelThinkingConfig, ParallelThinkingTask
+from .responses import ResponsesModel
 
 # Tool Calling
 from .tool_call import ToolCallingWrapper
@@ -47,6 +48,7 @@ models = {
     "gemini": GeminiModel,
     "vllm": VLLMModel,
     "sglang": VLLMModel,
+    "responses": ResponsesModel,
 }
 
 
--- a/nemo_skills/inference/model/responses.py
+++ b/nemo_skills/inference/model/responses.py
@@ -1,0 +1,343 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+from openai import AsyncOpenAI, OpenAI
+
+from nemo_skills.utils import get_logger_name
+
+from .base import BaseModel
+from .context_retry import with_context_retry
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+class ResponsesModel(BaseModel):
+    """Model implementation using OpenAI Responses API via LiteLLM.
+
+    This model uses the responses API endpoint instead of chat completions,
+    which is useful for models like gpt-oss that support the responses format.
+    """
+
+    MODEL_PROVIDER = "openai"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        # Initialize OpenAI clients for responses API
+        self.client = OpenAI(base_url=self.base_url, api_key=self.litellm_kwargs["api_key"])
+        self.async_client = AsyncOpenAI(base_url=self.base_url, api_key=self.litellm_kwargs["api_key"])
+
+    def _build_completion_request_params(
+        self,
+        prompt: str,
+        tokens_to_generate: int = 512,
+        temperature: float = 0.0,
+        top_p: float = 0.95,
+        top_k: int = -1,
+        min_p: float = 0.0,
+        repetition_penalty: float = 1.0,
+        random_seed: int = None,
+        top_logprobs: int | None = None,
+        timeout: int | None = None,
+        stop_phrases: list[str] | None = None,
+        stream: bool = False,
+        reasoning_effort: str | None = None,
+        extra_body: dict = None,
+        tools: list[dict] | None = None,
+    ) -> dict:
+        """Build request parameters for responses API with string input."""
+        request = {
+            "input": prompt,
+            "max_output_tokens": tokens_to_generate,
+            "temperature": temperature,
+            "top_p": top_p,
+            "stream": stream,
+            "tools": tools,
+            "extra_body": {
+                "seed": random_seed,
+                "reasoning_effort": reasoning_effort,
+                "timeout": timeout,
+                "stop": stop_phrases,
+                "top_logprobs": top_logprobs,
+                "top_k": top_k,
+                "min_p": min_p,
+                "repetition_penalty": repetition_penalty,
+                **(extra_body or {}),
+            },
+        }
+
+        return request
+
+    def _build_chat_request_params(
+        self,
+        messages: list[dict],
+        stream: bool,
+        tokens_to_generate: int = 512,
+        temperature: float = 0.0,
+        top_p: float = 0.95,
+        top_k: int = -1,
+        min_p: float = 0.0,
+        repetition_penalty: float = 1.0,
+        random_seed: int = 0,
+        stop_phrases: list[str] | None = None,
+        timeout: int | None = None,
+        top_logprobs: int | None = None,
+        reasoning_effort: str | None = None,
+        tools: list[dict] | None = None,
+        extra_body: dict = None,
+    ) -> dict:
+        """Build request parameters for responses API with messages input."""
+        request = {
+            "input": messages,
+            "max_output_tokens": tokens_to_generate,
+            "temperature": temperature,
+            "top_p": top_p,
+            "stream": stream,
+            "tools": tools,
+            "extra_body": {
+                "seed": random_seed,
+                "reasoning_effort": reasoning_effort,
+                "timeout": timeout,
+                "stop": stop_phrases,
+                "top_logprobs": top_logprobs,
+                "top_k": top_k,
+                "min_p": min_p,
+                "repetition_penalty": repetition_penalty,
+                **(extra_body or {}),
+            },
+        }
+
+        return request
+
+    def _parse_responses_response(self, response, include_response: bool = False, **kwargs) -> dict:
+        """Parse responses API response into standard format."""
+
+        result = {"generation": "", "num_generated_tokens": 0}
+
+        # Get token usage
+        if hasattr(response, "usage"):
+            result["num_generated_tokens"] = getattr(response.usage, "output_tokens", 0)
+
+        # Check for tool calls in the output array
+        tool_calls = []
+        reasoning_content = ""
+        generation_text = ""
+
+        if hasattr(response, "output") and response.output:
+            for output_item in response.output:
+                # Handle reasoning content
+                if hasattr(output_item, "type") and output_item.type == "reasoning":
+                    if hasattr(output_item, "content") and output_item.content:
+                        for content_item in output_item.content:
+                            if hasattr(content_item, "text"):
+                                reasoning_content += content_item.text + "\n"
+
+                # Handle function calls
+                elif hasattr(output_item, "type") and output_item.type == "function_call":
+                    tool_calls.append(output_item)
+
+                # Handle message content
+                elif hasattr(output_item, "type") and output_item.type == "message":
+                    if hasattr(output_item, "content") and output_item.content:
+                        for content_item in output_item.content:
+                            if hasattr(content_item, "text"):
+                                generation_text += content_item.text
+
+        # Set the appropriate response fields
+        if tool_calls:
+            result["tool_calls"] = tool_calls
+            result["generation"] = ""  # No text generation when there are tool calls
+        else:
+            result["generation"] = generation_text
+
+        # Add reasoning content if available
+        if reasoning_content:
+            result["reasoning_content"] = reasoning_content.strip()
+
+        # Add finish reason if available
+        if hasattr(response, "status"):
+            result["finish_reason"] = response.status
+
+        if include_response:
+            result["response"] = response
+
+        return result
+
+    @with_context_retry
+    async def generate_async(
+        self,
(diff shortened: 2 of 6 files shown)


nemo_skills/inference/model/tool_call.py
```diff
@@ -59,10 +64,28 @@ def __init__(
...
-        self.call_interpreter = ChatCompletionCallInterpreter()
-        self.response_formatter = ChatCompletionResponseFormatter()
+
+        # Detect model type and set up appropriate adapters
+        self._setup_adapters()
+
+    def _setup_adapters(self):
+        """Set up adapters based on model type."""
+        # Import here to avoid circular imports
+        from .responses import ResponsesModel
```
Kipok: what exactly is being circularly imported?

nemo_skills/mcp/adapters.py
```diff
@@ -116,3 +135,116 @@ def format(self, tool_call, result):
...
+class ResponsesConversationManager(ConversationManager):
+    """Manages conversation history for responses API models."""
+
+    def add_assistant_response(self, conversation: List[Dict[str, Any]], response: Dict[str, Any]) -> None:
+        """Add an assistant response to the conversation using serialized output."""
+        # Use the serialized output from the responses API
+        if "serialized_output" in response:
+            conversation.extend(response["serialized_output"])
+        else:
+            # Fallback to basic message format
```
Kipok: is this fallback how the api is supposed to work? Or does it mean something unexpected is happening and we should maybe throw an error?

nemo_skills/pipeline/utils/server.py
```diff
@@ -123,7 +123,7 @@ def get_server_command(
 
     # check if the model path is mounted if not vllm, sglang, or trtllm;
     # vllm, sglang, and trtllm can also pass model name as "model_path" so we need special processing
-    if server_type not in ["vllm", "sglang", "trtllm"]:
+    if server_type not in ["vllm", "sglang", "trtllm", "responses"]:
```
Kipok: I think this might need some refactoring to make sure responses endpoint is defined for all server types. The server type is used for both selecting the client model, but also to figure out the right commands for launching the server. Is there a way to refactor it such that responses api is supported for any server type instead?

nemo_skills/inference/model/tool_call.py
gwarmstrong: I thought there would be a circular import due to nemo_skills.inference.model importing both tool_call and responses, but checking again it doesn't seem to happen either on relative or absolute imports, so this can be moved to the top

nemo_skills/mcp/adapters.py
gwarmstrong: yeah probably better to throw an error instead

nemo_skills/pipeline/utils/server.py
gwarmstrong: I can see if I can factor out a `client_type` or something to increase the compatibility across different kinds of servers

nemo_skills/inference/model/__init__.py
```diff
@@ -47,6 +48,7 @@
     "gemini": GeminiModel,
     "vllm": VLLMModel,
     "sglang": VLLMModel,
+    "responses": ResponsesModel,
```
activatedgeek: Rename to `openai_responses` for semantic clarity.

nemo_skills/inference/model/responses.py
```diff
@@ -0,0 +1,375 @@
...
+        stream: bool = False,
+        reasoning_effort: str | None = None,
+        tools: list[dict] | None = None,
+        include_response: bool = False,
+        extra_body: dict = None,
+    ) -> dict:
+        """Generate response using responses API."""
+
+        # Check tool calls are a list of dict
+        if tools is not None:
```
activatedgeek: This check should be moved to `_build_chat_request_params` to avoid repetition.

nemo_skills/inference/model/responses.py
```diff
@@ -0,0 +1,375 @@
...
+    MODEL_PROVIDER = "openai"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        # Initialize OpenAI clients for responses API
+        self.client = OpenAI(base_url=self.base_url, api_key=self.litellm_kwargs["api_key"])
+        self.async_client = AsyncOpenAI(base_url=self.base_url, api_key=self.litellm_kwargs["api_key"])
+
+    def _build_completion_request_params(
```
activatedgeek: Can we combine the logic here with `_build_chat_completion_request`, since it is largely repeated, except for the `stream` argument?

nemo_skills/inference/model/responses.py
```diff
@@ -0,0 +1,375 @@
...
+        stream: bool = False,
+        reasoning_effort: str | None = None,
+        tools: list[dict] | None = None,
+        include_response: bool = False,
+        extra_body: dict = None,
+    ) -> dict:
+        """Synchronous version of generate using responses API."""
+
+        # Check tool calls are a list of dict
+        if tools is not None:
```
activatedgeek: This check should be moved to `_build_chat_request_params` to avoid repetition.

nemo_skills/inference/model/tool_call.py
activatedgeek: Generally, if an imported package is only referenced inside a function call, then circular imports won't be a problem. It is only a problem when the top level parsing of the module and its functions/classes require the execution at the first import. In this case, the `_setup_adapters` will only be called at class init, which likely won't happen at import time.

Also in the case where there is a risk of circular imports, it is generally advisable to always use absolute imports instead of relative imports.

nemo_skills/inference/model/responses.py
```diff
@@ -0,0 +1,375 @@
...
+
+from nemo_skills.utils import get_logger_name
+
+from .base import BaseModel
+from .context_retry import with_context_retry
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+class ResponsesModel(BaseModel):
```
activatedgeek: Rename to `OpenAIResponsesModel` for semantic clarity, indicating specific use of OpenAI api.

nemo_skills/inference/model/responses.py
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        request_params = self._build_request_params(prompt=prompt, stream=stream, **kwargs)
+
+        # Use OpenAI client for responses API
+        LOG.info(f"Making sync responses API call with params: {request_params}")
+        LOG.info(f"Model name: {self.litellm_kwargs['model']}")
+
+        # Use the original model name (without litellm prefix) for OpenAI client
+        model_name = self.model_name_or_path  # Just gpt-oss-20b
```
activatedgeek: Should we add a check for the compatible models in the constructor itself?

nemo_skills/pipeline/utils/server.py
activatedgeek: I think at this stage both [vLLM](https://github.com/vllm-project/vllm/pull/20504) and [SGLang](https://github.com/sgl-project/sglang/issues/4793) don't support responses API fully. And it is only the OpenAI api that does support it?

nemo_skills/pipeline/utils/server.py
```diff
@@ -152,7 +152,7 @@ def get_server_command(
             f"    --micro-batch-size 1 "  # that's a training argument, ignored here, but required to specify..
             f"    {server_args} "
         )
-    elif server_type == "vllm":
+    elif server_type in ["vllm", "responses"]:
```
activatedgeek: To understand this more, is the idea here that the OpenAI SDK internally does all the parsing of tool calls etc. required to work with the responses API, despite us launching a vLLM server? As in, internally, does the OpenAI client object just call raw text completion, and then does additional parsing on top to make it work with the responses API? 

If that's the case, then shouldn't we also have the sglang server supported as well?

nemo_skills/pipeline/utils/server.py
gwarmstrong: Not sure about SGLang. The thing with vLLM is that it supports responses creation, but it can't support updating a response after it has been created. So, this is fine for open source models where we effectively get the whole chat trace back with the response--so we can continue the conversation by uploading the whole chat history (similar to how we would for completions).

However, I should be more clear that this response handler won't actually work well for closed source models like openai api where they may hide parts of the reasoning trace. This also won't work with server side tool execution where the tool call tokens are hidden.

So this actually works totally fine locally with vllm, but we would need different changes to suport openai apis.

smahdavi4:
Thanks for adding the responses api! I just want to share two small suggestions, in case they might be helpful:
* Generalizing text completion, chat completion and responses into a “completion_type” and calling litellm text completion, litellm chat completion, or litellm responses api depending on the state. This allows for all types of completions with a single model instance and through a single argument in generate_async (so we don’t have to pass responses type to the model initializer). This will also allow for reusing the http clients and not having to reinitialize an openai client.
* deprecating “use_completions_api” in favour of a tri-state “completion_type”

nemo_skills/inference/model/base.py
```diff
@@ -29,18 +30,332 @@
...
+            extra_body["repetition_penalty"] = params["repetition_penalty"]
+
+        # Add any additional extra_body parameters
+        if params["extra_body"]:
+            extra_body.update(params["extra_body"])
+
+        if extra_body:
+            request["extra_body"] = extra_body
+
+        return request
```
Kipok, line 308: double return?

Kipok (commented):
@smahdavi4 could you also review this please? I looked through the changes on the high level, but didn't have a chance to dive into details

@smahdavi4 for your current comment I think @gwarmstrong found that responses api in litellm is in a fairly bad state and it's really hard to make it work, so going through a separate client is basically necessary until they fix it. About completion_type, I think maybe we can do it in another PR, that would be a breaking change and it's not fully clear how responses api will be eventually supported in litellm or other providers in a stable way, so we can also wait before doing a "first-class" support for it. But if we find the current structure inconvenient, we can definitely make this change

@gwarmstrong Since the changes are fairly big and touch many parts, could you please run slurm tests from this branch? Let me know if you need details on how to do this. I will also share another test offline that I will ask you to run for a non-public model that goes through tool-calling api to verify that nothing is broken there

nemo_skills/inference/model/base.py
```diff
@@ -182,6 +519,119 @@ def _initialize_tokenizer(self, tokenizer: str | None) -> WrapperAutoTokenizer |
...
+    # Public methods for client handlers to call
+    def get_supported_params(self) -> set:
+        """Public method for models to restrict parameters - override in subclasses"""
+        return set()  # Base implementation allows all
+
+    def apply_model_specific_params(self, request: dict, params: dict) -> dict:
+        """Public method for model-specific parameter handling - override in subclasses"""
+        return request
+
+    def parse_chat_completion_response(self, response, **kwargs) -> dict:
```
smahdavi4, line 531: Is this function inherited anywhere? If not, we can remove this function and directly call `_parse_chat_completion_response`

nemo_skills/inference/model/__init__.py
```diff
@@ -94,14 +112,26 @@ def get_parallel_thinking_model(
...
     tokenizer=None,
     additional_config=None,
     tool_modules: list[str] | None = None,
     tool_overrides: dict | None = None,
     **kwargs,
 ):
+    """A helper function to create a tool calling model.
+
+    Args:
+        model: Model name/path or model instance
```
smahdavi4, line 125: Some of the function docstrings and some code comments are not informative enough (probably LLM artifacts). Maybe we could remove them to avoid redundancy?

nemo_skills/inference/model/base.py
```diff
@@ -182,6 +519,119 @@ def _initialize_tokenizer(self, tokenizer: str | None) -> WrapperAutoTokenizer |
...
+        generation_text = ""
+
+        if hasattr(response, "output") and response.output:
+            for output_item in response.output:
+                # Handle reasoning content
+                if hasattr(output_item, "type") and output_item.type == "reasoning":
+                    if hasattr(output_item, "content") and output_item.content:
+                        for content_item in output_item.content:
+                            if hasattr(content_item, "text"):
+                                reasoning_content += content_item.text + "\n"
```
smahdavi4, line 565: Shall we somehow keep the order of events? If we want to consider a scenario where we want to distill the outputs of a model into a another model, we need a list of interleaved text and tools.

PR #848, opened 2025-09-25
Evaluation on OJBench

wasiahmad:
In this PR, we are adding evaluation support for [OJBench](https://github.com/He-Ren/OJBench) through nemo-skills. Primary changes are:

- Dataset download through nemo-skills/dataset/ojbench
- Evaluation logic implemented at nemo_skills/evaluation/evaluator/code.py


wasiahmad:
@Kipok the PR supports benchmarking LLMs on OJBench and I checked that we can reproduce scores reported in the paper. However, I believe there is a bug that causes producing follow metrics even if I executed 1 run only.

```
{
  "ojbench": {
    "pass@1": {
      "num_entries": 928,
      "avg_tokens": 19480,
      "gen_seconds": 2496,
      "accuracy": 27.370689655172413
    },
    "pass@2": {
      "num_entries": 928,
      "avg_tokens": 19480,
      "gen_seconds": 2496,
      "accuracy": 27.370689655172413
    },
    "pass@1[avg-of-2]": {
      "num_entries": 928,
      "avg_tokens": 19480,
      "gen_seconds": 2496,
      "accuracy": 27.370689655172413,
      "reasoning_tokens_statistics": {
        "avg": 0.0,
        "std_dev_across_runs": 0.0,
        "avg_sample_std_dev": 0.0,
        "std_err_across_runs": 0.0
      },
      "answer_tokens_statistics": {
        "avg": 19480.620689655174,
        "std_dev_across_runs": 0.0,
        "avg_sample_std_dev": 0.0,
        "std_err_across_runs": 0.0
      },
      "accuracy_statistics": {
        "avg": 0.27370689655172414,
        "std_dev_across_runs": 0.0,
        "avg_sample_std_dev": 0.0,
        "std_err_across_runs": 0.0
      }
    }
  },
...
```

So, I am not sure how pass@2 is even calculated.

wasiahmad:
All issues are fixed and as per my testing, this PR is safe to merge. @Kipok

nemo_skills/dataset/ojbench/prepare.py
```diff
@@ -0,0 +1,83 @@
...
+
+import json
+import os
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+REPO_URL = "https://huggingface.co/datasets/He-Ren/OJBench_testdata"
+HF_TOKEN = os.environ.get("HF_TOKEN")
```
Kipok, line 23: please raise some meaningful error if this is not defined if HF_TOKEN is required

nemo_skills/dataset/ojbench/prepare.py
```diff
@@ -0,0 +1,83 @@
...
+
+    try:
+        if destination.exists() or destination.is_symlink():
+            print(f"Destination '{destination}' already exists. Removing it...")
+            if destination.is_dir():
+                shutil.rmtree(destination)
+            else:
+                destination.unlink()
+
+        auth_url = url.replace("https://huggingface.co/", f"https://user:{HF_TOKEN}@huggingface.co/", 1)
```
Kipok, line 44: why do we need to clone it? Can't we use datasets api instead and save the data through that? The rabbit's comment about HF_TOKEN leakage is actually an important thing to fix and should be done automatically if we don't add it to repo url

nemo_skills/dataset/ojbench/prepare.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: does this require data_dir usage? If so, please mention this in the docs and give example command

nemo_skills/evaluation/evaluator/ojbench.py
```diff
@@ -0,0 +1,135 @@
...
+        clone_cmd = "git clone https://github.com/He-Ren/OJBench.git"
+        result, _ = await sandbox.execute_code(clone_cmd, language="shell", timeout=300)
+        if result["process_status"] != "completed":
+            LOG.warning(f"Failed to clone OJBench repo: {result.get('stderr', 'Unknown error')}")
+            return False
+
+        install_cmd = "pip install -e OJBench"
+        result, _ = await sandbox.execute_code(install_cmd, language="shell", timeout=300)
+        if result["process_status"] != "completed":
+            LOG.warning(f"Failed to install ojbench: {result.get('stderr', 'Unknown error')}")
```
Kipok: can we raise this as an exception? What's the point to do a warning and gracefully exit if this means the whole job really failed

nemo_skills/evaluation/evaluator/ojbench.py
```diff
@@ -0,0 +1,135 @@
...
+                for line in f_in:
+                    sample = json.loads(line)
+                    sample = preprocess_code(sample, sample["language"], strip_whitespace=True)
+                    sample["prompt"] = sample.pop("question")
+                    sample["content"] = f"```{sample['language']}\n{sample['completion']}\n```"
+                    sample.pop("completion")
+                    samples.append(sample)
+
+            with open(jsonl_file, "w", encoding="utf-8") as f_out:
+                f_out.writelines(json.dumps(sample) + "\n" for sample in samples)
```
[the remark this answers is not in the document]
Kipok: @wasiahmad please address this, it's very important to only replace files at the very end to avoid costly mistakes of losing data

nemo_skills/evaluation/evaluator/ojbench.py
```diff
@@ -0,0 +1,135 @@
...
+            cmd = f'env -i PATH="/usr/local/bin:/usr/bin:/bin" python3 -c {shlex.quote(eval_code)}'
+            output, _ = await sandbox.execute_code(
+                cmd,
+                language="shell",
+                timeout=eval_config.timeout * len(samples) + 60,
+                max_output_characters=100_000,
+            )
+
+            if output.get("process_status") != "completed":
+                LOG.error(f"Evaluation failed for {jsonl_file}. Stderr: {output.get('stderr')}")
```
Kipok: what's the point of skipping this? Let's just fail with error, otherwise it might lead to hard-to-notice issues

nemo_skills/evaluation/evaluator/ojbench.py
```diff
@@ -0,0 +1,135 @@
...
+                continue
+
+            with open(eval_results_path, "rt", encoding="utf-8") as fin:
+                results = [json.loads(line) for line in fin]
+
+            for sample, result in zip(samples, results):
+                sample["verdict"] = result["verdict"]
+                sample["is_passed"] = result["is_passed"]
+
```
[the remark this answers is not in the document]
Kipok: @wasiahmad probably a good idea to do this

nemo_skills/evaluation/evaluator/ojbench.py
```diff
@@ -0,0 +1,135 @@
...
+from nemo_skills.evaluation.evaluator.code import preprocess_code
+from nemo_skills.utils import get_logger_name, nested_dataclass, unroll_files
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+@nested_dataclass(kw_only=True)
+class OJBenchConfig:
+    sandbox: dict = field(default_factory=lambda: {"sandbox_type": "local"})
+    keep_mounts_for_sandbox: bool = True
```
Kipok: is this used?

nemo_skills/pipeline/utils/eval.py
```diff
@@ -365,6 +371,9 @@ def prepare_eval_commands(
             if benchmark_args.requires_sandbox and not with_sandbox:
                 LOG.warning("Found benchmark (%s) which requires sandbox, enabled sandbox for it.", benchmark)
 
+            if benchmark_args.requires_sandbox and keep_mounts_for_sandbox:
```
Kipok: I don't think we need a special message here, but instead can you message that mounts are kept if benchmark_args.keep_mounts_for_sandbox is True

nemo_skills/pipeline/utils/exp.py
```diff
@@ -343,6 +343,8 @@ def add_task(
     partition=None,
     time_min=None,
     with_sandbox=False,
+    keep_mounts_for_sandbox=False,
+    sandbox_mount_paths: list[str] | None = None,
```
Kipok: we don't really need this argument yet, as it's never set to anything but original mounts. So for simplicity we can just assume we are always using original mounts here and we can refine it later when we have a way to pass different list

nemo_skills/pipeline/eval.py
```diff
@@ -171,6 +171,9 @@ def eval(
         False, help="If True, will re-run jobs even if a corresponding '.done' file already exists"
     ),
     with_sandbox: bool = typer.Option(False, help="If True, will start a sandbox container alongside this job"),
+    keep_mounts_for_sandbox: bool = typer.Option(
```
Kipok, line 174: Please add it to all pipelines that have with_sandbox argument defined. We want to make sure arguments in all pipelines are consistent

nemo_skills/pipeline/eval.py
```diff
@@ -171,6 +171,9 @@ def eval(
...
+        False, help="If True, will keep the mounts for the sandbox container"
```
Kipok: Please add a message that this is risky given that sandbox executes LLM commands and could potentially lead to data loss. So we advise not to use this unless really necessary

Kipok (changes requested):

nemo_skills/evaluation/evaluator/ojbench.py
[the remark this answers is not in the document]
wasiahmad: Done.

nemo_skills/dataset/ojbench/prepare.py
wasiahmad: @Kipok I think we can't use the datasets api because the repo is not configured to be used in that way.

nemo_skills/evaluation/evaluator/ojbench.py
wasiahmad: Followed this https://github.com/NVIDIA/NeMo-Skills/blob/74b8ed88052603488d6fe040f20301a828b05202/nemo_skills/evaluation/evaluator/scicode.py#L115 @Kipok But I think it is a good idea because without installing it, evaluation won't work anyway.

nemo_skills/evaluation/evaluator/ojbench.py
wasiahmad: No. Removed.

nemo_skills/pipeline/utils/exp.py
wasiahmad: @Kipok How do we pass `mounts` to `add_task`. We need this in the following code-block. 

```
if with_sandbox:
        sandbox_env_updates = {
            "LISTEN_PORT": sandbox_port,
            "NGINX_PORT": sandbox_port,
        }
        current_env_vars = cluster_config.get("env_vars", []).copy()
        for override in current_env_vars:
            if "PYTHONPATH" in override:
                if override.startswith("PYTHONPATH="):
                    override = override[11:]
                sandbox_env_updates["PYTHONPATH"] = override + ":/app"

        with temporary_env_update(cluster_config, sandbox_env_updates):
            commands.append(get_sandbox_command(cluster_config))
            sandbox_executor = get_executor(
                cluster_config=cluster_config,
                container=cluster_config["containers"]["sandbox"],
                num_nodes=executors[0].nodes if cluster_config["executor"] == "slurm" else 1,
                tasks_per_node=1,
                gpus_per_node=0,
                partition=partition,
                time_min=time_min,
                mounts=sandbox_mount_paths if keep_mounts_for_sandbox else [],
                dependencies=dependencies,
                job_name=task_name,
                log_dir=log_dir,
                log_prefix="sandbox",
                extra_package_dirs=extra_package_dirs,
                slurm_kwargs=slurm_kwargs,
                heterogeneous=heterogeneous,
                het_group=het_group,
                total_het_groups=total_het_groups,
                overlap=True,
                with_ray=with_ray,
            )
            executors.append(sandbox_executor)
            het_group_indices.append(het_group)
        het_group += 1
        LOG.info("Sandbox command: %s", commands[-1])
```

dockerfiles/Dockerfile.sandbox
```diff
@@ -15,14 +15,19 @@
...
+# Install dependencies required for Lean 4, pypy3, and other tools
 RUN apt-get update && \
-    apt-get install -y curl git net-tools bzip2 && \
+    apt-get install -y curl git net-tools bzip2 build-essential libseccomp-dev && \
     curl -L https://downloads.python.org/pypy/pypy3.10-v7.3.17-linux64.tar.bz2 -o /tmp/pypy.tar.bz2 && \
     tar -xjf /tmp/pypy.tar.bz2 -C /opt/ && \
-    ln -s /opt/pypy3.10-v7.3.17-linux64/bin/pypy3 /usr/local/bin/pypy3 && \
+    ln -s /opt/pypy3.10-v7.3.17-linux64/bin/pypy3 /usr/bin/pypy3 && \
+    /usr/bin/pypy3 -m ensurepip && \
     rm /tmp/pypy.tar.bz2
```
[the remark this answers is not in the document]
wasiahmad: Skipping.

nemo_skills/evaluation/evaluator/ojbench.py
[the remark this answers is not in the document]
wasiahmad: @Kipok I am skipping this as this logic is used for all benchmarks eval. This is not going to cause costly loss of data, so this can be ignored. Let me know if you think differently.

nemo_skills/pipeline/utils/exp.py
Kipok: just don't pass it, the default should be what you want

docs/evaluation/code.md
```diff
@@ -201,4 +201,85 @@ all you need to do is replace `openhands` with `swe_agent` in the command above.
...
+
+- Benchmark is defined in [`nemo_skills/dataset/ojbench/__init__.py`](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/dataset/ojbench/__init__.py)
+- Original benchmark source is [here](https://github.com/He-Ren/OJBench/tree/main).
+
+#### Data preparation
+
+Before running ns eval, you will need to prepare the data with this command:
+
+```
+ns prepare_data --data_dir=<DATA_DIR> --cluster=<CLUSTER_NAME> ojbench
```
Kipok, line 216: Please update pipeline/prepare_data.py to require data_dir for this benchmark to avoid accidental mistakes https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/prepare_data.py#L30

docs/evaluation/code.md
```diff
@@ -201,4 +201,85 @@ all you need to do is replace `openhands` with `swe_agent` in the command above.
...
+Here's how to run a sample evaluation of [Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B) on a Slurm cluster.
+
+1. Prepare the data following instructions in the previous section.
+2. Run
+```
+ns eval \
+    --cluster=<CLUSTER_NAME> \
+    --model=Qwen/Qwen3-32B \
+    --server_type=vllm \
+    --server_args="--async-scheduling" \
```
Kipok:
```suggestion
```

docs/evaluation/code.md
```diff
@@ -201,4 +201,85 @@ all you need to do is replace `openhands` with `swe_agent` in the command above.
...
+    --split=test_python \
+    --data_dir=<DATA_DIR> \
+    --output_dir=<OUTPUT_DIR> \
+    ++inference.temperature=0.6 \
+    ++inference.top_p=0.95 \
+    ++inference.tokens_to_generate=32768
+```
+replacing <...> with your desired parameters.
+
+After all jobs are complete, you can check the results in `<OUTPUT_DIR>/eval-results/ojbench/metrics.json`. They should look something like this:
```
Kipok: could you instead add a printout of summarize-results? I think it's a little cleaner to have that in the docs

docs/evaluation/code.md
```diff
@@ -201,4 +201,85 @@ all you need to do is replace `openhands` with `swe_agent` in the command above.
...
+
+We encourage to download OJBench data into a Slurm cluster location because 15GB data will be downloaded by cloning [huggingface.co/datasets/He-Ren/OJBench_testdata](https://huggingface.co/datasets/He-Ren/OJBench_testdata). Two files will be created at `<DATA_DIR>` named `test_python.jsonl` and `test_cpp.jsonl`.
+
+Note that, data downloading require `HF_TOKEN` to be in the environment variables.
+
+#### Sample run
+
+Here's how to run a sample evaluation of [Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B) on a Slurm cluster.
+
+1. Prepare the data following instructions in the previous section.
```
Kipok, line 225: please double check that this is rendered properly by running `mkdocs serve` and checking this page in local copy of the documentation. It might be that formatting is broken here

nemo_skills/evaluation/evaluator/ojbench.py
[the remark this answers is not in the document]
Kipok: which other benchmark uses this? If that's the case we should update all other cases. The example of proper logic is here https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/evaluation/evaluator/scicode.py#L135, note that we only override files after all evals are done. If you open files for writing before, any accidental mistake in the eval code will cause data loss. So it's a simple change, just do all evals first and then override the files

docs/evaluation/code.md
wasiahmad: Done.

nemo_skills/pipeline/utils/exp.py
wasiahmad: Done.

Kipok (approved):

PR #850, opened 2025-09-25
Only run env var check for identity file when key available

activatedgeek:
The example cluster configurations suggest that `ssh_tunnel.identity` can be skipped, but running it leads to the following error:

```shell
  File "/code/NeMo-Skills/nemo_skills/pipeline/utils/cluster.py", line 267, in update_ssh_tunnel_config
    if "$" in cluster_config["ssh_tunnel"]["identity"]:
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: 'identity'
```

I change it to use `.get` for the dictionary with empty default string. This approach also allows users to directly use the identity available in the system's SSH agent (as supported by the underlying Fabric connector in Nemo Run).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when SSH tunnel configuration omits an identity value, improving stability for environments that do not supply this field.
  * Preserved validation for unresolved environment-variable references in the identity setting, so errors are still raised when variable placeholders remain unresolved.


shtoshni (approved):
Nice trick!

PR #851, opened 2025-09-26
handle the absence of nemo_skills in the path

dgtm777:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging name generation so logger identifiers are reliable when files or packages are located outside the expected project layout. This prevents malformed or confusing logger names in CI, bundled apps, or atypical installs, producing cleaner, more consistent log output and reducing logging-related issues in edge-case environments.

### DIFF
--- a/nemo_skills/utils.py
+++ b/nemo_skills/utils.py
@@ -128,8 +128,8 @@ def remove_handlers():
 
 
 def get_logger_name(file):
-    path = file.split("nemo_skills")[1] if "nemo_skills" in file else file
-    return "nemo_skills" + path.replace("/", ".").replace(".py", "")
+    path = "nemo_skills" + file.split("nemo_skills")[1] if "nemo_skills" in file else file
+    return path.replace("/", ".").replace(".py", "")
 
 
 def get_skills_root_dir():
--- a/nemo_skills/utils.py
+++ b/nemo_skills/utils.py
@@ -128,10 +128,10 @@ def remove_handlers():
 
 
 def get_logger_name(file):
-    if "nemo_skills" in file:
+    if "/nemo_skills/" in file:
         return "nemo_skills" + file.split("nemo_skills")[1].replace("/", ".").replace(".py", "")
     else:
-        return Path(file).stem
+        return f"[external] {Path(file).stem}"
 
 
 def get_skills_root_dir():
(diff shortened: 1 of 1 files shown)


Kipok:
I think the main use-case that's not covered by original implementation is logging inside file in recipes folder. @activatedgeek maybe we should strip on the top-level NeMo-Skills folder for those? Then logs will start from `recipes...` which feels clear enough?

activatedgeek:
Sure thing. I think we can just have simpler logic where for files outside the `nemo_skills` module just gets the filename with `Path(file).stem`.

nemo_skills/utils.py
```diff
@@ -128,7 +128,10 @@ def remove_handlers():
 
 
 def get_logger_name(file):
-    return "nemo_skills" + file.split("nemo_skills")[1].replace("/", ".").replace(".py", "")
+    if "nemo_skills" in file:
+        return "nemo_skills" + file.split("nemo_skills")[1].replace("/", ".").replace(".py", "")
+    else:
+        return Path(file).stem
```
activatedgeek: Should we consider adding a `nemo_skills.third_party.` prefix for cases where the file is outside the `nemo_skills` package?

Also, you might want to check for `/nemo_skills/` instead of `nemo_skills` for a hacky way to ensure it is a folder name and not just any substring.

activatedgeek (changes requested):

nemo_skills/utils.py
activatedgeek: Also @dgtm777, you'll want to complete the commit sign-off.

Kipok: good idea about `/nemo_skills/`. For third_party, not sure, it looks as if that's just a module name and could be confusing. Either we should make it more explicit or just using a file name seems ok for now and we can refine it later

activatedgeek: Yes, I think any reasonable indicator of an external file is fine. We could even have like a `[external]` prefix, since the name doesn't have to be in the module format.

Kipok:
@dgtm777 do you want to address this last small issue, so that we can merge this?

dgtm777:
@Kipok, @activatedgeek, thanks for the review! Here is the updated version. Let me know if anything else is missing.

Kipok (approved):

PR #853, opened 2025-09-27
Slurm: fix time format, and allow default timeout 

artbataev:
- Allow "default" timeout to be specified for partitions to avoid listing all available partitions in "timeouts" section
- Fix time format to `D-HH:MM:SS` according to the docs https://slurm.schedmd.com/sbatch.html#OPT_time (previously failed on some clusters)

### DIFF
--- a/cluster_configs/example-slurm.yaml
+++ b/cluster_configs/example-slurm.yaml
@@ -63,8 +63,9 @@ job_name_prefix: "nemo_skills:"
 # this will be used as a slurm parameter + to signal SFT job to finish
 # before the timeout to have time to save the last checkpoint
 # timeouts:
-#   partition_name1: 06:00:00
-#   partition_name2: 01:30:00
+#   default: "06:00:00"
+#   partition_name1: "06:00:00"
+#   partition_name2: "01:30:00"
 
 # define any environment variables. Note that HF_HOME is required by default and needs to be a mounted path!
 # env_vars:
--- a/nemo_skills/pipeline/utils/cluster.py
+++ b/nemo_skills/pipeline/utils/cluster.py
@@ -43,16 +43,15 @@ def get_timeout(cluster_config, partition):
     # Time format for SLURM: "minutes", "minutes:seconds", "hours:minutes:seconds",
     # "days-hours", "days-hours:minutes" and "days-hours:minutes:seconds"
     # https://slurm.schedmd.com/sbatch.html#OPT_time
-    default_timeout = cluster_config.get("timeouts", {}).get("default", "10000-00:00:00")
+    # timeout value passed to SBATCH
+    default_timeout = cluster_config.get("timeouts", {}).get("default", "100-00:00:00")
     try:
         timeout = cluster_config["timeouts"][partition or cluster_config["partition"]]
 
         # subtracting 15 minutes to account for the time it takes to save the model
-        # the format expected by nemo is days:hours:minutes:seconds
-        time_diff = datetime.strptime(timeout, "%H:%M:%S") - datetime.strptime("00:15:00", "%H:%M:%S")
-        timeout = (
-            f"0-{time_diff.seconds // 3600:02d}:{(time_diff.seconds % 3600) // 60:02d}:{time_diff.seconds % 60:02d}"
-        )
+        # the format expected by nemo is days-hours:minutes:seconds
+        time_diff = datetime.strptime(timeout, "%j-%H:%M:%S") - datetime.strptime("00:15:00", "%H:%M:%S")
+        timeout = f"{time_diff.days}-{time_diff.seconds // 3600:02d}:{(time_diff.seconds % 3600) // 60:02d}:{time_diff.seconds % 60:02d}"
     except KeyError:
         timeout = default_timeout
     return timeout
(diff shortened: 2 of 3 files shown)


cluster_configs/example-slurm.yaml
```diff
@@ -63,8 +63,9 @@ job_name_prefix: "nemo_skills:"
 # this will be used as a slurm parameter + to signal SFT job to finish
 # before the timeout to have time to save the last checkpoint
 # timeouts:
-#   partition_name1: 06:00:00
-#   partition_name2: 01:30:00
+#   default: "06:00:00"
```
Kipok: should we maybe move it to top level as `default_timeout`? To avoid an edge case where there might be a partition called "default"?

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -39,19 +39,52 @@
...
+
+def get_timeout(cluster_config, partition) -> str:
+    default_timeout = cluster_config.get("timeouts", {}).get("default", "100-00:00:00")
+    try:
+        timeout_str = cluster_config["timeouts"][partition or cluster_config["partition"]]
+    except KeyError:
+        timeout_str = default_timeout
+    timeout = _parse_slurm_timeout(timeout_str)
+    # subtracting 15 minutes to account for the time it takes to save the model
+    # the format expected by nemo is days-hours:minutes:seconds
```
Kipok, line 84: did you test that this works? We should ideally test both general slurm job as well as nemo-aligner / nemo-rl jobs. This timeout function is also used in training scripts (https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/train.py#L143, https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/pipeline/nemo_rl/sft.py#L137) and I'm not 100% sure this format is supported there. You could launch am `omr-simple-recipe` slurm test for this https://github.com/NVIDIA/NeMo-Skills/tree/main/tests/slurm-tests or run manually. Let me know if you need help with details on how to do that.

nemo_skills/pipeline/utils/exp.py
```diff
@@ -227,10 +228,7 @@ def get_executor(
                 slurm_kwargs = {}
             slurm_kwargs["exclusive"] = True
 
-    if "timeouts" not in cluster_config:
-        timeout = "10000:00:00:00"
-    else:
-        timeout = cluster_config["timeouts"][partition]
+    timeout = get_timeout(cluster_config, partition)
```
Kipok: here we shouldn't subtract 15 minutes. That logic should only be applied for training scripts. Maybe you can take this moment to make subtracted time a parameter that defaults to 0 and then update training scripts to set it to 15 minutes and then we can reuse this function in both cases. Or we subtract in a separate function that's only called in training scripts

artbataev: Fixed

nemo_skills/pipeline/utils/cluster.py
artbataev: Yep. I was trying to launch the job on cluster, and got "invalid time format" error with default timeout.
I investigated and found that this parameter is directly passed to SBATCH without modifications by nemo-run.
I can now run jobs on the cluster (I've just checked with latest changes).
So, there were two issues with previous code: 
- default timeout was not working (due to "days:..." instead of "days-...")
- format with "days" was not accepted (error in parsing)

I will double-check with tests, thank you for pointing on the tests

cluster_configs/example-slurm.yaml
artbataev: `default_timeout` looks a bit cleaner (but if the cluster has `default` partition, we could let its' timeout to be default :) )
Changed to `default_timeout`

nemo_skills/pipeline/utils/cluster.py
artbataev: If you are worried about different clusters, I can use something like 

```python
timeout_str_no_days = f"{timeout.seconds // 3600:02d}:{(timeout.seconds % 3600) // 60:02d}:{timeout.seconds % 60:02d}"
reduced_timeout_str = f"{timeout.days}-{timeout_str_no_days}" if timeout.days > 0 else timeout_str_no_days
```

artbataev: Launching jobs words as expected. Regarding tests - most of them work, there are some connectivity issues with HF, that I observed also on other jobs (but this is unrelated to changes).

Kipok: the problem is that current format is actually incompatible with logic in training framework, e.g. in nemo-rl here https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/utils/timer.py#L251. The training actually fails on this branch. So we should add some logic to use DD:HH:MM:SS when we use this in training scripts and keep your current format for slurm

artbataev: Looks like the confusion about time format goes significantly deeper than I was expecting. Let me check all tests completely

Kipok (approved):
looks good to me as long as your tests pass

artbataev:
@Kipok Thanks a lot!

PR #855, opened 2025-09-28
Fix HF_TOKEN assignment. Fix env vars priority: config -> environment -> default factory

artbataev:
## 1. Fix the issue when `HF_TOKEN` present in the config, but not present in the environment variables

Previously, in this condition, `HF_TOKEN` was set to `None`.
Reason: in `default_factories` `get_token()` returned None, which was converted to string "None".
When iterating over `optional_env_vars + always_optional_env_vars`, firstly `HF_TOKEN` was used from cluster config (with logging), and after that it was silently changed to "None" (without logging).

## 2. Fix env vars priority for optional variables: cluster config -> environment -> non-empty from default factory 

Previously: if the variable was in the environment, the value was silently (without logging) replaced with one from env.

## Story behind the issue
I was trying a first cluster example listed in the documentation from my laptop, and got many failures with error "401 Client Error: Unauthorized for url: https://huggingface.co/Qwen/Qwen2.5-1.5B-Instruct/resolve/main/config.json" (but no failure on my other desktop). 

```bash
ns generate \
    --cluster=draco-oci \
    --partition=interactive \
    --server_type=vllm \
    --model=Qwen/Qwen2.5-1.5B-Instruct \
    --server_gpus=1 \
    --input_file=/nemo_run/code/input.jsonl \
    ++prompt_config=/nemo_run/code/prompt.yaml \
    --output_dir=/workspace_mnt/exp/generation
```
I found that in sbatch script `HF_TOKEN` was set to string "None", despite that cluster config contained `HF_TOKEN`, and the code logged that `HF_TOKEN` was added to the job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents empty optional environment variables from overwriting existing values, reducing configuration regressions.
  * Improves detection and handling of authentication tokens to lower the chance of failed requests when a token is absent or malformed.

* **Chores**
  * Makes environment variable handling more consistent and reliable across runs, with clearer logging for missing or empty values.


artbataev:
> > I personally would suggest doing the opposite - prioritizing config, since locally one can experiment with keys with much more powerful permissions than required for cluster jobs
> 
> Agree, the order should always be - if defined in config explicitly, we take that, if not defined, we take from env var, if not defined, we take from default factory. Could you please fix this, I think the current order is basically a bug. And if we do this, then we don't need this extra check for empty value, right? I would probably also just not assign empty value from default factory - if it's empty there, we should probably just consider it not set. And could you also please change logging so that default factory log states explicitly that it's from default factory?

Sure, thank you for the feedback! I will fix it tomorrow

PR #856, opened 2025-09-28
Adding support for arm64 containers

Kipok:
1. Updated nemo-skills container to ubuntu instead of debian to make it easier to install apptainer on arm64
2. Upgraded trtllm / sglang containers
3. Added instructions for building on arm64
4. Removed nemo-aligner slurm test as we will soon fully remove support
5. Rolled-back nemo-rl dockerfile to a stable version (0.7.0)

## Summary by CodeRabbit

- New Features
  - Added ARM64/aarch64 build guide and multi-platform build instructions.
  - Introduced a benchmarks environment for builds.

- Chores
  - Upgraded default container images: TensorRT‑LLM → 1.0.0, SGLang → v0.5.3rc1‑cu126, NeMo‑Skills → 0.7.1.
  - Updated base image and dependency installation for build images.

- Documentation
  - Updated Docker build README and docs to reflect new images and ARM64 notes.

- Tests
  - Bumped CI/test image tags, simplified SLURM test path, and increased one test execution timeout.


dockerfiles/Dockerfile.nemo-skills
```diff
@@ -1,12 +1,26 @@
...
+
+# Install Python and other dependencies
+RUN apt-get update && \
+    apt-get install -y \
+    python3.10 \
+    python3-pip \
+    curl \
+    git \
+    git-lfs && \
+    ln -s /usr/bin/python3 /usr/bin/python
```
activatedgeek: Always a good practice to use:
```shell
rm -rf /var/cache/apt/archives /var/lib/apt/lists/*
```
in every layer that uses `apt`. Significantly reduces image size.

See https://docs.docker.com/build/building/best-practices/#apt-get

activatedgeek (changes requested):

dockerfiles/Dockerfile.nemo-skills
Kipok: sure - that's just cache, right? So apt-get commands would work normally in the final image if we add this?

activatedgeek: Right, we always will have to run `apt update` once before installing a new package later though.

activatedgeek (approved):

PR #859, opened 2025-09-29
Add check for wandb parameters correctness before submitting a job

wedu-nvidia:
As we can only see W&B errors after the job starts running on the cluster — at least that has been my experience a few times — adding a pre-check might be a good idea.

Rules (based on W&B conventions):
    - Must be <= 128 characters

## Summary by CodeRabbit

* **New Features**
  * Added a public validation utility for Weights & Biases (wandb) project/name/group/id and integrated it into generation, training (SFT, PPO/GRPO), evaluation, and result-logging workflows.

* **Bug Fixes**
  * Enforces a 128-character maximum for wandb project/name/group/id fields and fails fast with clear errors to prevent misconfigured runs.


nemo_skills/utils.py
```diff
@@ -187,6 +187,27 @@ def init_wandb(project, name, exp_dir=None, verbose=False):
...
+    """
+    Validate a Weights & Biases (W&B) project name.
+
+    Rules (based on W&B conventions):
+      - Must be <= 128 characters
+      - Can contain only lowercase letters, digits, underscores (_), and hyphens (-)
+    """
+
+    # Check maximum length (W&B enforces a safe upper limit of 128 chars)
+    if len(wandb_project) > 128:
```
Kipok: is this just for the project? Not for the name itself? Might also need to have checks on name and group if there are restrictions on those

nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -269,7 +269,7 @@ def sft_nemo_rl(
     extra_arguments = f"{' '.join(ctx.args)}"
     LOG.info("Starting training job")
     LOG.info("Extra arguments that will be passed to the underlying script: %s", extra_arguments)
-
+    validate_wandb_project_name(wandb_project)
```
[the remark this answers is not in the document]
Kipok: probably a good idea to do this ^

nemo_skills/utils.py
wedu-nvidia: it could be, I can check a few more.

nemo_skills/pipeline/verl/ppo.py
```diff
@@ -124,6 +124,9 @@ def format_wandb_args(self, disable_wandb, wandb_project, expname):
             cmd = f"{cmd} trainer.logger=['console'] "
         else:
             cmd = f"{cmd} trainer.logger=['console','wandb'] "
+            validate_wandb_project_name(
+                wandb_project=wandb_project,
```
Kipok, line 128: I guess it must be using expname as wandb name in here?

Kipok (approved):
just a small comment about verl part, but looks good other than that

nemo_skills/pipeline/verl/ppo.py
wedu-nvidia: Surprisingly, I did not find wandb name and expname here, but based on experiment, it seems yes. 
I can add this line of check here.

wedu-nvidia: @Kipok Updated

PR #860, opened 2025-09-29
Allow setting random seeds for benchmark groups

Kipok:
## Summary by CodeRabbit

- Improvements
  - Accuracy reporting now selects the highest available averaged pass@1 metric per category (e.g., avg-of-k) and falls back to standard pass@1 when unavailable, improving relevance.
  - Enforces a consistent max-k across subsets and will flag mismatches to prevent inconsistent aggregation.

- Bug Fixes
  - Specifying a sample count on a benchmark group no longer aborts; a warning is logged and the override is applied to all benchmarks in the group.


shtoshni (approved):
Looks good. I think the one tricky part would be that the final metric calculation would depend on pass@1[avg-of-N], and that N would just be different for the different benchmarks because some or the other job would crash. So might require multiple reruns for bigger N's

PR #865, opened 2025-09-29
Generation time + Input Sequence Length

shtoshni:


nemo_skills/inference/generate.py
```diff
@@ -520,8 +508,20 @@ async def process_single_datapoint(self, data_point, all_data):
...
         result = await self.llm.generate_async(**generation_params)
 
+        end_time = time.time()
+        # Add the generation time and input sequence length
+        if self.cfg.add_generation_stats:
+            result["generation_start_time"] = start_time
+            result["generation_end_time"] = end_time
+            result["generation_time"] = end_time - start_time
+
+        result["input_sequence_length"] = input_sequence_length
```
Kipok: should we not set it if it's None?

nemo_skills/prompt/utils.py
```diff
@@ -302,6 +302,29 @@ def fill(
...
+            int | None: Token count, or None if no tokenizer is set.
+        """
+        if self.tokenizer is None:
+            return None
+
+        if isinstance(messages, str):
+            return len(self.tokenizer.encode(messages, add_special_tokens=False))
+        elif isinstance(messages, list):
+            try:
+                return len(self.tokenizer.apply_chat_template(messages, tokenize=True))
```
Kipok:
```suggestion
                return len(self.tokenizer.apply_chat_template(messages, tokenize=True, add_generation_prompt=True))
```
is a little more accurate?

nemo_skills/inference/generate.py
```diff
@@ -471,22 +471,10 @@ def dump_outputs(self, outputs, data_points, fout):
...
-                if "generation_end_time" in original_data_point and "generation_time" in original_data_point:
-                    output["generation_start_time"] = (
-                        original_data_point["generation_end_time"] - original_data_point["generation_time"]
-                    )
-                else:
-                    output.pop("generation_start_time", None)
+            if not self.cfg.add_generation_stats:
+                output.pop("generation_start_time", None)
+                output.pop("generation_end_time", None)
+                output.pop("generation_time", None)
```
Kipok, line 497: we should also pop input_sequence_length here as otherwise judge jobs will override it

nemo_skills/inference/generate.py
shtoshni: Sure thing

nemo_skills/inference/generate.py
```diff
@@ -520,8 +508,25 @@ async def process_single_datapoint(self, data_point, all_data):
             if self.cfg.override_max_code_executions and self.cfg.total_code_executions_in_prompt is not None:
                 generation_params["max_code_executions"] = data_point["total_code_executions"]
 
+        # Tracking the tokens and generation time
+        input_sequence_length = None
+        if self.prompt is not None:
+            if generation_params["prompt"] is not None:
```
Kipok: we should ideally still calculate this statistic as we still have access to tokenizer even if prompt isn't defined

nemo_skills/inference/generate.py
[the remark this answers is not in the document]
shtoshni: wow, coderabbit you're amazing

nemo_skills/inference/generate.py
```diff
@@ -520,8 +508,25 @@ async def process_single_datapoint(self, data_point, all_data):
...
+                input_sequence_length = self.prompt.get_token_count(generation_params["prompt"])
+
+        # Start to track the generation time
+        start_time = time.time()
```
Kipok: actually could you move this outside of this function? Otherwise it's not going to work for cases when process_single_datapoint is overriden, which is quite common. We should just measure time for that method directly

shtoshni: Good point!

nemo_skills/inference/generate.py
shtoshni: The only way out is to initialize the tokenizer then. I added that part.

nemo_skills/inference/generate.py
```diff
@@ -321,8 +328,9 @@ def __init__(self, cfg: GenerateSolutionsConfig):
         self.output_lock = None
 
     def setup_prompt(self):
+        prompt = None
         if self.cfg.prompt_format == "openai":
-            return None
+            prompt = None
 
         prompt = get_prompt(
```
Kipok, line 346: this will fail for openai type as there is no prompt config

nemo_skills/inference/generate.py
```diff
@@ -271,8 +272,14 @@ def __init__(self, cfg: GenerateSolutionsConfig):
         if self.cfg.use_completions_api or self.cfg.server.get("enable_soft_fail", False):
             # These are the only cases where we need a tokenizer
             self.tokenizer = self.cfg.tokenizer or self.cfg.server["model"]
+            try:
```
Kipok: then we want to do it outside of this if, right? As we always want to try to use it for prompts?

on a second thought, how about we maybe don't log prompt length by default as for some cases (e.g. ruler) this will incur a non-trivial cost and might slow down generations. So maybe we add a parameter `count_prompt_tokens` that's false by default, but can be set to enable this? And then if it's set, we try to get tokenizer, but if we can't we actually hard-fail as users explicitly asked for that, but we can't do it

shtoshni: Good point!

nemo_skills/inference/generate.py
```diff
@@ -522,6 +531,10 @@ async def process_single_datapoint(self, data_point, all_data):
 
         result = await self.llm.generate_async(**generation_params)
 
+        if self.cfg.count_prompt_tokens:
+            input_sequence_length = get_token_count(self.hf_tokenizer, generation_params["prompt"])
+            if input_sequence_length is not None:
```
Kipok: can remove this if, I guess as we always expect this to be not None?

Kipok (approved):
one last small comment but otherwise looks good to me! Thanks!

nemo_skills/inference/generate.py
shtoshni: hmm i added it for precaution if prompt is None for some reason.

Kipok: but that will just crash the generation, right?

Kipok:
@shtoshni could you please try to update the model in generation tests from llama to nvidia/nvidia-nemotron-nano-9b-v2? I think the capacity for llama is maybe reduced and we are hitting service errors, but hopefully the nano model is served well there

nemo_skills/inference/generate.py
shtoshni: it should, let me remove it then.

shtoshni:
Actually Qwen small models are not hosted there

tests/test_generation.py
```diff
@@ -53,11 +53,11 @@ def test_eval_judge_api(tmp_path):
     cmd = (
         f"ns eval "
         f"    --server_type=openai "
-        f"    --model=meta/llama-3.1-8b-instruct "
+        f"    --model=nvidia/nvidia-nemotron-nano-9b-v2"
         f"    --server_address=https://integrate.api.nvidia.com/v1 "
         f"    --benchmarks=math-500 "
         f"    --output_dir={tmp_path} "
-        f"    --judge_model=meta/llama-3.1-8b-instruct "
+        f"    --judge_model=nvidia/nvidia-nemotron-nano-9b-v2 "
```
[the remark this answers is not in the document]
Kipok: good catch, rabbit! 🙂

tests/test_generation.py
```diff
@@ -53,11 +53,11 @@ def test_eval_judge_api(tmp_path):
     cmd = (
         f"ns eval "
         f"    --server_type=openai "
-        f"    --model=meta/llama-3.1-8b-instruct "
+        f"    --model=nvidia/nvidia-nemotron-nano-9b-v2"
```
Kipok:
```suggestion
        f"    --model=nvidia/nvidia-nemotron-nano-9b-v2 "
```

Issue #854, opened 2025-09-28
Create infrastructure for building containers and uploading them to clusters


Kipok:
Currently this is done manually but it is getting out of hand. We should automate this properly

PR #866, opened 2025-09-29
Add docker build script

activatedgeek:
Add a general script to be used later for building all images, as part of automated builds.

For #854 

## Summary by CodeRabbit

* **New Features**
  * Added a build utility to create (and optionally push) Docker images; it auto-determines image name and tag from Git, flags uncommitted changes, selects appropriate build context, and supports registry caching and multi-platform builds.

* **Documentation**
  * Replaced prior build instructions with usage for the new utility; documented environment variables for name, tag, push, and platform, and expanded arm64/aarch64 guidance.

### DIFF
--- a/dockerfiles/build.sh
+++ b/dockerfiles/build.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


Kipok (commented):
is there support for building for arm64 here?

activatedgeek:
Let me add those flags as well.

dockerfiles/build.sh
```diff
@@ -0,0 +1,90 @@
...
+        __project_name=/$(echo "${__project_name}" | cut -d. -f2)
+    else
+        unset __project_name
+    fi
+    DOCKER_NAME="gitlab-master.nvidia.com/${__repo_user}/${__repo_name}${__project_name}"
+fi
+DOCKER_NAME=$(echo "${DOCKER_NAME}" | tr "[:upper:]" "[:lower:]")
+
+if [[ -z "${DOCKER_TAG}" ]]; then
+    __git_sha=$(git rev-parse --short HEAD 2>/dev/null)
```
gwarmstrong, line 81: Can you put some check to make sure working directory is clean?

Also give an override for the check (people will find a way to work around it if you don't),  but make sure to add a `-dirty` suffix so we don't mix up clean and dirty images.

dockerfiles/build.sh
```diff
@@ -0,0 +1,90 @@
...
+    if [[ $? -ne 0 ]]; then
+        echo "Dockerfile is not part of a git repo. Set DOCKER_TAG explicitly."
+        exit 1
+    fi
+    DOCKER_TAG="$(date +"%Y.%m.%d")-${__git_sha}"
```
gwarmstrong, line 86: nit: I personally would push two tags--one with date and one without. That way the ones with date are easy to sort more or less chronologically, but if you want to find the image associated with a specific commit, it has a predictable tag.

dockerfiles/build.sh
```diff
@@ -0,0 +1,90 @@
...
+
+## If not a git repo, we go back to working directory.
+if [[ $__is_git_repo -eq 0 ]]; then
+    popd > /dev/null
+fi
+
+if [[ -z "${DOCKER_NAME}" ]]; then
+    __git_remote=$(git remote get-url origin 2>/dev/null)
+    if [[ $? -ne 0 ]]; then
+        echo "Dockerfile is not part of a git repo. Set DOCKER_NAME explicitly."
```
gwarmstrong: why is this necessary? just so it works with arbitrary dockerfiles?

gwarmstrong (changes requested):
One small request  for change and a couple comments/questions

dockerfiles/build.sh
activatedgeek: Good idea! For us though, this file will likely only be run in CI, so likely wouldn't be an issue. Fixed now.

dockerfiles/build.sh
activatedgeek: The pushing two images part is something that I can make changes in the CI. This file is just an atomic unit to build one Dockerfile. But I can make that change.

Currently, I was setting the latest tag and this auto-generated tag. I will change that to the following tags:
1. git SHA.
2. date
3. `latest`

Does that sound good? To keep things recognizable locally, perhaps I can let this be? Or do you think I should switch to default git SHA?

dockerfiles/build.sh
activatedgeek: Yep, solely to allow building any dockerfile.

dockerfiles/README.md
```diff
@@ -1,19 +1,27 @@
...
+./build.sh Dockerfile.nemo-skills
 ```
-It might take a long time for some of the images.
+
+Key configuration environment variables:
+- `DOCKER_NAME`: A fully qualified name of the docker image. The default is inferred from the git repository attributes.
+- `DOCKER_TAG`: Docker tag to use. Defaults to `yyyy.mm.dd-<commit_hash>`
+- `DOCKER_PUSH`: When set, pushes image after building.
+- `DOCKER_PLATFORM`: Directly passed to `--platform` for [multi-platform builds](https://docs.docker.com/build/building/multi-platform/).
+
```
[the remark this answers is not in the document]
Kipok: @activatedgeek this sounds like an issue?

dockerfiles/README.md
```diff
@@ -1,19 +1,27 @@
 # How to build all necessary dockerfiles
 
 Some dockerfiles are directly included in this folder and for some others the instructions to build them are below.
-To build one of the existing dockerfiles use a command like this
 
+To build one of the existing dockerfiles use a command like this
 ```
-docker build -t igitman/nemo-skills:0.7.1 -f dockerfiles/Dockerfile.nemo-skills .
+./build.sh Dockerfile.nemo-skills
```
Kipok, line 12: Maybe you can keep the original command as well just to give a simple path? And also say that build.sh can be used for some additional automation. The build.sh logic is a bit hard to understand from quick glance, so it's best to let people know that there is nothing very important going on there and they should be good building images with a straightforward docker build command as well

dockerfiles/build.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
Kipok, line 1: license header

dockerfiles/README.md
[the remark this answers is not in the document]
activatedgeek: I've removed the prefix now, since we set this explicitly in the CI anyway.

dockerfiles/README.md
activatedgeek: Fixed.

dockerfiles/build.sh
activatedgeek: Fixed.

dockerfiles/build.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
Kipok:
```suggestion
# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```

PR #872, opened 2025-09-30
Remove sharding docs

smahdavi4:
## Summary by CodeRabbit

- Documentation
  - Removed obsolete sharded model conversion section.
  - Updated dataset guides to use the non-sharded DeepSeek-R1 path and revised server arguments to reflect the new setup.

- Chores
  - Updated solution generation configs to use the non-sharded DeepSeek-R1 path.
  - Switched server arguments to an explicit parallelism setting, aligning configs with the updated runtime.
  - Simplifies setup and removes reliance on sharded model artifacts.


Kipok (approved):
thanks!

PR #878, opened 2025-10-02
add code timeout metrics

wedu-nvidia:
https://github.com/NVIDIA/NeMo-Skills/issues/834

## Summary by CodeRabbit

* **New Features**
  * Responses now include a code_timeout field reporting how many executed code blocks timed out during a run.
  * Timeout occurrences are aggregated per request and returned alongside other execution details.
  * No other behavior changes to execution results; all existing fields remain unchanged.

### DIFF
--- a/nemo_skills/inference/model/code_execution.py
+++ b/nemo_skills/inference/model/code_execution.py
@@ -126,6 +126,7 @@ class CodeExecutionWrapper:
         generation_time = 0
         code_execution_time = 0
         stopped_on_repetition = False
+        code_timeout = 0
         # adding plus one to make sure there is always some completion after the last requested code block
         try:
             for generation_index in range(effective_max_code_executions + 1):
@@ -185,6 +186,9 @@ class CodeExecutionWrapper:
                         remaining_code_executions,
                     )
 
+                    if "process_status" in execution_dict and execution_dict["process_status"] == "timeout":
+                        code_timeout += 1
+
                     if is_openai_format:
                         request["prompt"][-2]["content"] += code_output
                     else:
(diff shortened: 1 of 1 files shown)


nemo_skills/inference/model/code_execution.py
```diff
@@ -126,6 +126,7 @@ async def _generate_single(
         generation_time = 0
         code_execution_time = 0
         stopped_on_repetition = False
+        code_timeout = 0
```
activatedgeek: Can we rename this variable to be more descriptive?

Perhaps, `num_code_timeouts`? `code_timeout` could also refer to a flag instead of a count, so it would be nice to avoid ambiguity.

activatedgeek (changes requested):

nemo_skills/inference/model/code_execution.py
wedu-nvidia: Sure, I can change it.
@Kipok Do we just add this to generations?  As the issue says, adding the average number of code timeout metrics. Can you clarify it a little bit more?

Kipok: we want to eventually add this to metrics, but we can merge this one first and do it separately. I think @i-vainn might also have some WIP on that, so he can comment. Main issue is 1) code_execution.py is somewhat deprecated, so we don't want to have to much implementation related to that 2) when we do it through tool-calling interface which should eventually fully support this, we might want to have a more general way to report tool-specific metrics. 

So I'm not sure what's the best implementation right now, but at least keeping it in output files is already a good change

wedu-nvidia: Thanks for the clarification, I revised the variable and can update this later when new wip is added for this.

activatedgeek (approved):

PR #880, opened 2025-10-02
Add Brave summarizer search as MCP tool

activatedgeek:
## Summary by CodeRabbit

- New Features
  - Added a Brave web search summarization tool accessible via MCP with CLI and API-key support.

- Improvements
  - Standardized --api-key handling across tools; CLI now enforces presence of API keys at startup.
  - Exa tool renamed to "exa" with a simplified public interface and improved output handling and error reporting.

- Refactor
  - Removed legacy Exa helper connectors and narrowed utility surface to streamline integrations.

### DIFF
--- a/nemo_skills/mcp/servers/brave_tool.py
+++ b/nemo_skills/mcp/servers/brave_tool.py
@@ -5,11 +5,9 @@ from dataclasses import dataclass
 from typing import Annotated
 
 import httpx
-from mcp import StdioServerParameters
 from mcp.server.fastmcp import FastMCP
 from pydantic import Field
 
-from nemo_skills.mcp.clients import MCPStdioClient
 from nemo_skills.mcp.tool_providers import MCPClientTool
 
 logger = logging.getLogger(__name__)
@@ -33,9 +31,6 @@ async def summarize(
     query: Annotated[str, Field(description="Search query.")],
 ) -> ExecutionResult:
     """Get a summary of search results from the web using Brave."""
-
-    if not BRAVE_API_KEY:
-        return {"error": "Missing Brave API key"}
 
     base_url = "https://api.search.brave.com/res/v1"
 
@@ -74,25 +69,20 @@ class BraveSearchTool(MCPClientTool):
                     "args": ["-m", "nemo_skills.mcp.servers.brave_tool"],
                 },
                 "hide_args": {},
-                "init_hook": "nemo_skills.mcp.servers.brave_tool.stdio_connector",
+                "init_hook": "hydra",
             }
         )
 
 
-def stdio_connector(client: MCPStdioClient):
-    client.server_params = StdioServerParameters(
-        command=client.server_params.command,
-        args=list(client.server_params.args) + ["--api-key", os.getenv("BRAVE_API_KEY", "")],
-    )
-
-
 def main():
     parser = argparse.ArgumentParser(description="MCP server for Brave web search tool")
-    parser.add_argument("--api-key", type=str, required=False, help="Brave API Key")
+    parser.add_argument("--api-key", type=str, default=os.getenv("BRAVE_API_KEY"), help="Brave API Key")
     args = parser.parse_args()
 
+    if not args.api_key:
+        raise ValueError("Missing Brave API key.")
+
     global BRAVE_API_KEY
-    # Prefer CLI arg; do not fall back to environment unless explicitly desired
     BRAVE_API_KEY = args.api_key
 
     mcp.run(transport="stdio")
--- a/nemo_skills/mcp/servers/exa_tool.py
+++ b/nemo_skills/mcp/servers/exa_tool.py
@@ -20,12 +20,11 @@ from dataclasses import dataclass
 from typing import Annotated
 
 import httpx
-from mcp import StdioServerParameters
 from mcp.server.fastmcp import FastMCP
 from mcp.types import CallToolResult
 from pydantic import Field
 
-from nemo_skills.mcp.clients import MCPStdioClient, MCPStreamableHttpClient
+from nemo_skills.mcp.clients import MCPStreamableHttpClient
 from nemo_skills.mcp.tool_providers import MCPClientTool
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


nemo_skills/mcp/servers/brave_tool.py
```diff
@@ -0,0 +1,102 @@
...
+                "hide_args": {},
+                "init_hook": "nemo_skills.mcp.servers.brave_tool.stdio_connector",
+            }
+        )
+
+
+def stdio_connector(client: MCPStdioClient):
+    client.server_params = StdioServerParameters(
+        command=client.server_params.command,
+        args=list(client.server_params.args) + ["--api-key", os.getenv("BRAVE_API_KEY", "")],
```
gwarmstrong: We should be passing through the environment variables from the original process by default now. Can you try just grabbing BRAVE_API_KEY from the env in this file? It would be simpler to not have to pass it or deal with a custom connector.

nemo_skills/mcp/servers/brave_tool.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description="MCP server for Brave web search tool")
+    parser.add_argument("--api-key", type=str, required=False, help="Brave API Key")
```
gwarmstrong: possibly able to remove in favor of env variables

nemo_skills/mcp/servers/exa_tool.py
```diff
@@ -60,19 +65,19 @@ async def exa_websearch(
...
-        return {"result": response.json()["answer"]}
+
+    return {"result": response.json()["answer"]}
 
 
 def main():
     # Parse CLI arguments
     parser = argparse.ArgumentParser(description="MCP server for Exa web search tool")
-    parser.add_argument("--exa-api-key", dest="exa_api_key", type=str, required=False, help="Exa API key")
+    parser.add_argument("--api-key", type=str, required=False, help="Exa API key")
```
gwarmstrong: same here, can probably be removed in favor of env variables, but double check

nemo_skills/mcp/servers/exa_tool.py
```diff
@@ -83,6 +88,13 @@ def main():
 # ==============================
 
 
+def exa_stdio_connector(client: MCPStdioClient):
+    client.server_params = StdioServerParameters(
+        command=client.server_params.command,
+        args=list(client.server_params.args) + ["--api-key", os.getenv("EXA_API_KEY", "")],
```
gwarmstrong: same here

nemo_skills/mcp/servers/brave_tool.py
```diff
@@ -0,0 +1,102 @@
...
+
+## See docs https://api-dashboard.search.brave.com/app/documentation/summarizer-search/get-started
+@mcp.tool()
+async def summarize(
+    query: Annotated[str, Field(description="Search query.")],
+) -> ExecutionResult:
+    """Get a summary of search results from the web using Brave."""
+
+    if not BRAVE_API_KEY:
+        return {"error": "Missing Brave API key"}
```
Kipok: By the way, I think for errors like this for both this tool and python tool (and in general for any tools) we should do a hard fail and stop the job. These errors aren't recoverable and if users asked us to use a particular tool, we don't want to keep running it and just returning errors on every request. We should only soft-fail errors that could be recoverable, not missing api key or connection to sandbox kind of things

gwarmstrong (changes requested):
The Brave parts look fine to me, just look into if you can fully convert to using the environment variable for the api key directly, rather than relying on a custom connector.

nemo_skills/mcp/servers/brave_tool.py
activatedgeek: Good point. Let me just raise an error here.

@gwarmstrong To confirm, a raised exception here should fail the job, right?

Kipok: we might have some exception catching in the main class as well - I think we had a blank except for any tool errors or something like this. Would be good to add a test

gwarmstrong: should be easy to test by just setting the api key to None, yeah?

nemo_skills/mcp/servers/exa_tool.py
activatedgeek: Removed this function.

nemo_skills/mcp/servers/exa_tool.py
activatedgeek: I've updated this CLI arg to default to reading from the env.

nemo_skills/mcp/servers/brave_tool.py
activatedgeek: Update this arg to default to read from the env

nemo_skills/mcp/servers/brave_tool.py
activatedgeek: Removed this function.

nemo_skills/mcp/servers/brave_tool.py
```diff
@@ -0,0 +1,92 @@
+import argparse
```
Kipok, line 1: license header. By the way, we can probably automate this check with https://github.com/NVIDIA-NeMo/RL/blob/main/.github/workflows/copyright-check.yml

nemo_skills/mcp/servers/exa_tool.py
```diff
@@ -32,22 +36,18 @@ class ExecutionResult:
...
+mcp = FastMCP(name="exa")
 
 # Populated from CLI args in main()
 EXA_API_KEY: str | None = None
 
 
 @mcp.tool()
-async def exa_websearch(
-    query: Annotated[str, Field(description="Search query for Exa.")],
+async def answer(
```
Kipok, line 46: name of this function is being passed to an LLM as well as the description, so it might be important to make it more descriptive. You can see an example in python tool

nemo_skills/mcp/servers/brave_tool.py
activatedgeek: Created #909 to track this.

PR #881, opened 2025-10-02
Fix sandbox container build for arm64

wasiahmad:
This PR fixes the issue for #876 

## Summary by CodeRabbit

* **New Features**
  * Multi-architecture sandbox images (x86_64 and ARM64) with automatic selection and setup of the appropriate PyPy runtime.

* **Chores**
  * Add build argument to enable architecture-aware builds and explicit handling for unsupported architectures.
  * Update runtime symlinking and remove temporary artifacts and package cache to reduce image size.

### DIFF
--- a/dockerfiles/Dockerfile.sandbox
+++ b/dockerfiles/Dockerfile.sandbox
@@ -27,7 +27,8 @@ RUN apt-get update && \
     tar -xjf /tmp/pypy.tar.bz2 -C /opt/ && \
     ln -s /opt/pypy3.10-v7.3.17-$PYPY_ARCH/bin/pypy3 /usr/bin/pypy3 && \
     /usr/bin/pypy3 -m ensurepip && \
-    rm /tmp/pypy.tar.bz2
+    rm /tmp/pypy.tar.bz2 && \
+    rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 
 # Install the DMOJ judge-server using pip (OJBench eval requirement)
 RUN pip install git+https://github.com/DMOJ/judge-server.git@11bf2cd03df83f0df5970a08e98b4cec2dfaecd5
--- a/dockerfiles/Dockerfile.sandbox
+++ b/dockerfiles/Dockerfile.sandbox
@@ -16,7 +16,6 @@
 FROM tiangolo/uwsgi-nginx-flask:python3.10
 
 # Install dependencies required for Lean 4, pypy3, and other tools
-ARG TARGETARCH
 RUN apt-get update && \
     apt-get install -y curl git net-tools bzip2 build-essential libseccomp-dev && \
     case "$TARGETARCH" in \
--- a/dockerfiles/Dockerfile.sandbox
+++ b/dockerfiles/Dockerfile.sandbox
@@ -19,10 +19,12 @@ FROM tiangolo/uwsgi-nginx-flask:python3.10
 ARG TARGETARCH
 RUN apt-get update && \
     apt-get install -y curl git net-tools bzip2 build-essential libseccomp-dev && \
-    case "$TARGETARCH" in \
+    ARCH="${TARGETARCH:-$(dpkg --print-architecture)}" && \
+    case "$ARCH" in \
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile.sandbox
```diff
@@ -16,11 +16,13 @@
 FROM tiangolo/uwsgi-nginx-flask:python3.10
 
 # Install dependencies required for Lean 4, pypy3, and other tools
+ARG TARGETARCH
```
activatedgeek, line 19: Docker build already provides the specified platform variables as argument. See `TARGET*` variables https://docs.docker.com/build/building/variables/#multi-platform-build-arguments.

Can we instead reuse them?

activatedgeek (changes requested):

dockerfiles/Dockerfile.sandbox
```diff
@@ -18,9 +18,14 @@ FROM tiangolo/uwsgi-nginx-flask:python3.10
...
+        amd64) PYPY_ARCH=linux64 ;; \
+        arm64|aarch64) PYPY_ARCH=aarch64 ;; \
+    *) echo "Unsupported TARGETARCH '$TARGETARCH'" >&2; exit 1 ;; \
+    esac && \
+    curl -L https://downloads.python.org/pypy/pypy3.10-v7.3.17-$PYPY_ARCH.tar.bz2 -o /tmp/pypy.tar.bz2 && \
     tar -xjf /tmp/pypy.tar.bz2 -C /opt/ && \
-    ln -s /opt/pypy3.10-v7.3.17-linux64/bin/pypy3 /usr/bin/pypy3 && \
+    ln -s /opt/pypy3.10-v7.3.17-$PYPY_ARCH/bin/pypy3 /usr/bin/pypy3 && \
     /usr/bin/pypy3 -m ensurepip && \
     rm /tmp/pypy.tar.bz2
```
activatedgeek: Always a good practice to use for all apt-related layers:
```shell
rm -rf /var/cache/apt/archives /var/lib/apt/lists/*
```

See https://docs.docker.com/build/building/best-practices/#apt-get.

activatedgeek (changes requested):

dockerfiles/Dockerfile.sandbox
```diff
@@ -18,9 +18,14 @@ FROM tiangolo/uwsgi-nginx-flask:python3.10
 # Install dependencies required for Lean 4, pypy3, and other tools
 RUN apt-get update && \
     apt-get install -y curl git net-tools bzip2 build-essential libseccomp-dev && \
-    curl -L https://downloads.python.org/pypy/pypy3.10-v7.3.17-linux64.tar.bz2 -o /tmp/pypy.tar.bz2 && \
+    case "$TARGETARCH" in \
```
activatedgeek: Does this need to redefined after the first `FROM` call?

As it recommends here: https://docs.docker.com/build/building/variables/#multi-platform-build-arguments

dockerfiles/Dockerfile.sandbox
wasiahmad: @activatedgeek where should I add this line?

wasiahmad: oh ok, understood, made the changes.

dockerfiles/Dockerfile.sandbox
wasiahmad: Just understood the issue, I think in our particular scenario, we perhaps don't need it.

activatedgeek: Sounds good. As long as it is doing the right thing after building.

dockerfiles/Dockerfile.sandbox
```diff
@@ -18,11 +18,17 @@ FROM tiangolo/uwsgi-nginx-flask:python3.10
...
```
activatedgeek: `Unsupported TARGETARCH ''`

@wasiahmad You'll need to add `ARG TARGETARCH` after the first from. Otherwise this fails.

activatedgeek (changes requested):

PR #882, opened 2025-10-02
update patch

wedu-nvidia:
Update the container to cherry-pick some code changes (as pr not merged from nemo-rl) to accelerate SFT training. 
For our 30b with 16k context window training, we see 30-40% speedup.

### DIFF
--- a/dockerfiles/nemo-rl.patch
+++ b/dockerfiles/nemo-rl.patch
@@ -1,5 +1,5 @@
 diff --git a/nemo_rl/algorithms/sft.py b/nemo_rl/algorithms/sft.py
-index 6de9ac81..fc232ac6 100644
+index 6de9ac8..fc232ac 100644
 --- a/nemo_rl/algorithms/sft.py
 +++ b/nemo_rl/algorithms/sft.py
 @@ -137,6 +137,7 @@ def setup(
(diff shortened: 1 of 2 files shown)


dockerfiles/nemo-rl.patch
```diff
@@ -19,15 +19,83 @@ index 6de9ac81..fc232ac6 100644
...
 +        if "layernorm_epsilon" in self.cfg["megatron_cfg"]:
 +            model_cfg.layernorm_epsilon = self.cfg["megatron_cfg"]["layernorm_epsilon"]
 
          checkpoint_config = CheckpointConfig(
              save_interval=100,
+@@ -918,9 +923,7 @@ class MegatronPolicyWorker:
+                         do_not_average_loss=True,
+                     )
+
+-                # Empty unused memory.
```
Kipok: I'm a little confused, why do we need this patch if the parameter is there already? Won't it work if we just set empty_unused_memory_level=0 directly?

nemo_skills/training/nemo_rl/configs/sft.yaml
```diff
@@ -164,12 +165,14 @@ data:
   output_key: output
   force_reprocess: false
   shuffle: true
+  num_workers: 1
```
Kipok: let's set to 10 by default?

nemo_skills/training/nemo_rl/configs/sft.yaml
```diff
@@ -164,12 +165,14 @@ data:
...
```
Kipok, line 169: should we also set empty_unused_memory_level=0?

PR #884, opened 2025-10-03
Adding transformers server type

Kipok:


nemo_skills/pipeline/utils/server.py
```diff
@@ -123,7 +125,7 @@ def get_server_command(
 
     # check if the model path is mounted if not vllm, sglang, or trtllm;
     # vllm, sglang, and trtllm can also pass model name as "model_path" so we need special processing
-    if server_type not in ["vllm", "sglang", "trtllm"]:
+    if server_type in ["megatron"]:
```
activatedgeek, line 128: Should we update the comment since the logic has changed?

activatedgeek (approved):

PR #888, opened 2025-10-03
Refactor task construction for generation

gwarmstrong:
## Summary by CodeRabbit

- **New Features**
  - Declarative pipeline API for composing and running jobs with dependencies, heterogeneous multi‑group execution, run‑after semantics, sandbox support, and unified hardware/metadata propagation.
  - Helpers for server/sandbox orchestration with automatic port selection and environment wiring, plus a command wrapper for env/working-dir composition.
  - Utility to build grouped generation/server tasks with reuse support.

- **Tests**
  - Expanded test coverage for pipeline behaviors, dependency resolution, sandbox/server metadata, heterogeneous jobs, and reuse.

### DIFF
--- a/nemo_skills/pipeline/utils/declarative.py
+++ b/nemo_skills/pipeline/utils/declarative.py
@@ -276,7 +276,7 @@ class Pipeline:
         jobs: List[Dict],
         reuse_code: bool = True,
         reuse_code_exp: Optional[str] = None,
-        skip_hf_home_check: bool = False,
+        skip_hf_home_check: bool | None = None,
         with_ray: bool = False,
         run_after: Optional[Union[str, List[str]]] = None,  # Pipeline-level dependency on other experiments
     ):
@@ -284,6 +284,9 @@ class Pipeline:
         self.cluster_config = cluster_config
         self.reuse_code = reuse_code
         self.reuse_code_exp = reuse_code_exp
+        # If not explicitly set, resolve from cluster config (matching exp.py behavior)
+        if skip_hf_home_check is None:
+            skip_hf_home_check = cluster_config.get("skip_hf_home_check", False)
         self.skip_hf_home_check = skip_hf_home_check
         self.with_ray = with_ray
         self.run_after = run_after


nemo_skills/pipeline/utils/commands.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns:
+        Tuple of (lambda, metadata_dict)
+    """
+    if port is None:
+        port = get_free_port(strategy="random")
+
+    metadata = {
+        "port": port,
+        "log_prefix": "sandbox",
+        "mounts": [],  # Sandbox doesn't mount anything
```
Kipok: this has recently been changed, there is now a parameter that controls this

nemo_skills/pipeline/utils/commands.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    if port is None:
+        port = get_free_port(strategy="random")
+
+    metadata = {
+        "port": port,
+        "log_prefix": "server",
+    }
+
+    # Return lambda that will call get_server_command when cluster_config is available
```
Kipok: what's the reason we can't just accept cluster config as an argument?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+    Using a lambda allows references to work correctly in heterogeneous jobs:
+        Command(command=lambda: f"curl {server.hostname_ref()}:5000")
+
+    Metadata from command builders (like port) can be referenced:
+        server = Command(command=vllm_server_command(...))
+        client = Command(command=lambda: f"curl {server.hostname_ref()}:{server.meta_ref('port')}")
+    """
+
+    command: Union[
+        str,
```
Kipok: might be my personal preference, but I think such large types are just making things harder to read and aren't useful. If you have a clear type which is concise and helpful (is this a list or a dict) that's one thing, but if you're trying to type out something that can be 5 different things it's just taking a space and not really adding a value imo

and again in my experience this will very quickly go out of sync with the actual implementation and will be both very long and also incorrect which is worse than just not having it at all

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+    working_dir: str = "/nemo_run/code"
+    env_vars: Dict[str, str] = field(default_factory=dict)
+    installation_command: Optional[str] = None
+    port: Optional[int] = None  # Can be set from metadata
+    metadata: Dict[str, any] = field(default_factory=dict)  # Stores metadata from command builders
+    het_group_index: Optional[int] = None  # Set per-job by Pipeline (not global)
+
+    def __post_init__(self):
+        # Initialize defaults
+        if self.name is None:
```
Kipok: why not just set this as default value?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+            self.command = wrap_command(self.command, self.working_dir, self.env_vars)
+
+    def _extract_metadata(self):
+        """Extract metadata from command if it's a tuple."""
+        if not callable(self.command):
+            if isinstance(self.command, tuple):
+                cmd, self.metadata = self.command
+                self.command = cmd  # Can be None for sandbox
+
+    def hostname_ref(self) -> str:
```
Kipok, line 169: why _ref? server.hostname() looks a bit cleaner?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+        3. Wraps with installation_command if provided
+
+        Returns:
+            Tuple of (final_command, execution_config)
+        """
+        # 1. Evaluate if callable (resolves cross-group references or cluster_config needs)
+        if callable(self.command):
+            # Check if lambda needs cluster_config (for sandbox)
+
+            sig = inspect.signature(self.command)
```
Kipok: can we just always accept cluster_config? Would make this simpler and unify the signature?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+    """Command group where commands run together with shared resource requirements."""
+
+    def __init__(
+        self,
+        commands: List[Command],
+        hardware: Optional[HardwareConfig] = None,
+        name: Optional[str] = None,
+        log_dir: Optional[str] = None,
+    ):
+        self.components = commands
```
Kipok:
```suggestion
        self.commands = commands
```
for consistency?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+    Supports two modes:
+    1. Groups: groups=[cmdgroup1, cmdgroup2] - combines all into one job
+    2. Jobs: jobs=[{...}, {...}] - supports dependencies and multi-group jobs
+    """
+
+    def __init__(
+        self,
+        name: str,
+        cluster: Optional[str] = None,
+        groups: Optional[List[CommandGroup]] = None,  # Legacy mode
```
Kipok: why do we need a legacy mode?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+            cluster: Cluster name to load config from (optional if cluster_config provided)
+            dry_run: If True, validate without executing
+            log_dir: Default log directory for groups that don't specify one (optional)
+            _reuse_exp: Internal - reuse existing experiment object (for eval.py integration)
+        """
+        if not self.jobs:
+            LOG.info("No jobs to execute")
+            return None
+
+        # Determine cluster config to use
```
Kipok: why do we need to have multiple places of resolution and also some cases here? Would be better if cluster_config is just resolved in some place (probably outside this class) and then passed around?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+                    job_dependencies = []
+
+                # If no job-level dependencies, apply pipeline-level run_after
+                if not job_dependencies and self.run_after:
+                    run_after_list = self.run_after if isinstance(self.run_after, list) else [self.run_after]
+                    job_dependencies = run_after_list
+
+                for dep in job_dependencies:
+                    if isinstance(dep, str):
+                        # String dependency - could be job name, task handle, or experiment name
```
Kipok: feels like this can easily collide. Like if I have an experiment with generic name like "eval" and also define a job with this name, then we can't properly distinguish?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,748 @@
...
+                heterogeneous=heterogeneous,
+                het_group=het_group,
+                total_het_groups=total_het_groups,
+                overlap=overlap,
+                mounts=exec_config.get("mounts"),
+                with_ray=self.with_ray,
+                slurm_kwargs={"exclusive": hardware.exclusive} if (hardware and hardware.exclusive) else None,
+            )
+
+    def _plan_and_add_job(
```
Kipok, line 534: maybe we can split this file into some more modular utils? It's a bit too large currently, could probably benefit from separating some logic if possible

Kipok: but we can keep it as is if there is no natural split

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: this yields an expression like `$(scontrol show $SLURM_JOB_NODELIST_HET_GROUP_1 | head -n 1`. It's not really properly a hostname, because it is defined at runtime and has to be run evaluated--if you treat it like a hostname you will get it wrong

nemo_skills/pipeline/utils/commands.py
gwarmstrong: That's a little bit of a holdover from when I was using cluster_config for the het group configuration--but I've adjusted it slightly--so this one we could move to just return the command directly. There's a few places where lambda's are acceptable because you need to be able to work with a het group id before the het group index is defined. This isn't currently one of them, so will change 👍

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: Fair enough, the most important thing is to just understand that a lambda can be passed for cross group hosts, so I'll simplify the type hint and add that as a comment

nemo_skills/pipeline/utils/commands.py
gwarmstrong: fixed in b0f263301f06ed663da957edb9ae54bf0aeaf059

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: simplified in 5d1924a9c20508ca8e488021ce209f462896fb9e

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: b754e6c0e1ac0fd7e87a623913b405f007d8ba62

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: updated in 9dee62a1cfb3a035a2553e291495dfa05cc52989
much simpler

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: simplified in b90131b9230c500ea04a16214f68bda14011b2a6

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: updated in bd4f2bef071ac8c34eb527d379aa56ba32d1b764

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: we don't--it was really just wording, updated here: 13873a18cac3f6242de60d85265f471dcac30669
but we can also just remove "groups" and go to "jobs" completely, it's pretty much just a convenience syntax.

nemo_skills/pipeline/utils/commands.py
```diff
@@ -0,0 +1,139 @@
...
+        args: Additional server arguments
+        entrypoint: Custom entrypoint script
+
+    Returns:
+        Tuple of (command_string, metadata_dict)
+    """
+    if port is None:
+        port = get_free_port(strategy="random")
+
+    cmd, num_tasks = get_server_command(
```
activatedgeek, line 57: While we are at this, can we now isolate all logic in `get_server_command` into this function. In the other function, can we throw a warning,
```
import warnings
warnings.warn("deprecation message", DeprecationWarning)
```

In `get_server_command`, we can then just return the `cmd` from this function. 

I think we'll need to do this to slowly transition away from old designed functions, and keep track on things that need to be updated or removed.

nemo_skills/pipeline/utils/commands.py
```diff
@@ -0,0 +1,139 @@
...
+        cluster_config: Cluster configuration dictionary
+        port: Port to use for sandbox
+
+    Returns:
+        Tuple of (command_string, metadata_dict)
+    """
+    if port is None:
+        port = get_free_port(strategy="random")
+
+    cmd = get_sandbox_command(cluster_config)
```
activatedgeek, line 90: I suggest the same deprecation warning here. Whenever we get a chance next, we can then search for the deprecation warnings, and migrate away from those methods.

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,703 @@
...
+        # Wrap plain strings with environment setup
+        if isinstance(self.command, str) and (self.env_vars or self.working_dir):
+            self.command = wrap_command(self.command, self.working_dir, self.env_vars)
+
+    def hostname_ref(self) -> str:
+        """Get hostname reference for hetjob cross-component communication."""
+        if self.het_group_index is None:
+            return "127.0.0.1"  # Local fallback
+        # For heterogeneous SLURM jobs, resolve nodelist to actual hostname
+        return f"$(scontrol show hostnames $SLURM_JOB_NODELIST_HET_GROUP_{self.het_group_index} | head -n1)"
```
activatedgeek, line 174: Is this an environment variable set by slurm? I couldn't find it in the docs https://slurm.schedmd.com/sbatch.html

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,703 @@
...
+    env_vars: Dict[str, str] = field(default_factory=dict)
+    installation_command: Optional[str] = None
+    port: Optional[int] = None  # Can be set from metadata
+    metadata: Dict[str, any] = field(default_factory=dict)  # Stores metadata from command builders
+    het_group_index: Optional[int] = None  # Set per-job by Pipeline (not global)
+
+    def __post_init__(self):
+        # Wrap plain strings with environment setup
+        if isinstance(self.command, str) and (self.env_vars or self.working_dir):
+            self.command = wrap_command(self.command, self.working_dir, self.env_vars)
```
activatedgeek, line 167: Just a stylistic choice, but I will suggest methods like `wrap_command` be co-located with the `Command` data class, since they mostly won't have any other utility.

If they do end up with a utility outside, we can always make them a `@staticmethod`.

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,703 @@
...
+                    else:
+                        self.metadata[key] = value
+            else:
+                final_command = result
+        else:
+            final_command = self.command
+
+        # 2. Wrap with installation_command if provided
+        if self.installation_command:
+            final_command = install_packages_wrap(final_command, self.installation_command)
```
activatedgeek, line 215: Here also the same stylistic suggestion for co-locating `install_packages_wrap`.

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: good point--I updated it so that we use direct reference to the object to set dependencies within the pipeline, and we can use strings for any external pipelines. cd0a63a735dd31cbba80780a50ed8912aa626874

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: actually just went ahead and removed the `groups` in 670572e265d6422e63133a465ceb859257aa1b45, now there's just one path to configure the jobs.

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,721 @@
...
+
+        # 3. Build execution config from metadata
+        execution_config = {
+            "num_tasks": self.metadata.get("num_tasks", 1),
```
activatedgeek, line 219: We may want to document that `num_tasks` is always 1. But there's no reason it can't be more generally right?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,721 @@
...
+
+    def __init__(
+        self,
+        commands: List[Command],
+        hardware: Optional[HardwareConfig] = None,
+        name: Optional[str] = None,
+        log_dir: Optional[str] = None,
+    ):
+        self.commands = commands
+        self.hardware = hardware or HardwareConfig()
```
activatedgeek, line 257: How does this hardware config interact with the hardware-related arguments in `Command`?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,721 @@
...
+    def __init__(
+        self,
+        name: str,
+        cluster_config: Dict,
+        jobs: List[Dict],
+        reuse_code: bool = True,
+        reuse_code_exp: Optional[str] = None,
+        skip_hf_home_check: bool = False,
+        with_ray: bool = False,
+        run_after: Optional[Union[str, List[str]]] = None,  # Pipeline-level dependency on other experiments
```
activatedgeek, line 281: In this code's context, just confirming by experiments you mean other "pipelines"?

nemo_skills/pipeline/utils/declarative.py
activatedgeek: We definitely should! Perhaps the pipeline class itself could be its own file.

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -0,0 +1,721 @@
...
+                            reuse_exp = run.Experiment.from_id(reuse_exp)
+                        except Exception:
+                            try:
+                                reuse_exp = run.Experiment.from_title(reuse_exp)
+                            except Exception:
+                                LOG.warning(f"Failed to load experiment {reuse_exp} for code reuse")
+                                reuse_exp = None
+                    if reuse_exp is not None:
+                        LOG.info(f"Trying to reuse code from experiment {reuse_exp._title}")
+                        reuse_key = get_packaging_job_key(reuse_exp._id, "nemo-run")
```
activatedgeek, line 656: Again functions like these can perhaps we isolated along with the pipeline since they likely have no other utility.

nemo_skills/pipeline/generate.py
activatedgeek, line 1: Are we confident enough in our tests related to generate?

If our test coverage is low, I will strong recommend that we keep the old `generate.py` as `generate_legacy.py` so that people getting the new version automatically use the new command (with warnings of course). So in case, it does cause any bugs, they can always have an escape hatch for this transition phase.

activatedgeek (commented):
Looking really good! I like the general interface. Just some suggestions for how we can plan for a transition, and slowly phase out old code.

nemo_skills/pipeline/generate.py
Kipok: We should run a few more tests, but I would probably just go ahead and switch to the new interface directly. As long as there are no silent issues (so it doesn't corrupt generations, but just throws an error), it should be easy to fix, and we will quickly resolve all problems. If we find that there are some really tricky things that we didn't handle, we can always just roll this back or add generate_legacy.py whenever we face this situation

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: it's specifically in the heterogeneous jobs docs: https://slurm.schedmd.com/heterogeneous_jobs.html#env_var

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: I'll look into this--an experiment in this context is a nemo-run experiment handle, so you could build a pipeline as a dependency of of a job not run through the Pipeline construction. I'll try to put together a convenient way to get the experiment handle of a pipeline so you can also easily make dependent pipelines.

PR #894, opened 2025-10-06
Fix API Key

smahdavi4:
Setting api key to empty only when we have a self-hosted server.

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted API key validation to only require an OpenAI key when using the OpenAI endpoint, preventing unnecessary errors with other providers.
  - Preserved behavior for NVIDIA endpoints; they continue to work as before.
  - Improved compatibility with custom/non-OpenAI endpoints by attempting to use an OpenAI key if available but not failing when it’s absent.
  - Enhanced domain-aware error handling for clearer, context-specific feedback.

### DIFF
--- a/nemo_skills/inference/model/openai.py
+++ b/nemo_skills/inference/model/openai.py
@@ -52,7 +52,7 @@ class OpenAIModel(BaseModel):
                 api_key = os.getenv("NVIDIA_API_KEY")
                 if not api_key:
                     raise ValueError("NVIDIA_API_KEY is required for NVIDIA models and could not be found.")
-            elif "api.openai.com" in base_url:
+            else:
                 api_key = os.getenv("OPENAI_API_KEY")
                 if not api_key:
                     raise ValueError("OPENAI_API_KEY is required for OpenAI models and could not be found.")
--- a/nemo_skills/inference/model/openai.py
+++ b/nemo_skills/inference/model/openai.py
@@ -54,7 +54,7 @@ class OpenAIModel(BaseModel):
                     raise ValueError("NVIDIA_API_KEY is required for NVIDIA models and could not be found.")
             else:
                 api_key = os.getenv("OPENAI_API_KEY")
-                if not api_key:
+                if not api_key and "api.openai.com" in base_url:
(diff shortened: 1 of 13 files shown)


nemo_skills/inference/model/vllm.py
```diff
@@ -28,6 +28,10 @@ class VLLMModel(BaseModel):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+    def _get_api_key(self, api_key: str | None, api_key_env_var: str | None, base_url: str) -> str | None:
```
Kipok: Should we do it for megatron model as well?

smahdavi4: done.

smahdavi4:
The litellm's logic is to get it from `api_key` if that exists, otherwise it will read the environment variable for the key. So, having None as api key gives freedom to use other providers in the future, but I can remove the if statement from openai as well to be cleaner.

smahdavi4:
yeah makes sense. i reverted the changes and kept only the if else statement.

nemo_skills/inference/model/openai.py
```diff
@@ -52,7 +52,7 @@ def _get_api_key(self, api_key: str | None, api_key_env_var: str | None, base_ur
                 api_key = os.getenv("NVIDIA_API_KEY")
                 if not api_key:
                     raise ValueError("NVIDIA_API_KEY is required for NVIDIA models and could not be found.")
-            elif "api.openai.com" in base_url:
+            else:
                 api_key = os.getenv("OPENAI_API_KEY")
                 if not api_key:
                     raise ValueError("OPENAI_API_KEY is required for OpenAI models and could not be found.")
```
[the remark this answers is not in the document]
Kipok: sorry for going back-and-forth, @smahdavi4 can you then also remove the ValueError unless the url is matching openai? So we basically move the check for url to only raise an error, but if url is anything else we still pass in the value, but we don't fail if it's not checked

Kipok (approved):
Thanks!

PR #896, opened 2025-10-06
Token count for BFCL

shtoshni:
## Summary by CodeRabbit

- New Features
  - Added prompt token counting (num_input_tokens) for single- and multi-turn generations.
  - Exposes per-turn token stats: num_generated_tokens_list and optional num_input_tokens_list, plus aggregated totals.

- Refactor
  - Renamed output field from input_sequence_length to num_input_tokens for consistency.
  - Generation results now return structured payloads (including token counts) even on error and merge parsed outputs into results.
  - Token counting utility expanded to accept richer message formats and normalize non-dict messages.

- Chores
  - Cleanup now removes num_input_tokens when generation stats are disabled.

### DIFF
--- a/nemo_skills/inference/eval/bfcl.py
+++ b/nemo_skills/inference/eval/bfcl.py
@@ -21,7 +21,10 @@ from functools import partial
 import hydra
 from transformers import AutoTokenizer
 
-from nemo_skills.dataset.bfcl_v3.utils import convert_to_tool, func_doc_language_specific_pre_processing
+from nemo_skills.dataset.bfcl_v3.utils import (
+    convert_to_tool,
+    func_doc_language_specific_pre_processing,
+)
 from nemo_skills.inference.eval.bfcl_utils import (
     DEFAULT_USER_PROMPT_FOR_ADDITIONAL_FUNCTION_FC,
     MAXIMUM_STEP_LIMIT,
@@ -29,11 +32,20 @@ from nemo_skills.inference.eval.bfcl_utils import (
     execute_multi_turn_func_call,
     is_empty_execute_response,
 )
-from nemo_skills.inference.generate import GenerateSolutionsConfig, GenerationTask, InferenceConfig
+from nemo_skills.inference.generate import (
+    GenerateSolutionsConfig,
+    GenerationTask,
+    InferenceConfig,
+)
 from nemo_skills.inference.model import server_params
 from nemo_skills.inference.model.utils import is_context_window_exceeded_error
 from nemo_skills.prompt.utils import get_token_count
-from nemo_skills.utils import get_help_message, get_logger_name, nested_dataclass, setup_logging
+from nemo_skills.utils import (
+    get_help_message,
+    get_logger_name,
+    nested_dataclass,
+    setup_logging,
+)
 
 LOG = logging.getLogger(get_logger_name(__file__))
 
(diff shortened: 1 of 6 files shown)


nemo_skills/inference/eval/bfcl.py
```diff
@@ -227,6 +228,15 @@ async def _generate_single_assistant_turn(self, inference_state_dict):
 
         input_dict = self.message_parser.construct_input_dict(messages, tools)
 
+        return_dict = {}
+        if self.cfg.count_prompt_tokens:
+            try:
+                num_input_tokens = get_token_count(self.hf_tokenizer, input_dict["prompt"])
+                return_dict["num_input_tokens"] = num_input_tokens
+            except ValueError as e:
```
Kipok: why do we need this exception? We don't have it in base class, so probably shouldn't have it here?

shtoshni: When there's server side parsing, we keep the messages in the ChatCompletion format used by OpenAI - [here](https://github.com/NVIDIA/NeMo-Skills/blob/main/nemo_skills/inference/eval/bfcl.py#L180). We can't get token count for such messages in the list. However, with client-side parsing we can get the token count as the entire chat can be tokenized. Let me add the comment to explain the reason behind this try/catch. This try/catch is only needed because of multi-turn tasks + server side parsing.

Kipok: Why can't we count tokens for messages? As in - the current code needs an update to do that, or it's not possible in principle?

shtoshni: I don't think there's an easy way to do it tbh, at least from my cursory search I didn't find any solution. One contrived solution is to get the count from the ChatCompletion response, and somehow patch it with the user message lengths when calculating this for multi-turn.

Kipok: I don't fully understand - the problem is that we can't get counts for tool responses? As in why server side vs client side is a problem? The model will decide to call a tool and the output will be parsed by the server, not client, but I don't fully get why it affects input token counts. All the user messages should be available to us, should they not? Can you give an example of where this is tricky?

shtoshni: `Prompt: [{'role': 'system', 'content': '/no_think'}, {'role': 'user', 'content': "Move 'final_report.pdf' within document directory to 'temp' directory in document. Make sure to create the directory"}, Message(content=None, role='assistant', tool_calls=[ChatCompletionMessageToolCall(function=Function(arguments='{"dir_name": "temp"}', name='mkdir'), id='chatcmpl-tool-a186d74a8b6148779438725a0a7d8bb1', type='function')], function_call=None, provider_specific_fields={'refusal': None}), {'role': 'tool', 'content': 'None', 'tool_call_id': 'chatcmpl-tool-a186d74a8b6148779438725a0a7d8bb1'}]`

shtoshni: User messages are fine. We keep the assistant response in the `Message` object - coming from here.       
`output_dict["message"] = output_dict["response"].choices[0].message`

```
# Add the message to the state dict for chat history
state_dict["messages"].append(model_response["message"])
```

Kipok: but why do we care about assistant messages for calculating input tokens? Maybe I misunderstand something - let me know if we should discuss offline

shtoshni: Ok, I added the dictionary conversion part to get_token_count. Works for both server side and client side parsing. Let me know if changes are needed!

PR #897, opened 2025-10-06
Add check for mount paths before running experiment and QoL

titu1994:
# Changelog
- Check for remote cluster existance of mount paths before running experiment
- Add support for os.environ injection inside `env_vars` and `required_env_vars`

### DIFF
--- a/nemo_skills/pipeline/utils/cluster.py
+++ b/nemo_skills/pipeline/utils/cluster.py
@@ -134,11 +134,6 @@ def get_env_variables(cluster_config):
                 env_var_name, value = env_var.split("=")
                 env_var_name = env_var_name.strip()
                 value = value.strip()
-
-                # If the value contains a '$', we treat it as an env var reference
-                if "$" in value:
-                    value = os.path.expanduser(value)
-
             else:
                 raise ValueError(f"Invalid required environment variable format: {env_var}")
             env_vars[env_var_name] = value
@@ -202,6 +197,18 @@ def get_env_variables(cluster_config):
             if env_var_name not in _logged_optional_env_vars:
                 LOG.info(f"Optional environment variable {env_var_name} not found in user environment; skipping.")
                 _logged_optional_env_vars.add(env_var_name)
+
+    # replace placeholders with actual env var values
+    for key, value in env_vars.items():
+        if isinstance(value, str) and "$" in value:
+            if key in os.environ:
+                env_vars[key] = os.path.expandvars(value)
+                LOG.info(
+                    f"Resolved environment variable {key} inside the placeholder value: {value} with {env_vars[key]}"
+                )
+            else:
+                raise ValueError(f"Cannot resolve environment variable {key} inside the placeholder value: {value}")
+
     return env_vars
 
 
--- a/nemo_skills/pipeline/utils/cluster.py
+++ b/nemo_skills/pipeline/utils/cluster.py
@@ -134,6 +134,11 @@ def get_env_variables(cluster_config):
                 env_var_name, value = env_var.split("=")
                 env_var_name = env_var_name.strip()
                 value = value.strip()
+
+                # If the value contains a '$', we treat it as an env var reference
+                if "$" in value:
+                    value = os.path.expanduser(value)
+
             else:
                 raise ValueError(f"Invalid required environment variable format: {env_var}")
             env_vars[env_var_name] = value
--- a/nemo_skills/pipeline/utils/exp.py
+++ b/nemo_skills/pipeline/utils/exp.py
@@ -34,7 +34,12 @@ from nemo_skills.pipeline.utils.cluster import (
     temporary_env_update,
     tunnel_hash,
 )
-from nemo_skills.pipeline.utils.mounts import get_mounts_from_config, get_unmounted_path, is_mounted_filepath
+from nemo_skills.pipeline.utils.mounts import (
+    check_remote_mount_directories,
+    get_mounts_from_config,
+    get_unmounted_path,
+    is_mounted_filepath,
+)
 from nemo_skills.pipeline.utils.packager import (
     get_packager,
     get_registered_external_repo,
@@ -624,6 +629,13 @@ def run_exp(exp, cluster_config, sequential=False, dry_run=False):
     if dry_run:
         LOG.info("Dry run mode is enabled, not running the experiment.")
         return
+
+    if "mounts" in cluster_config:
+        # Can only check cluster mounts here, not those added to add_task
+        mounts = get_mounts_from_config(cluster_config)
+        mount_sources = [m.split(":")[0] for m in mounts]
+        check_remote_mount_directories(mount_sources, cluster_config, exit_on_failure=True)
+
     if cluster_config["executor"] != "slurm":
         exp.run(detach=False, tail_logs=True, sequential=sequential)
     else:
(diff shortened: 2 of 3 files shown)


Kipok:
@activatedgeek do you think you can help to review this?

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -197,6 +197,18 @@ def get_env_variables(cluster_config):
             if env_var_name not in _logged_optional_env_vars:
                 LOG.info(f"Optional environment variable {env_var_name} not found in user environment; skipping.")
                 _logged_optional_env_vars.add(env_var_name)
+
+    # replace placeholders with actual env var values
+    for key, value in env_vars.items():
```
activatedgeek, line 203: Perhaps, we should have a clarifying comment that these env vars are only resolved on the host submitting the job, not the host where the jobs are run?

nemo_skills/pipeline/utils/exp.py
```diff
@@ -412,6 +417,10 @@ def add_task(
         sandbox_port = get_free_port(strategy="random")
 
     env_vars = get_env_variables(cluster_config)
+    # If not set by env variable, try to resolve from config
+    if skip_hf_home_check is None:
+        skip_hf_home_check = cluster_config.get("skip_hf_home_check", False)
```
activatedgeek, line 425: Why don't we simply write,
```
skip_hf_home_check = cluster_config.get("skip_hf_home_check", skip_hf_home_check)
```

and keep the default in above arguments as `False`?

nemo_skills/pipeline/utils/mounts.py
```diff
@@ -88,9 +88,28 @@ def check_mounts(
...
+
+                # Replace ${VAR} with the value from env_vars
+                for var in env_vars:
+                    if var not in env_vars:
+                        raise ValueError(
+                            f"Cluster config item `{path}` contains env variable placeholders, but the required env var is "
+                            f"not provided in the cluster config or environment to resolve."
+                        )
+
+                    path = path.replace(f"${{{var}}}", env_vars[var])
```
activatedgeek: Doesn't `os.path.expandvars` already handle both `$VAR` and `${VAR}`? https://docs.python.org/3/library/os.path.html#os.path.expandvars

nemo_skills/pipeline/utils/mounts.py
```diff
@@ -118,6 +137,20 @@ def check_mounts(
...
+
+                    # Replace ${VAR} with the value from env_vars
+                    for var in env_vars:
+                        if var not in env_vars:
+                            raise ValueError(
+                                f"Cluster config item `{path}` contains env variable placeholders, but the required env var is "
+                                f"not provided in the cluster config or environment to resolve."
+                            )
+
+                        path = path.replace(f"${{{var}}}", env_vars[var])
```
activatedgeek: Same here. Was `os.path.expandvars` not sufficient?

activatedgeek (changes requested):

nemo_skills/pipeline/utils/cluster.py
titu1994: Done

nemo_skills/pipeline/utils/exp.py
titu1994: The function arg should override env arg. So only if function arg is unset (None), then resolve via env var next, and if that also doesnt exist, fall back to False (original default).

nemo_skills/pipeline/utils/mounts.py
titu1994: `env_vars` can be a different from os.environ. These are only those variables that are explicitly listed in cluster config, not all variables in the environment.

titu1994: I've updated the logic to use os.path.expandvars() and injected the value of `env_vars` inside `os.environ` in the `get_env_variables()` function

nemo_skills/pipeline/utils/mounts.py
titu1994: Updated

nemo_skills/pipeline/utils/mounts.py
```diff
@@ -56,6 +56,18 @@ def check_if_mounted(cluster_config, path_to_check):
         raise ValueError(f"The path '{path_to_check}' is not mounted. Check cluster config.")
 
 
+def _resolve_path_placeholders(path: str) -> str:
+    """Resolve environment variable placeholders in a path."""
+    if path is not None and "${" in path and "}" in path:
```
activatedgeek, line 61: This condition might have edge cases in case the path actually contains `${` as an escaped string, but perhaps it is very unlikely and we can let this go.

activatedgeek (approved):

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -197,6 +197,22 @@ def get_env_variables(cluster_config):
...
+            if key in os.environ:
+                env_vars[key] = os.path.expandvars(value)
+                LOG.info(
+                    f"Resolved environment variable {key} inside the placeholder value: {value} with {env_vars[key]}"
+                )
+            else:
+                raise ValueError(f"Cannot resolve environment variable {key} inside the placeholder value: {value}")
+
+    # Inject env_vars into os.environ for resolution via os.path.expandvars in other parts of the code
+    os.environ.update(env_vars)
```
Kipok: we shouldn't be doing this - we don't want to change the environment for one job based on values from another job. As in, cluster config can be changed between pipeline launches, but python session might be the same and if you write to os.environ directly it can lead to hard to catch issues when one job affects another job in unexpected ways. Can we instead pass things through env_vars dict?

nemo_skills/pipeline/utils/exp.py
```diff
@@ -412,6 +417,10 @@ def add_task(
         sandbox_port = get_free_port(strategy="random")
 
     env_vars = get_env_variables(cluster_config)
+    # If not set by env variable, try to resolve from config
```
Kipok:
```suggestion
    # If not explicitly set, resolve from cluster config
```

nemo_skills/pipeline/utils/exp.py
```diff
@@ -620,6 +629,16 @@ def run_exp(exp, cluster_config, sequential=False, dry_run=False):
...
         LOG.info("Dry run mode is enabled, not running the experiment.")
         return
+
+    if "mounts" in cluster_config:
+        # Can only check cluster mounts here, not those added to add_task
+        mounts = get_mounts_from_config(cluster_config)
+        mount_sources = [m.split(":")[0] for m in mounts]
+
+        LOG.info("Checking mount paths: %s", mount_sources)
+        exit_if_failure = not cluster_config.get("NEMO_SKILLS_DISABLE_MOUNT_CHECK", False)
```
Kipok: was this meant to be an env var? Otherwise, would be more standard and clear to call it `disable_mount_check` (lowercase and without NEMO_SKILLS as it's already config from nemo skills config)

nemo_skills/pipeline/utils/cluster.py
titu1994: Im rolling back changes for mounts support for env vars then. Not the point of this pr anyway.

nemo_skills/pipeline/utils/exp.py
titu1994: This is meant to be an env var. Let me fix. Applied the change from previous change.

PR #899, opened 2025-10-07
Control max_concurrent_requests in subclasses with parallel generations

Kipok:
Before this change, we would only guarantee that at most max_concurrent_requests calls to process_single_datapoint would run at the same time. But those sometimes do multiple parallel requests, e.g. check_contamination can do up to 10 at a time in parallel for some settings. This resulted in a very hard-to-find issue when server got hit with >5000 requests at a time and just froze for 1.5 hours without doing anything. The current fix should ensure that we control number of llm requests, not number of process_single_datapoint calls and resolves this problem.

## Summary by CodeRabbit

- New Features
  - Semaphore-controlled generation to manage concurrent LLM calls.
  - Consolidated generate-and-save flow with per-datapoint timing metadata.
  - More accurate progress updates during async runs.

- Performance
  - Improved stability and throughput under high load with predictable concurrency.

- Reliability
  - Thread-safe output writing to prevent race conditions and corrupted results.
  - Consistent single-turn generation path across tasks for more reliable behavior.

### DIFF
--- a/nemo_skills/inference/check_contamination.py
+++ b/nemo_skills/inference/check_contamination.py
@@ -20,18 +20,9 @@ from dataclasses import field
 
 import hydra
 
-from nemo_skills.inference.generate import (
-    GenerateSolutionsConfig,
-    GenerationTask,
-    InferenceConfig,
-)
+from nemo_skills.inference.generate import GenerateSolutionsConfig, GenerationTask, InferenceConfig
 from nemo_skills.inference.model import server_params
-from nemo_skills.utils import (
-    get_help_message,
-    get_logger_name,
-    nested_dataclass,
-    setup_logging,
-)
+from nemo_skills.utils import get_help_message, get_logger_name, nested_dataclass, setup_logging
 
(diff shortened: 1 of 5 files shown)


Kipok:
Can we do both? Have a hard limit of 2k requests on the model size and also a user-controlled limit in generation class? Or that is hard to implement due to how it works in asyncio?

Kipok:
Wouldn't it be safer to do both to ensure there is always a hard limit in case someone doesn't use generate_with_semaphore in subclasses? Probably not that important I guess, but might be a bit safer.

Any downside to removing the limits? Even with current limits the server fails sometimes and it's better to have the server fail than to have the job just hang, so I guess we should, but not fully sure on all tradeoffs here

Kipok:
ok, how about we merge this and then maybe you can help to do the remaining changes in a separate pr? Let me know if you have any concerns about that

smahdavi4:
the PR looks good. thanks

PR #905, opened 2025-10-07
Slurm tests for code execution timeouts

i-vainn:
This PR adds slurm tests for code execution timeouts.
The tests launch evaluation of gpt-oss-120b model with python code execution on hmmt_feb25 with 16 random seeds and num_jobs=1, calculates total number of timeouts based on `output-rs*.jsonl` and checks if it exceed 160 (on average there are 5-6 timeouts per random seed, usually resulting in 80-90 total timeouts).


Kipok (changes requested):
does this have to be hmmt instead of aime25? If so, can you please update the existing aime24/25 test to replace one with hmmt and add assert there? Best to combine timeouts and accuracy test together, so that we don't waste compute

PR #908, opened 2025-10-08
BFCLv4 support

Glorf:
## Summary by CodeRabbit

- **New Features**
  - BFCL v4 support: dataset downloader/prep pipeline, expanded splits (multi-turn, live, language-specific), memory and web-search handling.
  - New evaluation outputs: agentic (memory + web-search) and hallucination scores.
  - Built-in web-search backend with query and URL-fetch capabilities.

- **Refactor**
  - Migration of dataset, evaluation, and inference flows to BFCL v4 naming and score structure; multi-turn and memory flows enhanced.

- **Chores**
  - Updated constants, default evaluator model, ignored dataset paths, and added runtime dependencies.

### DIFF
--- a/dockerfiles/Dockerfile.nemo-skills
+++ b/dockerfiles/Dockerfile.nemo-skills
@@ -58,4 +58,3 @@ RUN pip install git+https://github.com/NVIDIA/NeMo-speech-data-processor@29b9b1e
 RUN pip install --no-cache-dir -r /opt/NeMo-Skills/requirements/main.txt
 # Fix http mismatch between lepton and dggs by manually downloading dggs here
 RUN pip install ddgs
-# Due to large nv library downloads on github actions, we need to install sentence_transformers here


dockerfiles/Dockerfile.nemo-skills
```diff
@@ -35,7 +35,7 @@ RUN git clone https://github.com/allenai/IFBench.git /opt/benchmarks/IFBench --d
 RUN cd /opt/benchmarks/IFBench && sed -i '/^unicodedata[=<>]*.*$/d' requirements.txt && pip install -r requirements.txt
 
 RUN git clone https://github.com/ShishirPatil/gorilla.git /opt/gorilla
-RUN cd /opt/gorilla && git checkout d2177992bbba9aa228b53c0645bf8f5613a5a7c6
+RUN cd /opt/gorilla && git checkout d85e9ad3a3009be3d609c5550799389a16599f88
```
shtoshni: I think we want two checkouts.. Maybe checkout in `/opt/bfcl_v3/gorilla` and `/opt/bfcl_v4/gorilla`

nemo_skills/dataset/bfcl_v3/__init__.py
```diff
@@ -15,12 +15,12 @@
 DATASET_GROUP = "tool"
 
 SPLITS = [
-    "simple",  # Simple function calls
+    "simple_python",  # Simple function calls
```
shtoshni, line 18: Can you rever the changes made to bfcl_v3?

nemo_skills/dataset/bfcl_v3/constants.py
```diff
@@ -16,13 +16,13 @@
 
 TEST_COLLECTION_MAPPING = {
     "all": [
-        "simple",
+        "simple_python",
```
shtoshni, line 22: Needs to be reverted

nemo_skills/dataset/bfcl_v3/bfcl_score.py
```diff
@@ -136,7 +133,7 @@ def calculate_non_live_single_turn_accuracy(metrics):
     return {
```
shtoshni, line 138: need s to be reverted

nemo_skills/dataset/bfcl_v3/constants.py
shtoshni: All the changes in this file I mean

nemo_skills/dataset/bfcl_v3/prepare.py
```diff
@@ -114,7 +114,7 @@ def process_file(repo_root_dir, input_file, output_file, model_type="llama-nemot
             f_out.write(json.dumps(instance) + "\n")
 
 
-def download_and_process_bfcl_data(repo_url, subfolder_path, output_dir, file_prefix="BFCL_v3", model_type="nemotron"):
+def download_and_process_bfcl_data(repo_url, subfolder_path, output_dir, file_prefix="BFCL_v4", model_type="nemotron"):
```
shtoshni: This file also needs to be reverted

nemo_skills/dataset/bfcl_v4/utils.py
```diff
@@ -0,0 +1,118 @@
+import json
```
shtoshni: Is it the same as `bfcl_v3/utils.py`? If yes, you can just refer to that file, and avoid duplication.

dockerfiles/Dockerfile.nemo-skills
Glorf: Ok, let's do that!

Glorf: Wait, actually - we cannot easily pip install both versions of bfcl_eval. I made the changes in such a way that there is a minimum set of modifications and now testing if the original BFCLv3 scores are reproducible - they should be as both the logic and the datasets didn't change (v4 is a superset of v3)

nemo_skills/dataset/bfcl_v4/utils.py
Glorf: nope, it's quite different in fact :(

requirements/main.txt
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 bs4
+ddgs # Needed for BFCLv4
```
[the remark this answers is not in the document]
Glorf: Needed for running memory kv evals

nemo_skills/inference/eval/bfcl.py
```diff
@@ -237,13 +248,16 @@ def parse_output_dict(self, output_dict: dict):
         tool_calls = [] if output_dict["message"].tool_calls is None else output_dict["message"].tool_calls
 
         try:
+            tool_calls = output_dict["message"].tool_calls
```
gwarmstrong, line 250: Is this not redundant with L248 above?

nemo_skills/dataset/bfcl_v3/constants.py
```diff
@@ -14,15 +14,18 @@
...
         "irrelevance",
         "parallel",
         "multiple",
         "parallel_multiple",
-        "java",
-        "javascript",
+        "simple_java",
```
gwarmstrong, line 27: @Glorf what is the reason for changing here and breaking backward compatibility?

nemo_skills/evaluation/evaluator/bfcl.py
```diff
@@ -39,16 +41,18 @@ def eval_bfcl(cfg):
...
+    test_category = str(Path(jsonl_file).absolute().parent.name).removeprefix("bfcl_v3.").removeprefix("bfcl_v4.")
 
     # Convert Nemo-Skills output file to BFCL format
     output_dir = Path("/opt/gorilla/berkeley-function-call-leaderboard") / f"result/{model_name}"
     score_file = (
         Path("/opt/gorilla/berkeley-function-call-leaderboard")
         / f"score/{model_name}"
-        / f"BFCL_v3_{test_category}_score.json"
+        / get_directory_structure_by_category(test_category)
+        / f"BFCL_v4_{test_category}_score.json"
```
gwarmstrong, line 61: should this have a version dependent identifier? E.g., `f"BFCL_v{version_num}_...`?

nemo_skills/inference/eval/bfcl_web_search.py
```diff
@@ -0,0 +1,263 @@
...
+
+            except DDGSException as e:
+                if "No results found" in str(e):
+                    wait_time = backoff + random.uniform(0, backoff)
+                    error_block = (
+                        "*" * 100
+                        + f"\n❗️❗️ [WebSearchAPI] Hit rate limit on DuckDuckGo requests. This is a common behaviour. If unable to run eval due to repeated rate limits, try to decrease job parallelism. Retrying in {wait_time:.1f} seconds…"
+                        + "*" * 100
+                    )
+                    print(error_block)
```
gwarmstrong, line 160: Is this copied from somewhere, or did we write this? If we wrote it, can we using `logging` here? It will make this to a little more reusable. (e.g., better integration with MCPs with Stdio transport). Similar throughout this file.

nemo_skills/evaluation/evaluator/bfcl.py
gwarmstrong, line 41: Is there a way to evaluate an individual sample at a time? If so, it would definitely be good to move onto the new `Evaluator` class with an `eval_single` implementation: https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/evaluation/evaluator/base.py#L34-L91

This would let samples be evaluated during the generation loop--potentially saving time if the evaluations are CPU intensive.

Otherwise, it would still be good to move onto that, but only implement `eval_full`.

gwarmstrong (commented):
@Glorf I have a couple questions about the current implementation.

nemo_skills/dataset/bfcl_v3/constants.py
Glorf: The authors renamed the evals in the repo: https://github.com/ShishirPatil/gorilla/commit/58f57e9124ea981403792dd51e00a6577e621fae
I originally tried to put bfclv3 aside from bfclv4, but as they both implement bfcl_eval package, I decided to keep bfclv3 as the subset of bfclv4 and use bfclv4 evaluation/logic (as it's technically the same, just need to drop agentic evals)

nemo_skills/evaluation/evaluator/bfcl.py
Glorf: As above, the bfcl_eval expects it to be BFCL_v4, but we evaluate subset of v4 as v3

nemo_skills/inference/eval/bfcl_web_search.py
Glorf: This is this: https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl_eval/eval_checker/multi_turn_eval/func_source_code/web_search.py but with the alteration so that it uses DDGS (public access to API) instead of SERPApi (paid API service)

nemo_skills/inference/eval/bfcl.py
Glorf: Good catch, will drop it :)

nemo_skills/evaluation/evaluator/bfcl.py
Glorf: I don't think we can - at least not at this control level - we offload all the eval logic to the bfcl_eval - the original repo

requirements/main.txt
```diff
@@ -15,7 +15,9 @@
 bs4
 click < 8.2.0  # https://github.com/ai-dynamo/dynamo/issues/1039
 datasets<4  # lcb problem with datasets 4.0.0
+ddgs # Needed for BFCLv4
```
gwarmstrong: this requirement is causing build issues--the dockerfile cannot be built with it.

Glorf: This seems to need httpx >=0.28 while the nemo_skills also relies on lepton package - which seems to be abandoned and fixed at 0.27. I checked their codebase and the reason for fixing on 0.27 was just fixing some CI. I will therefore install ddgs manually in-container to overwrite httpx version

dockerfiles/Dockerfile.nemo-skills
```diff
@@ -55,4 +55,7 @@ COPY pyproject.toml README.md /opt/NeMo-Skills/
 COPY requirements /opt/NeMo-Skills/requirements/
 # installing sdp in container only
 RUN pip install git+https://github.com/NVIDIA/NeMo-speech-data-processor@29b9b1ec0ceaf3ffa441c1d01297371b3f8e11d2
-RUN pip install -r /opt/NeMo-Skills/requirements/main.txt
+RUN pip install --no-cache-dir -r /opt/NeMo-Skills/requirements/main.txt
+# Fix http mismatch between lepton and dggs by manually downloading dggs here
+RUN pip install ddgs
+# Due to large nv library downloads on github actions, we need to install sentence_transformers here
```
gwarmstrong:
```suggestion
```

gwarmstrong (approved):
LGTM

PR #919, opened 2025-10-09
Update IOI to support interleaving + small improvements

SeanNaren:
- Adds interleaving support for IOI
- Expose UWSGI parameters that can be set in the dataset init.py file
- Update dataset metadata to reduce bloat in the main JSONL file
- Added additional summarization logic (per problem/per subtask breakdown)

## Summary by CodeRabbit

- New Features
  - IOI25 dataset configuration and class-based IOI evaluator added; benchmarks can provide per-benchmark sandbox environment overrides.

- Improvements
  - Safer, environment-aware sandbox defaults with a stateful sandbox option and echoed effective UWSGI settings; sandbox-related vars now propagate into run jobs.
  - Evaluation now tracks, prints, and returns per-problem and per-subtask scores.

- Dataset
  - IOI24 metadata: "score" renamed to "subtask_score" and workflow fields moved to subtask level.

- Refactor
  - IOI evaluation reworked into a sandboxed, parallel, class-based flow.

### DIFF
--- a/nemo_skills/prompt/config/generic/genselect.yaml
+++ b/nemo_skills/prompt/config/generic/genselect.yaml
@@ -1,6 +1,6 @@
 
 user: |-
-  You will be given a problem followed by an enumerated list of {num_solutions} candidate solutions. Your task is to systematically analyze these solutions to identify the best approach.
+  You will be given a problem followed by an enumerated list of {num_solutions} candidate solutions. Your task is to systematically analyze these solutions to identify the best solution among them.
 
   Problem: {problem}
 
@@ -10,3 +10,7 @@ user: |-
   End your evaluation with exactly:
   Judgment: [IDX]
   where IDX is the index 0-{max_idx} of the best solution.
+
+  Example: If Solution 2 (index 2) is best, write: "Judgment: 2" at the end.
+
+  Important: Use only the format "Judgment: 2" with no special formatting, quotes, or additional characters.


dockerfiles/sandbox/start-with-nginx.sh
```diff
@@ -12,12 +12,18 @@ echo "Workers: $NUM_WORKERS, Nginx port: $NGINX_PORT"
...
-export UWSGI_PROCESSES
-export UWSGI_CHEAPER
-echo "Forced UWSGI settings for session affinity: PROCESSES=$UWSGI_PROCESSES, CHEAPER=$UWSGI_CHEAPER"
+# Respect pre-set environment variables; fall back to sane defaults
+# If the caller exported UWSGI_PROCESSES / UWSGI_CHEAPER (or the helper
+# variables used by start_local_sandbox.sh), keep those.  Otherwise pick
+# defaults that work for most cases.
+
+# shell parameter expansion with default value
+: "${UWSGI_PROCESSES:=2}"   # default to 2 processes per worker if unset
```
Kipok: @gwarmstrong any problem for ipython session affinity with this default?

nemo_skills/dataset/ioi25/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: is there a corresponding prepare.py?

nemo_skills/pipeline/utils/exp.py
```diff
@@ -516,6 +516,9 @@ def add_task(
             "LISTEN_PORT": sandbox_port,
             "NGINX_PORT": sandbox_port,
         }
+        for override in cluster_config.get("module_env_vars", []):
+            key, value = override.split("=", 1)
+            sandbox_env_updates.setdefault(key, value)
```
Kipok: if these are just for the sandbox, let's rename ENV_VARS to SANDBOX_ENV_VARS ?

nemo_skills/pipeline/utils/eval.py
```diff
@@ -185,6 +185,11 @@ def get_benchmark_args_from_module(
         benchmark_module, "KEEP_MOUNTS_FOR_SANDBOX", False, override_dict
     )
 
+    # Add any benchmark-specific environment variables to cluster config
+    env_vars_from_module = getattr(benchmark_module, "ENV_VARS", getattr(benchmark_module, "env_vars", []))
+    if env_vars_from_module:
+        cluster_config["module_env_vars"] = list(env_vars_from_module)
```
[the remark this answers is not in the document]
Kipok: @SeanNaren the rabbit is right, we should ideally have a way to clean this up, otherwise it could lead to hard-to-catch mistakes. Best not to carry this through cluster config but return directly, that way it shouldn't be shared across benchmarks

nemo_skills/dataset/ioi25/__init__.py
SeanNaren: Not yet, we are working on this!

nemo_skills/pipeline/utils/eval.py
```diff
@@ -517,6 +524,16 @@ def prepare_eval_commands(
                     job_needs_sandbox_to_keep_mounts = any(
                         benchmarks_dict[b].keep_mounts_for_sandbox for b in job_benchmarks
                     )
+                    # Aggregate per-job sandbox env overrides from participating benchmarks (first key wins)
```
Kipok, line 528: we don't want to do this. Instead we want to fail if they don't match as otherwise it can lead to silently broken setup for some benchmarks. Users can always submit them as separate jobs or with higher num_jobs

dockerfiles/sandbox/start-with-nginx.sh
Kipok: Can we keep default to 1 if you're anyway overriding this in your benchmarks?

nemo_skills/dataset/ioi25/__init__.py
Kipok: Can you add a comment about this then here, so that users aren't confused?

nemo_skills/dataset/ioi24/__init__.py
```diff
@@ -17,3 +17,6 @@
 DATASET_GROUP = "code"
 METRICS_TYPE = "ioi"
 EVAL_ARGS = "++eval_type=ioi"
+
+# environment variables required by this benchmark
+ENV_VARS = ["UWSGI_PROCESSES=1024", "UWSGI_CPU_AFFINITY=8", "UWSGI_CHEAPER=1023", "NUM_WORKERS=1"]
```
Kipok: Rename to SANDBOX_ENV_VARS ?

dockerfiles/sandbox/start-with-nginx.sh
gwarmstrong: yeah changing this would cause problems for session affinity, agree with @Kipok

nemo_skills/pipeline/utils/eval.py
```diff
@@ -517,6 +525,23 @@ def prepare_eval_commands(
...
+                    env_source: Dict[str, str] = {}
+                    for b in ordered_benchmarks:
+                        for override in benchmarks_dict[b].sandbox_env_overrides:
+                            key, value = override.split("=", 1)
+                            if key in env_map and env_map[key] != value:
+                                raise ValueError(
+                                    "Conflicting sandbox environment overrides for key '{key}': "
+                                    f"'{env_map[key]}' (from {env_source[key]}) vs '{value}' (from {b}). "
+                                    "Please submit the benchmarks as separate jobs or increase num_jobs so they do not share a job."
+                                )
```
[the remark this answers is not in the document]
Kipok: @SeanNaren ^

nemo_skills/dataset/ioi24/__init__.py
Kipok:
```suggestion
SANDBOX_ENV_VARS = ["UWSGI_PROCESSES=1024", "UWSGI_CPU_AFFINITY=8", "UWSGI_CHEAPER=1023", "NUM_WORKERS=1"]
```

PR #926, opened 2025-10-09
Fix gpt-5 azure 

smahdavi4:
Fixing gpt-5 calling for openai azure models.

## Summary by CodeRabbit

- New Features
  - GPT-5 models are now automatically treated as reasoning models for enhanced reasoning behavior.
  - reasoning_effort can be provided and is honored for reasoning models.

- Bug Fixes
  - Improved API key detection for Azure OpenAI configurations, reducing failures due to key resolution in mixed OpenAI/Azure environments.

### DIFF
--- a/nemo_skills/inference/model/azure.py
+++ b/nemo_skills/inference/model/azure.py
@@ -30,7 +30,8 @@ class AzureOpenAIModel(OpenAIModel):
         self.litellm_kwargs["api_version"] = api_version
 
     def _get_api_key(self, api_key: str | None, api_key_env_var: str | None, base_url: str) -> str | None:
-        api_key = super()._get_api_key(api_key, api_key_env_var, base_url)
+        # OPENAI_API_KEY is used by non-Azure OpenAI models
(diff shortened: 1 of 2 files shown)


nemo_skills/inference/model/azure.py
```diff
@@ -30,7 +30,8 @@ def __init__(
         self.litellm_kwargs["api_version"] = api_version
 
     def _get_api_key(self, api_key: str | None, api_key_env_var: str | None, base_url: str) -> str | None:
-        api_key = super()._get_api_key(api_key, api_key_env_var, base_url)
+        # OPENAI_API_KEY is used by non-Azure OpenAI models
+        api_key = super(OpenAIModel, self)._get_api_key(api_key, api_key_env_var, base_url)
```
Kipok, line 34: won't this break other models by defaulting to OPENAI_API_KEY instead of AZURE? As in, we probably want to take AZURE one by default if it's defined unless perhaps it's gpt-5 or whichever models need exception?

PR #928, opened 2025-10-10
fix gemini calling

smahdavi4:
## Summary by CodeRabbit

* **Bug Fixes**
  * Avoids unnecessary startup waits by treating a missing or explicit "None" server address as not provided, preventing hang-ups when no server is configured.
  * Fixes model initialization when the base URL is empty so the model reliably recognizes "no base URL" and avoids stale or incorrect connection state.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -653,6 +653,9 @@ class GenerationTask:
         self.cleanup_litellm_cache()
 
     def wait_for_server(self):
+        if not self.cfg.server.get("base_url") and not self.cfg.server.get("host") and not self.cfg.server.get("port"):
+            LOG.info("Skipping server wait as no server address is provided.")
+            return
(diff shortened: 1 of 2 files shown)


nemo_skills/inference/generate.py
```diff
@@ -664,7 +654,8 @@ def restore_async_order(self):
 
     def wait_for_server(self):
         server_address = self.cfg.server.get("base_url") or f"{self.cfg.server['host']}:{self.cfg.server['port']}"
-        if not server_address:
+        # Hydra sets None parameters to "None" string
+        if server_address == "None":
```
Kipok, line 661: maybe have both conditions? I think it only sets the to "None" if you do `++server_address=None` but if you do `++server_address=null` it will be correctly set to `None` python value

nemo_skills/inference/model/base.py
```diff
@@ -117,6 +117,7 @@ def __init__(
         elif base_url == "":
             # We don't want to use base_url if it is an empty string
             base_url = None
```
Kipok: I guess this line can be removed then?

PR #933, opened 2025-10-12
add support for checkpoint average

wedu-nvidia:
Add support for checkpoint average
https://github.com/NVIDIA-NeMo/Skills/issues/890

## Summary by CodeRabbit

* **New Features**
  * Built-in checkpoint averaging to merge multiple safetensors checkpoints into a single final model.
  * New CLI options: average_steps (comma-separated) and remove_checkpoints_after_average for optional cleanup.

* **Improvements**
  * Shard-aware averaging with automatic index handling, metadata recording, auxiliary-file propagation, strict shape/dtype checks, and per-shard progress logs.
  * Backend-aware preprocessing and conversion support for different checkpoint formats.

* **Behavior**
  * Per-step conversion tasks can be orchestrated into a final averaging step; original step directories may be removed if requested.

* **Configuration**
  * Added GRPO config keys (epochs, filtering, async settings) and Megatron/generation tuning flags; logger toggle for swanlab.

### DIFF
--- a/nemo_skills/pipeline/nemo_rl/sft.py
+++ b/nemo_skills/pipeline/nemo_rl/sft.py
@@ -243,7 +243,7 @@ def sft_nemo_rl(
         help="List of commas separated checkpoint steps to average. E.g '1000,2000,3000,4000,5000'. If None, skip average step and only convert last checkpoint",
     ),
     remove_checkpoints_after_average: bool = typer.Option(
-        False, help="Whether to delete original step directories after averaging (default: True)."
+        False, help="Whether to delete original step directories after averaging (default: False)."
     ),
     wandb_project: str = typer.Option("nemo-skills", help="Weights & Biases project name"),
     wandb_group: str = typer.Option(None, help="Weights & Biases group name."),
--- a/nemo_skills/pipeline/nemo_rl/grpo.py
+++ b/nemo_skills/pipeline/nemo_rl/grpo.py
@@ -246,7 +246,7 @@ def grpo_nemo_rl(
         help="List of commas separated checkpoint steps to average. E.g '1000,2000,3000,4000,5000'. If None, skip average step and only convert last checkpoint",
     ),
     remove_checkpoints_after_average: bool = typer.Option(
-        False, help="Whether to delete original step directories after averaging (default: True)."
+        False, help="Whether to delete original step directories after averaging (default: False)."
     ),
     wandb_project: str = typer.Option("nemo-skills", help="Weights & Biases project name"),
     wandb_group: str = typer.Option(None, help="Weights & Biases group name."),


nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -184,6 +184,27 @@ def get_checkpoint_convert_cmd(output_dir, final_hf_path, step, backend):
     return cmd
 
 
+def get_checkpoint_average_cmd(output_dir, average_steps, backend, cleanup):
```
Kipok: Let's add this to grpo as well?

nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -184,6 +184,27 @@ def get_checkpoint_convert_cmd(output_dir, final_hf_path, step, backend):
...
+    cmd = "export PYTHONPATH=$PYTHONPATH:/nemo_run/code && export UV_PROJECT=/opt/NeMo-RL && cd /nemo_run/code && "
+
+    if backend == "fsdp":
+        raise ValueError("Only support 'megatron' backend now")
```
Kipok: Should we do it on the level of HF checkpoints? So first, we convert checkpoint to HF and then average weights. This way it will work for both backends (or any other new backend in the future)?

nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -207,6 +228,13 @@ def sft_nemo_rl(
     num_gpus: int = typer.Option(..., help="Number of GPUs"),
     num_training_jobs: int = typer.Option(1, help="Number of training jobs"),
     conversion_step: int = typer.Option(None, help="The step of checkpoint that needs to be converted"),
+    average_steps: str = typer.Option(
+        None,
+        help="List of commas separated checkpoint steps to average. E.g '1000,2000,3000,4000,5000'. If None, skip average step and only convert last checkpoint",
```
Kipok, line 243:
```suggestion
        help="List of commas separated checkpoint steps to average. E.g '1000,2000,3000,4000,5000'. If None, skip average step and only convert checkpoint from `conversion_step`",
```

Kipok: and maybe mention in conversion_step doc that it's last by default. Or maybe we explicitly use "last" as the value to make it clearer

nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -207,6 +228,13 @@ def sft_nemo_rl(
...
+    ),
+    cleanup: bool = typer.Option(
```
Kipok: why do we need this? I think we should just preserve all checkpoints and users can cleanup on their side?

nemo_skills/pipeline/nemo_rl/sft.py
Kipok: I guess this is already how it's done? So then why isn't fsdp supported?

wedu-nvidia: Currently, I did not test for fsdp, as after hf conversion, it provides a complete file called pytorch_model.bin. 
I can give a test for this.

nemo_skills/pipeline/nemo_rl/sft.py
wedu-nvidia: Sure

nemo_skills/pipeline/nemo_rl/sft.py
wedu-nvidia: I can removed this flag, or set cleanup=false, and user can remove later, which way do you suggest.

nemo_skills/pipeline/nemo_rl/sft.py
wedu-nvidia: ok, sure.

nemo_skills/pipeline/nemo_rl/sft.py
Kipok: let's rename it to remove_checkpoints_after_average and set to False by default

Kipok (approved):

PR #939, opened 2025-10-13
Queue sandbox jobs with client polling to avoid overload timeouts

i-vainn:
This PR queues sandbox jobs for sequential execution and adds client-side polling to avoid HTTP timeouts, with session affinity tests updated to match the new server API.

## Summary by CodeRabbit

* **New Features**
  * Executions may be queued: submit returns a job ID, queue position, and worker info; view job status, cancel jobs, and trigger a worker reset.
  * New session management endpoints: list and delete active sessions; jobs tied to sessions for affinity and related-job cancellation.

* **Bug Fixes**
  * Improved session restoration, cleanup, timeout handling, and clearer truncated output indicators.

* **Tests**
  * Added tests for 202-accepted polling, timeouts, and session-affinity behavior.


nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -605,35 +717,30 @@ def execute_shell(command, timeout):
...
-        result["stderr"] = result["stderr"][:max_output_characters] + "<output cut>"
-
-    return result
+    request_dict = request.json
+    request_dict["session_id"] = session_id
+    job_id = job_manager.submit(request_dict)
+    with job_manager.lock:
+        queued_ahead = sum(
+            1 for j in job_manager.jobs.values() if j.job_id != job_id and j.status in ("queued", "running")
+        )
```
gwarmstrong: Can you write this as a method on the `job_manager`? I would recommend against using `job_manager.lock` outside of the `job_manager` methods. It can get confusing to debug later if there is an issue with the locking logic and not all of the methods that use the same lock are in the same place.

gwarmstrong: At the very least to get the id's/statuses queued ahead. Feel free to do the 1/0 and sum ops here.

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
+        j = output.json()
+        if "job_id" not in j:
+            return self._parse_request_output(output)
+        url = f"http://{self.host}:{self.port}/jobs/{j['job_id']}"
+        # Estimate the deadline based on the queued ahead jobs; cut it off at 1200 seconds
+        deadline_seconds = min(1200, (float(j.get("queued_ahead", 0)) + 1) * (timeout + 2))
+        deadline = time.monotonic() + deadline_seconds
+        while True:
+            if time.monotonic() > deadline:
+                raise httpx.TimeoutException("Client timed out")
```
gwarmstrong: wouldn't we also want to cancel the job since we will not have a chance of getting it's result? If we don't, this issue would probably just compound.

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
+                import sshtunnel_requests
+
+                def s_get():
+                    return sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path).get(
+                        url, timeout=timeout, headers=extra_headers
+                    )
+
+                resp = await asyncio.to_thread(s_get)
+            else:
+                resp = await self.http_session.get(url, timeout=timeout, headers=extra_headers)
```
gwarmstrong: Ideally it wouldn't end up being relevant, but I would make a much shorter timeout for this. With all the extra locking code on the server now, it would be better to have this be a very short-lived call. If this call starts timing out, it's a server issue and will need to be debugged differently.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -41,6 +45,101 @@
...
+        with self.lock:
+            self.jobs[job_id] = job
+            # schedule serially via single-worker executor
+            future = self.executor.submit(self._run_job, job_id)
+            self.futures[job_id] = future
+        return job_id
+
+    def get_job(self, job_id: str):
+        with self.lock:
+            return self.jobs.get(job_id)
```
gwarmstrong: you probably want to distinguish whether it is a finished job or a queued/incomplete job. If finished, it should probably be evicted after it is returned in order to save memory. This will also mean finished jobs can only be "gotten" once.

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
             output = await asyncio.to_thread(ssh_request)
         else:
             output = await self.http_session.post(
-                url=self._get_execute_url(),
+                self._get_execute_url(),
                 content=json.dumps(request),
                 timeout=timeout + 5.0,
```
gwarmstrong: remove this timeout pad--also reduce this timeout as mentioned below. This operation should be fast.

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
-        # retrying 502 errors
-        if output.status_code == 502:
-            raise httpx.TimeoutException("502 error")
-        return self._parse_request_output(output)
+        j = output.json()
+        if "job_id" not in j:
+            return self._parse_request_output(output)
+        url = f"http://{self.host}:{self.port}/jobs/{j['job_id']}"
+        # Estimate the deadline based on the queued ahead jobs; cut it off at 1200 seconds
+        deadline_seconds = min(1200, (float(j.get("queued_ahead", 0)) + 1) * (timeout + 2))
```
gwarmstrong: This basically shouldn't happen unless timeout * queued_ahead > 1200, right? Otherwise, there is some issue on the sandbox side. 

If that happens, it might be better to back off and take some steps to remediate the sandbox, rather than continue to send requests and kill them, yeah? I don't see anything here that would help a worker recover from this state.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -41,6 +45,101 @@
...
+    status: str = "queued"
+    created_at: float = field(default_factory=time.time)
+    started_at: Optional[float] = None
+    finished_at: Optional[float] = None
+    result: Optional[Dict[str, Any]] = None
+
+
+class JobManager:
+    def __init__(self, shell_manager):
+        self.jobs = {}
```
gwarmstrong, line 61: A potential issue with keeping this all in memory, is if the worker stops responding or crashes for some reason, all of your state disappears with the worker. Might not be something we need to fix right now, but if we start having issues of workers stopping/crashing, this will be a crux.

gwarmstrong: It's not clear to me if we need something more robust yet, though.

nemo_skills/code_execution/local_sandbox/local_sandbox_server.py
```diff
@@ -605,35 +717,30 @@ def execute_shell(command, timeout):
...
-
-    return result
+    request_dict = request.json
+    request_dict["session_id"] = session_id
+    job_id = job_manager.submit(request_dict)
+    with job_manager.lock:
+        queued_ahead = sum(
+            1 for j in job_manager.jobs.values() if j.job_id != job_id and j.status in ("queued", "running")
+        )
+    return {"job_id": job_id, "queued_ahead": queued_ahead}, 202
```
gwarmstrong: maybe also return `"pid": os.getpid()` or log it here alongside the job_id

It will be useful to us later if we can determine whether timeouts are occurring all on a single worker (e.g., worker died, unresponsive), or if they are distributed across the workers.

nemo_skills/code_execution/sandbox.py
gwarmstrong: I would also log the session_id here

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
+            else:
+                resp = await self.http_session.get(url, timeout=timeout, headers=extra_headers)
+            if getattr(resp, "status_code", 200) != 200:
+                await asyncio.sleep(0.2)
+                continue
+            p = resp.json()
+            if p.get("status") in {"completed", "timeout", "error", "failed", "canceled"}:
+                r = p.get("result") or {}
+                return r if "process_status" in r else {**r, "process_status": p.get("status", "error")}
+            await asyncio.sleep(0.2)
```
gwarmstrong: log session id? maybe 0.2 is too often to log that though, if you want to log it every so often? Will be helpful for tracing timeouts.

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
+
+                resp = await asyncio.to_thread(s_get)
+            else:
+                resp = await self.http_session.get(url, timeout=timeout, headers=extra_headers)
+            if getattr(resp, "status_code", 200) != 200:
+                await asyncio.sleep(0.2)
+                continue
+            p = resp.json()
+            if p.get("status") in {"completed", "timeout", "error", "failed", "canceled"}:
+                r = p.get("result") or {}
```
gwarmstrong: log session id? maybe 0.2 is too often to log that though, if you want to log it every so often? Will be helpful for tracing timeouts.

also, if you follow suggestion to return pid, log and pop it here alongside the session id

nemo_skills/code_execution/sandbox.py
```diff
@@ -102,15 +109,40 @@ def ssh_request():
...
+
+                def s_get():
+                    return sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path).get(
+                        url, timeout=timeout, headers=extra_headers
+                    )
+
+                resp = await asyncio.to_thread(s_get)
+            else:
+                resp = await self.http_session.get(url, timeout=timeout, headers=extra_headers)
+            if getattr(resp, "status_code", 200) != 200:
```
gwarmstrong, line 144: this shouldn't really happen, right? unless there is some kind of intermittent error.

Let's log the error so that we could actually verify it's intermittent, because if you get a 404 I don't see how that would be resolved by looping--you would just get stuck here and it wouldn't even stop if you time out, so as far as I can tell you would basically be freezing a whole worker.

gwarmstrong (commented):
various comments and questions--some to verify intended behavior, in other places I think some additional logging will be helpful for future debugging.

nemo_skills/code_execution/sandbox.py
i-vainn: Good point. I’d prefer to handle that on the server side, though, so it goes into `sandbox.log` instead of `main.log`. That way, we keep the main logs clean and avoid excessive noise from frequent polling.

gwarmstrong: yeah I mean either way I would probably increase the time between the status poll logging, and probably also log when you update the state

nemo_skills/code_execution/sandbox.py
i-vainn: I don't know what is the best way to remediate the sandbox in this case. For now, I've implemented worker hard reset and explicitly logging error.

nemo_skills/code_execution/sandbox.py
i-vainn: Good point. Let's raise an error for now and see if this ever happens.

nemo_skills/code_execution/sandbox.py
i-vainn: Added logging on state update. Not sure if we want to log each polling request though

nemo_skills/code_execution/sandbox.py
gwarmstrong: the hard reset you've implemented is of the shell, right? I was also concerned here that you might see issues at the API layer, which in the current setup, you cannot kill because that layer is storing all your queued jobs, and you would basically delete all pending jobs if that layer crashed.

gwarmstrong (approved):
@i-vainn Appears to address the major functional concerns from before. I have some lingering concerns about robustness of the overall system (e.g., if workers start failing), but it's probably best to see what happens in practice. Happy to help debug any sessions you find that are having failures with the sandbox interaction.

Kipok:
@gwarmstrong @i-vainn what's our plan for this pr?

Kipok:
given this doesn't look to be particularly important, should we just close it? Or test more and merge?

gwarmstrong:
@Kipok my vote would be to close it

PR #942, opened 2025-10-13
OpenReasoning Recipe Update - GenSelect

shtoshni:
## Summary by CodeRabbit

* **New Features**
  * Integrated genselect into the evaluation flow to run parallel-thinking generations within the same eval command.

* **Refactor**
  * Standardized model size labels to uppercase (e.g., 1.5B, 7B, 14B, 32B).
  * Updated model source paths to centralized registry-style names.
  * Simplified genselect output directories to a single consistent layout.
  * Consolidated benchmark configuration into a single benchmarks argument and streamlined eval context construction.
* **Bug Fixes**
  * Corrected judge model reference to the registry-style path.

### DIFF
--- a/dockerfiles/sandbox/start-with-nginx.sh
+++ b/dockerfiles/sandbox/start-with-nginx.sh
@@ -12,12 +12,21 @@ echo "Workers: $NUM_WORKERS, Nginx port: $NGINX_PORT"
 # Override nginx config for multi-worker mode (single mode uses original config)
 echo "Configuring nginx for multi-worker load balancing..."
 
-# Force session affinity settings: 1 process per worker with minimal cheaper
-UWSGI_PROCESSES=1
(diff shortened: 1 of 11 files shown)


recipes/openreasoning/eval.py
```diff
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nemo_skills.pipeline.cli import eval, genselect, wrap_arguments
+from nemo_skills.pipeline.cli import eval, wrap_arguments
 
 size_to_eval_gpus = {
-    "1.5b": 1,
```
Kipok, line 18: Should we also change to reference models not from /workspace but from HF?

shtoshni: sure thing, let me edit that.

PR #951, opened 2025-10-15
Add Lepton integration

artbataev:


Kipok:
will reopen if we have a use-case

PR #952, opened 2025-10-15
Fix AA-LCR evaluation

hsiehjackson:
## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluation now requires both an answer and its accompanying reasoning to be present before marking a judgment correct, avoiding false positives from empty outputs.

* **Behavior Change**
  * Scoring is stricter for incomplete predictions; some items previously marked correct may now be labeled incorrect.

* **New Data**
  * Source items now preserve the original question text used in prompts for clearer context.

* **Notes**
  * Prompt references updated to use the preserved original question.

### DIFF
--- a/nemo_skills/evaluation/metrics/aalcr_metrics.py
+++ b/nemo_skills/evaluation/metrics/aalcr_metrics.py
@@ -61,7 +61,7 @@ class AALCRMetrics(BaseMetrics):
 
         # Primary evaluation method: LLM-based equality checker
         if "judgement" in prediction:
-             # Invalid generation: reasoning in not finished or non-reasoning generation is empty
+             # Invalid generation: reasoning is not finished or non-reasoning generation is empty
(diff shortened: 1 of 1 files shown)


nemo_skills/evaluation/metrics/aalcr_metrics.py
```diff
@@ -61,10 +61,9 @@ def _get_score_dict(self, prediction: dict) -> dict[str, bool | int | float]:
 
         # Primary evaluation method: LLM-based equality checker
         if "judgement" in prediction:
-            is_valid_generation = len(prediction["generation"].strip()) > 0
-            correctness_dict["judge_correct"] = (
-                self.is_aalcr_correct(prediction["judgement"]) if is_valid_generation else False
-            )
+             # Invalid generation: reasoning in not finished or non-reasoning generation is empty
```
fayejf: typo  # Invalid generation: reasoning is not finished or non-reasoning generation is empty

PR #953, opened 2025-10-15
RULER remove answer prefix

hsiehjackson:
## Summary by CodeRabbit

- New Features
  - None

- Refactor
  - Simplified preparation by removing generation-related options from settings and output.
  - Question text now appends the answer prefix; the generation field is no longer emitted.
  - Max sequence length now accounts for template tokens earlier for more accurate limits.

- Chores
  - Reduced samples per prepare task from 500 to 100 to speed up data preparation.
  - Stopped injecting token count defaults into per-task settings.


hsiehjackson:
Yes this is the same as https://github.com/NVIDIA-NeMo/Skills/pull/1147. Let me close this request.

PR #955, opened 2025-10-16
update config to make both cp + sp work for fsdp backend

wedu-nvidia:
## Summary by CodeRabbit

* **Refactor**
  * Sequence-length divisibility now shifts from a static reference to a runtime-computed value, improving packing alignment across training modes when parallel sizes exceed 1.

* **New Features**
  * Adds a runtime utility that computes the required sequence-length divisibility (uses LCM-like logic when multiple parallels apply, otherwise sensible fallbacks).

* **Chores**
  * Configuration wiring updated so the computed divisibility is applied consistently before configs are finalized.

### DIFF
--- a/nemo_skills/training/nemo_rl/configs/grpo.yaml
+++ b/nemo_skills/training/nemo_rl/configs/grpo.yaml
@@ -87,6 +87,7 @@ policy:
     logprob_mb_tokens: ${mul:${policy.max_total_sequence_length}, ${policy.logprob_batch_size}}
     algorithm: "modified_first_fit_decreasing"
     sequence_length_round: 64
+  #If this value is set to null, it will be automatically assigned in the code.
   make_sequence_length_divisible_by: null
   max_grad_norm: 1.0
 
--- a/nemo_skills/training/nemo_rl/configs/sft.yaml
+++ b/nemo_skills/training/nemo_rl/configs/sft.yaml
@@ -129,6 +129,7 @@ policy:
     algorithm: "modified_first_fit_decreasing"
     sequence_length_round: 64
     logprob_mb_tokens: ${mul:${policy.max_total_sequence_length}, ${policy.train_micro_batch_size}}
+  #If this value is set to null, it will be automatically assigned in the code.
   make_sequence_length_divisible_by: null
   max_grad_norm: null
 


nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -90,7 +90,7 @@ policy:
 
   # makes the training sequence length divisible by the tensor parallel size
   # this is useful for sequence parallel training
-  make_sequence_length_divisible_by: ${policy.dtensor_cfg.tensor_parallel_size}
+  make_sequence_length_divisible_by: ${mul:${policy.context_parallel_size}, 2}
```
Kipok: so tp -> 2*cp? If that's the right change, can you update the comment above?

wedu-nvidia: Yes, I see an error when running fsdp and got the suggestion as above. 
The most safe way should be

> Ideally, this should be smallest common multiple of ${mul:${policy.dtensor_cfg.context_parallel_size}, 2}  and ${policy.dtensor_cfg.context_parallel_size} .That is safest way to make both CP and SP work. 

I think i can add like this, let me push another change.

wedu-nvidia: @Kipok Just tested on cluster and updated.

nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -87,10 +87,7 @@ policy:
     logprob_mb_tokens: ${mul:${policy.max_total_sequence_length}, ${policy.logprob_batch_size}}
     algorithm: "modified_first_fit_decreasing"
     sequence_length_round: 64
-
-  # makes the training sequence length divisible by the tensor parallel size
-  # this is useful for sequence parallel training
-  make_sequence_length_divisible_by: ${policy.dtensor_cfg.tensor_parallel_size}
+  make_sequence_length_divisible_by: 1
```
Kipok: the current logic doesn't allow the user to override our default. Can we instead set this to null and only override automatically if it's None? That way users can still explicitly set it if they want to use a different value for whatever reason

wedu-nvidia: Sure, the parameter will be initialized with a default value of None and subsequently overridden if a value is provided. If the user chooses to set this parameter, they assume responsibility for providing a valid and appropriate value

nemo_skills/training/nemo_rl/configs/grpo.yaml
```diff
@@ -87,7 +87,7 @@ policy:
     logprob_mb_tokens: ${mul:${policy.max_total_sequence_length}, ${policy.logprob_batch_size}}
     algorithm: "modified_first_fit_decreasing"
     sequence_length_round: 64
-  make_sequence_length_divisible_by: 1
+  make_sequence_length_divisible_by: null
```
Kipok, line 91: thanks, and can you please also add a comment here explaining that it's automatically set in the code if null?

wedu-nvidia: sure.

Kipok (approved):

PR #965, opened 2025-10-17
Customize `num_tasks` per node in `run_cmd`

artbataev:
Allow customizing `num_tasks` (num tasks per node) in `run_cmd`.
Default current value is preserved.

## Release Notes

* **New Features**
  * Added configurable parameter to specify the number of tasks per node (default: 1).

* **Documentation**
  * Clarified GPU allocation option to reflect per-node configuration rather than total allocation.


nemo_skills/pipeline/run_cmd.py
```diff
@@ -54,8 +54,9 @@ def run_cmd(
     ),
     qos: str = typer.Option(None, help="Specify Slurm QoS, e.g. to request interactive nodes"),
     time_min: str = typer.Option(None, help="If specified, will use as a time-min slurm parameter"),
-    num_gpus: int | None = typer.Option(None, help="Number of GPUs to use"),
+    num_gpus: int | None = typer.Option(None, help="Number of GPUs per node to use"),
```
Kipok, line 57: might be a good idea to update message in all pipelines, not just here

PR #973, opened 2025-10-18
Update omr receipe

wedu-nvidia:
https://github.com/NVIDIA-NeMo/Skills/issues/963

## Summary by CodeRabbit

* **New Features**
  * Added a --backend CLI option (default: megatron) to run training and evaluation on one or multiple backends.
  * Pipeline now runs per-backend training and per-backend initial/final evaluations, producing backend-specific models and result outputs.

* **Tests**
  * Test and verification flows updated to iterate backends, pass the backend option through, and validate backend-specific outputs.


wedu-nvidia:
@Kipok I tested the full loop and the results are still in the range, so no need to update the range.

Kipok (changes requested):
so the current version repeats the whole script for different backend. This is how it was before, but it's inefficient because we don't need to repeat baseline evals / generation two times as they are identical. Could you please change it such that instead it uses backend as a list, so by default it's running only megatron, but we can configure it to run both megatron and fsdp inside the same job in parallel and then do two final evals for both of them. But the initial steps will be shared

recipes/openmathreasoning/scripts/simplified_recipe.py
```diff
@@ -110,50 +110,56 @@ def run_training(workspace, cluster, num_gpus, expname_prefix, wandb_params):
...
+            server_type="vllm",
+            server_gpus=num_gpus,
+            benchmarks="aime24:8,aime25:8",
+            output_dir=f"{workspace}/evals/after-training-{training_backend}",
+            num_jobs=1,
+            expname=f"{expname_prefix}-final-eval",
+            run_after=f"{expname_prefix}-training-{training_backend}",
+            wandb_name=f"{expname_prefix}-final-eval" if not wandb_params["disable_wandb"] else None,
+            wandb_project=wandb_params["wandb_project"],
+        )
```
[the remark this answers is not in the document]
Kipok: @wedu-nvidia let's do this

tests/slurm-tests/omr_simple_recipe/run_test.py
```diff
@@ -51,6 +59,7 @@ def main():
             f"{args.expname_prefix}-final-eval",
             f"{args.expname_prefix}-baseline-eval",
         ],
+        reuse_code=False,
```
Kipok: why this change?

recipes/openmathreasoning/scripts/simplified_recipe.py
```diff
@@ -24,7 +24,7 @@
 )
 
 
-def prepare(workspace, cluster, num_gpus, expname_prefix, wandb_params):
+def prepare(workspace, cluster, num_gpus, expname_prefix, backend, wandb_params):
```
Kipok: why do we need backend in this and run_sdg and initial_eval?

Kipok (changes requested):

tests/slurm-tests/omr_simple_recipe/run_test.py
wedu-nvidia: It should not change, I am not sure it got changed, I will revise back.

recipes/openmathreasoning/scripts/simplified_recipe.py
wedu-nvidia: I thought original used it for convenience so that it can pass *args. Now I reivsed it.

Kipok (approved):

PR #976, opened 2025-10-21
Add code execution evaluator

aleksficek:
Added code execution as its own evaluation. This is necessary for running code execution independent of a particular evaluation benchmark.

## Summary by CodeRabbit

* **New Features**
  * Added a code-execution evaluator that runs generated code against sandboxed test cases.
  * Configurable language and timeout settings for execution.
  * Returns per-test execution details (status, stdout/stderr, correctness) and an aggregated score for overall evaluation.

### DIFF
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -258,15 +258,18 @@ class Sandbox(abc.ABC):
         """Readiness check against the sandbox health endpoint."""
         url = f"http://{self.host}:{self.port}/health"
 
-        if self.ssh_server and self.ssh_key_path:
-            import sshtunnel_requests
+        try:
+            if self.ssh_server and self.ssh_key_path:
+                import sshtunnel_requests
 
-            sshtunnel_request = sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path)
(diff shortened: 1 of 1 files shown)


nemo_skills/evaluation/evaluator/code.py
```diff
@@ -33,6 +36,73 @@
...
+                )
+                if output.get("process_status") == "completed":
+                    self._sandbox_ready = True
+                    return
+                else:
+                    LOG.warning(f"Sandbox returned status: {output.get('process_status')}")
+            except Exception:
+                LOG.warning("Sandbox not ready yet, retrying...")
+            await asyncio.sleep(self.eval_config.timeout)
+        raise RuntimeError("Sandbox not available after multiple retries")
```
SeanNaren: I too use this functionality in the IOI code AFAIK (definitely in my own WIP code). should we put this into the local sandbox and have a new function`wait_for_sandbox` @Kipok?

Kipok: yes, we should put it into the main inference/generate.py, similar to how we wait for llm. There is an issue for this https://github.com/NVIDIA-NeMo/Skills/issues/978

SeanNaren: @aleksficek this is now complete, I think we can remove this wait_for_sandbox_ready function as we now support inside the generation module.

PR #977, opened 2025-10-21
Removing Llama from GPU tests

shtoshni:
This PR removes llama from GPU tests. For now, context retry tests are replaced with a TODO. In a separate PR, adding a faster version of that. 

## Summary by CodeRabbit

* **Tests**
  * GPU test suite refocused exclusively on Qwen models; Llama tests and scripts removed.
  * Missing test env vars now raise errors instead of being skipped.
  * Added/expanded contamination, training, and tiny-model generation scenarios; adjusted model choices and sample/threshold expectations.
  * Increased inference tokens to 2048 and enabled eager server execution across tests.

* **Chores**
  * Test-runner API endpoint updated to use localhost.

* **New Features**
  * Tiny-model export now saves a bf16 model and its tokenizer.

### DIFF
--- a/tests/gpu-tests/run_qwen.sh
+++ b/tests/gpu-tests/run_qwen.sh
@@ -20,8 +20,8 @@ pytest tests/gpu-tests/test_contamination.py -s -x
 # pytest tests/gpu-tests/test_context_retry.py -s -x
 
 # for sft we are using the tiny random model to run much faster
-ns run_cmd --cluster test-local --config_dir tests/gpu-tests --container nemo \
-    python /nemo_run/code/tests/gpu-tests/make_tiny_llm.py --model_type $NEMO_SKILLS_TEST_MODEL_TYPE
+ns run_cmd --cluster test-local --config_dir tests/gpu-tests --container vllm \
+    python3 /nemo_run/code/tests/gpu-tests/make_tiny_llm.py --model_type $NEMO_SKILLS_TEST_MODEL_TYPE
 
 # converting the model through test
 export NEMO_SKILLS_TEST_HF_MODEL=/tmp/nemo-skills-tests/$NEMO_SKILLS_TEST_MODEL_TYPE/tiny-model-hf


tests/gpu-tests/test_contamination.py
```diff
@@ -51,22 +49,25 @@ def test_check_contamination():
     run_cmd(
         cluster="test-local",
         config_dir=Path(__file__).absolute().parent,
-        container="nemo-rl",
+        container="nemo-skills",
```
Kipok: I think we need pytorch in the container and nemo-skills doesn't have it. Maybe we can use vllm container?

tests/gpu-tests/run_qwen.sh
```diff
@@ -3,19 +3,22 @@
...
-export NEMO_SKILLS_TEST_HF_MODEL=Qwen/Qwen3-4B
+export NEMO_SKILLS_TEST_HF_MODEL=Qwen/Qwen3-1.7B
 # generation/evaluation tests
 pytest tests/gpu-tests/test_eval.py -s -x
 pytest tests/gpu-tests/test_generate.py -s -x
 pytest tests/gpu-tests/test_judge.py -s -x
 pytest tests/gpu-tests/test_run_cmd_llm_infer.py -s -x
+
+# For contamination test, reasoning models are not a good choice. Switching to a instruct model.
+export NEMO_SKILLS_TEST_HF_MODEL=Qwen/Qwen3-4B-Instruct-2507
```
Kipok, line 16: can we just add ++prompt_suffix=' /no_think'

tests/gpu-tests/test_eval.py
```diff
@@ -182,10 +182,11 @@ def test_megatron_eval():
     if not model_path:
         pytest.skip("Define NEMO_SKILLS_TEST_MEGATRON_MODEL to run this test")
     model_type = os.getenv("NEMO_SKILLS_TEST_MODEL_TYPE")
+    print(f"model_type: {model_type}")
     if not model_type:
         pytest.skip("Define NEMO_SKILLS_TEST_MODEL_TYPE to run this test")
-    if model_type != "llama":
-        pytest.skip("Only llama models are supported in Megatron.")
+    if model_type != "qwen":
+        pytest.skip("Only qwen models are supported in Megatron.")
```
Kipok: let's fail with error instead of skipping - otherwise it's easy to be in a situation where some tests are accidentally skipped forever (we had it some time back, so I'm trying to minimize number of skips we do)

tests/gpu-tests/run_qwen.sh
shtoshni: Let me test it.

tests/gpu-tests/test_contamination.py
shtoshni: Interesting, the test passed in my local eval.

tests/gpu-tests/run_qwen.sh
shtoshni: Unfortunately, `Qwen/Qwen3-1.7B` just keeps spitting out false for all instances.

tests/gpu-tests/test_contamination.py
shtoshni: Ok, works with vllm container as well but requires a few more package installations.

shtoshni:
The failing GPU tests are due to NeMo-Run, and happen when there's a race condition induced by multiple generation random seeds / chunks in our tests in `test_generate.py` - `test_vllm_generate_greedy_chunked` and `test_vllm_generate_seeds`. I have informed Hemil about it and he's looking into it.

Kipok (approved):
thanks!

PR #985, opened 2025-10-23
Support environment variable expansion in mount paths

ynahshan:
## Summary
  This change adds support for `${VAR}` style environment variable expansion in mount paths,
  consistent with how `ssh_tunnel` fields are processed.

  ## Problem
  Previously, only dict fields like `ssh_tunnel.user` and `ssh_tunnel.job_dir` supported `${VAR}`
  expansion via `os.path.expandvars()`. Mount list items only supported `{VAR}` format for full-path
   replacement (e.g., `{DATA_PATH}:/data` where `DATA_PATH` contains the entire source path).

  This inconsistency meant users couldn't create user-agnostic cluster configurations with partial
  path substitution in mounts.

  ## Solution
  Added `os.path.expandvars()` call in `get_mounts_from_config()` (at
  `nemo_skills/pipeline/utils/mounts.py:421-424`) to expand `${VAR}` patterns before processing
  mount paths.

### DIFF
--- a/nemo_skills/pipeline/utils/mounts.py
+++ b/nemo_skills/pipeline/utils/mounts.py
@@ -418,8 +418,14 @@ def get_mounts_from_config(cluster_config: dict):
         if ":" not in mount:
             raise ValueError(f"Invalid mount format: {mount}. The mount path must be separated by a colon.")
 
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
Looks good ! Thanks !

PR #986, opened 2025-10-23
GenSelect Test + Fixes + Prompt filling fix

shtoshni:
This PR adds:
- Offline and Online GenSelect slurm tests using Qwen3-4B
- Few fixes in GenSelect
- The default prompt filling logic is to return list now.

### DIFF
--- a/tests/slurm-tests/qwen3_4b_evals/check_results.py
+++ b/tests/slurm-tests/qwen3_4b_evals/check_results.py
@@ -74,7 +74,7 @@ def check_results(eval_dir: str):
         eval_dir, "offline_genselect", "initial_solutions", "eval-results", "aime24", "metrics.json"
     )
     offline_initial_solutions_accuracy = float(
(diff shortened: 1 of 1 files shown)


tests/test_prompts.py
```diff
@@ -102,7 +102,7 @@ def test_generic_math_problem_augmentation_prompt():
 After the problem is completed finish your response right away.<|eot_id|><|start_header_id|>assistant<|end_header_id|>
 
 """
-    assert prompt.fill({"problem": "What's the meaning of life?"}) == expected_prompt
+    assert prompt.fill({"problem": "What's the meaning of life?"}, format_as_string=True) == expected_prompt
```
Kipok, line 105: please check the documentation for instances of prompt.fill and also prepare_data utils (that's used for sft), they all need to be updated with this

shtoshni: Yup, I fixed all the tests. Let me fix the docs.

shtoshni: Should be good to merge if the tests pass!

PR #994, opened 2025-10-26
Add os.path.expandvars() support for mount list items

ynahshan:
This change adds support for ${VAR} style environment variable expansion in mount paths, consistent with how ssh_tunnel fields are processed.

Previously, only dict fields like ssh_tunnel.user and ssh_tunnel.job_dir supported ${VAR} expansion via os.path.expandvars(). Mount list items only supported {VAR} format for full-path replacement.

This enhancement allows partial path substitution in mounts, enabling user-agnostic configurations like:
  - /long_path/users/${CLUSTER_USER}/models:/models

The change maintains backward compatibility with the existing {VAR} format.

Author: ynahshan@nvidia.com


Kipok:
thanks!

PR #999, opened 2025-10-27
Mehrzad/icpc25

mehrzads:
Adding ICPC 2025 evaluator and metrics module

### DIFF
--- a/nemo_skills/evaluation/evaluator/icpc.py
+++ b/nemo_skills/evaluation/evaluator/icpc.py
@@ -79,7 +79,7 @@ def _precompile_grader(
         sandbox = LocalSandbox()
         sandbox._owner_tid = threading.get_ident()
 
-    pre_dir = f"/tmp/icpc_pre_{problem_name}_{os.getpid()}"
+    pre_dir = f"/nemo_run/icpc_pre_{problem_name}_{os.getpid()}"
     # Create directories and files locally; sandbox shares the same filesystem
     os.makedirs(os.path.join(pre_dir, "graders"), exist_ok=True)
 
@@ -116,7 +116,7 @@ def _precompile_grader(
 
 def run_test_case(task_args: dict, worker_id: int) -> dict:
     # Use high-resolution timestamp to guarantee uniqueness across parallel calls.
-    unique_dir = f"/tmp/icpc_run_{worker_id}_{os.getpid()}_{time.time_ns()}"
+    unique_dir = f"/nemo_run/icpc_run_{worker_id}_{os.getpid()}_{time.time_ns()}"
 
     try:
         # 1. Create all necessary files locally (sandbox shares filesystem)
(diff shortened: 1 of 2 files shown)


Kipok:
> There is a lot of duplicated code between IOI and ICPC, however I'm a bit worried about combining these incase there are errors that haven't been caught yet for each individual benchmark. @Kipok does this make sense?

yeah, up to you. It will just be harder for you to maintain this with a lot of duplication, but if you think it makes sense at this stage, no concerns from my side

nemo_skills/evaluation/evaluator/icpc.py
```diff
@@ -0,0 +1,372 @@
...
+    loop = getattr(_precompile_loop_tls, "loop", None)
+    if loop is None or loop.is_closed():
+        loop = asyncio.new_event_loop()
+        _precompile_loop_tls.loop = loop
+
+    # Use the loop within this thread exclusively.
+    return loop.run_until_complete(sandbox.execute_code(cmd, language=language, timeout=timeout))[0]
+
+
+def wait_for_sandbox(sandbox, timeout: int = 240, poll: float = 1.0):
```
SeanNaren: This function isn't needed anymore as we now wait for the sandbox server higher level!

nemo_skills/evaluation/evaluator/icpc.py
```diff
@@ -0,0 +1,480 @@
...
+    return hashlib.sha256(text.encode("utf-8", errors="replace")).hexdigest()
+
+
+@nested_dataclass(kw_only=True)
+class ICPCEvaluatorConfig(BaseEvaluatorConfig):
+    test_file: str = "test_metadata.json"
+    input_file: str = None
+    num_workers: int = 16  # number of test workers
+    test_batch_size: int = 16  # number of tests to run concurrently
+    overwrite: bool = False
```
SeanNaren: I think we can remove this variable, its not used anywhere

nemo_skills/evaluation/evaluator/icpc.py
```diff
@@ -0,0 +1,480 @@
...
+
+    # Run compile.sh inside the sandbox (same filesystem)
+    _sandbox_exec_sync(sandbox, f"cd {pre_dir} && ./compile.sh || true", language="shell", timeout=120)
+
+    return pre_dir
+
+
+def run_test_case(task_args: dict, worker_id: int) -> dict:
+    # Use high-resolution timestamp to guarantee uniqueness across parallel calls.
+    unique_dir = f"/tmp/icpc_run_{worker_id}_{os.getpid()}_{time.time_ns()}"
```
SeanNaren: @Kipok can we assume /tmp is always shared between sandbox and this code?

Kipok: I think it's only shared if there is an explicit mount into /tmp and you enable mounts for sandbox. So generally, it won't be shared. A better way might be to save whatever you need into /nemo_run/code (which should always be there if mounts are enabled) and then make sure to clean it up as code folder can be shared across multiple jobs. In fact that folder should be there even if general mounts are disabled, so this could be a good way to do it (and we don't have to write to /code, can write just to /nemo_run/<>)

SeanNaren: hmm, @mehrzads are you sure /tmp isn't mounting somehow? from what igor is saying your code shouldn't work?

SeanNaren (approved):

PR #1000, opened 2025-10-28
Wait for sandbox to initialize

### LINKED ISSUE #978: Add functionality to wait for sandbox
when there is no llm server, we need to explicitly wait for sandbox to start otherwise the job crashes

SeanNaren:
Fixes #978

### DIFF
--- a/docs/basics/cluster-configs.md
+++ b/docs/basics/cluster-configs.md
@@ -10,7 +10,7 @@ The cluster config defines an executor (local or slurm), mounts for data/model a
 such as account, partition, ssh-tunnel arguments and so on.
 
 The recommended way to launch jobs on slurm is by running all commands locally and specifying `ssh_tunnel` portion in cluster config
-to let [NeMo-Run](https://github.com/NVIDIA/NeMo-Run) know how to connect there.
+to let [NeMo-Run](https://github.com/NVIDIA-NeMo/Run) know how to connect there.
 But if you prefer to run from the cluster directly, you can install Nemo-Skills there
 and then only specify `job_dir` parameter without using `ssh_tunnel` section in the config.
 
--- a/docs/pipelines/index.md
+++ b/docs/pipelines/index.md
@@ -80,7 +80,7 @@ All scripts inside pipeline folder have the following parameters.
   You can also use `NEMO_SKILLS_CONFIG_DIR` environment variable for this purpose.
 - **--log_dir**: Can be used to customize the location of slurm logs.
 - **--expname**: You can always specify an experiment name, which is a
-  [NeMo-Run](https://github.com/NVIDIA/NeMo-Run) concept. This will control where
+  [NeMo-Run](https://github.com/NVIDIA-NeMo/Run) concept. This will control where
   the metadata is stored, the slurm job name and allows you to chain jobs one
   after the other using the `--run_after` argument.
 - **--run_after**: Can be used in conjunction with `--expname` to chain jobs to
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -30,7 +30,7 @@ langcodes
 litellm[caching] == 1.77.5 # Requires patching the logging worker (See nemo_skills/inference/patch_litellm_logging.py)
 math-verify[antlr4_9_3]
 mcp
-nemo_run @ git+https://github.com/NVIDIA/NeMo-Run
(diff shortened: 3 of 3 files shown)


nemo_skills/code_execution/sandbox.py
```diff
@@ -253,6 +253,15 @@ async def is_proof_correct(self, pred_output, timeout=30.0):
             return "timeout"
         return determine_proof_status(output)
 
+    async def wait_for_sandbox(self, timeout: int = 5):
+        while True:
+            try:
+                output, _ = await self.execute_code("print('test')", language="python", timeout=timeout)
```
gwarmstrong: is there any specific reason to not check the health check endpoint instead?

https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/code_execution/local_sandbox/local_sandbox_server.py#L680-L682

nemo_skills/inference/generate.py
```diff
@@ -624,6 +624,9 @@ async def _generate_and_save_datapoint(self, data_point, all_data, fout, pbar):
     async def async_loop(self, data):
         """Async loop to generate generations using asyncio."""
 
+        # First wait for sandbox to be initialized (if needed)
+        await self.wait_for_sandbox()
```
Kipok: let's move this inside generate() right after self.wait_for_server(). Will make it a bit easier in case people want to override the full async loop in subclasses but still reuse waiting functionality

nemo_skills/code_execution/sandbox.py
SeanNaren: ah didn't know that existed, let me use this instead!

nemo_skills/inference/generate.py
SeanNaren: It is an async function which is why I put it into the async loop; can I put it in wait_for_server even though it is async?

SeanNaren: Fixed now!

nemo_skills/code_execution/sandbox.py
```diff
@@ -253,6 +254,31 @@ async def is_proof_correct(self, pred_output, timeout=30.0):
...
 
+    def _check_ready(self, timeout: float = 5.0) -> bool:
+        """Readiness check against the sandbox health endpoint."""
+        url = f"http://{self.host}:{self.port}/health"
+
+        if self.ssh_server and self.ssh_key_path:
+            import sshtunnel_requests
+
+            try:
+                sshtunnel_request = sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path)
```
gwarmstrong: NIT: can you restructure so exception handling and status code parsing is only written once?

nemo_skills/code_execution/sandbox.py
```diff
@@ -253,6 +254,31 @@ async def is_proof_correct(self, pred_output, timeout=30.0):
...
+                output = sshtunnel_request.get(url=url, timeout=timeout)
+                return 200 <= output.status_code < 300
```
gwarmstrong: check for status_code == 200 should be sufficient

Kipok (approved):

nemo_skills/code_execution/sandbox.py
```diff
@@ -253,6 +254,27 @@ async def is_proof_correct(self, pred_output, timeout=30.0):
...
+            if self.ssh_server and self.ssh_key_path:
+                import sshtunnel_requests
+
+                sshtunnel_request = sshtunnel_requests.from_url(f"ssh://{self.ssh_server}:22", self.ssh_key_path)
+                response = sshtunnel_request.get(url=url, timeout=timeout)
+            else:
+                with httpx.Client() as client:
+                    response = client.get(url=url, timeout=timeout)
+            return response.status_code == 200
+        except Exception:
```
Kipok: actually on a second thought, I think we shouldn't have a blank exception here. If there is some error in this part, this can lead to the job just being stuck forever and could be hard to see why it happens. The exception shouldn't normally be thrown, right? So we can probably just remove that part and then should be good to merge

Kipok (changes requested):
realized we should probably remove exception

nemo_skills/code_execution/sandbox.py
SeanNaren: good catch, done!

PR #1001, opened 2025-10-28
Add an interface for general sbatch arguments 

samodi-nv:
Adds `sbatch_arguments` parameter into the pipelines so users may pass in arbitrary sbatch arguments (e.g. `--comment`) to the launch script.

The arguments should be given as a JSON-string.

### DIFF
--- a/nemo_skills/pipeline/utils/exp.py
+++ b/nemo_skills/pipeline/utils/exp.py
@@ -179,6 +179,60 @@ def get_executor(
     overlap: bool = False,
     with_ray: bool = False,
 ):
+    """Create and configure a nemo-run executor for the target environment.
+
+    Depending on `cluster_config['executor']`, this returns one of:
+    - `"none"`: a LocalExecutor (execute directly without container/scheduler)
+    - `"local"`: a DockerExecutor (runs locally with host networking and mounts)
+    - `"slurm"`: a SlurmExecutor (submits jobs to SLURM with inferred settings)
+
+    The function derives environment variables, mounts, container image, resource
+    flags, and logging details from `cluster_config` and the provided parameters.
+    For SLURM, it sets srun/sbatch arguments, selects the partition based on
+    `gpus_per_node`, wires job dependencies, and configures heterogeneous job
+    groups and Ray-specific logging when requested. For local Docker, all GPUs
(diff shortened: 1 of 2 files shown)


nemo_skills/pipeline/megatron_lm/train.py
```diff
@@ -208,6 +213,17 @@ def train_megatron_lm(
         extra_arguments=extra_arguments,
     )
 
+    # Add extra sbatch arguments
```
Kipok: could you please move this into a new function inside nemo_skills/pipeline/utils/cluster.py (or another file in that folder), so that we can make it just a one-line change in each of the pipeline scripts and reuse all logic?

nemo_skills/pipeline/megatron_lm/train.py
```diff
@@ -208,6 +213,17 @@ def train_megatron_lm(
...
+    slurm_kwargs = {"exclusive": exclusive} if exclusive else {}
+    if sbatch_arguments:
+        try:
+            sbatch_kwargs = json.loads(sbatch_arguments)
+            slurm_kwargs.update(sbatch_kwargs)
+        except json.JSONDecodeError as e:
```
Kipok: let's also check if it's already a dict, then we skip json conversion. This way people can provide it as both json-string or a dict which is more convenient when launching from python

Kipok:
can you please also force-push commits with -s flag to fix https://github.com/NVIDIA-NeMo/Skills/pull/1001/checks?check_run_id=53893226069 check

nemo_skills/pipeline/megatron_lm/train.py
samodi-nv: addressed this and the dict comment

nemo_skills/pipeline/megatron_lm/train.py
```diff
@@ -162,6 +163,10 @@ def train_megatron_lm(
         "E.g. 'pip install my_package'",
     ),
     dry_run: bool = typer.Option(False, help="If True, will not run the job, but will validate all arguments."),
+    sbatch_arguments: str | dict = typer.Option(
```
Kipok: I think this fails typer checks, we can just keep it as str here as type validation is bypassed in python mode anyway. The docstring reflects already that this can be passed as dict

PR #1002, opened 2025-10-28
Gnalbandyan/robust v02

gnalbandyan:
Update prompt set for robustness evaluation, add usage in cli for `ns robust_eval`, add relaxed MCQ answer extraction, update docstrings

### DIFF
--- a/docs/basics/cluster-configs.md
+++ b/docs/basics/cluster-configs.md
@@ -10,7 +10,7 @@ The cluster config defines an executor (local or slurm), mounts for data/model a
 such as account, partition, ssh-tunnel arguments and so on.
 
 The recommended way to launch jobs on slurm is by running all commands locally and specifying `ssh_tunnel` portion in cluster config
-to let [NeMo-Run](https://github.com/NVIDIA/NeMo-Run) know how to connect there.
+to let [NeMo-Run](https://github.com/NVIDIA-NeMo/Run) know how to connect there.
 But if you prefer to run from the cluster directly, you can install Nemo-Skills there
 and then only specify `job_dir` parameter without using `ssh_tunnel` section in the config.
 
--- a/docs/pipelines/index.md
+++ b/docs/pipelines/index.md
@@ -80,7 +80,7 @@ All scripts inside pipeline folder have the following parameters.
   You can also use `NEMO_SKILLS_CONFIG_DIR` environment variable for this purpose.
 - **--log_dir**: Can be used to customize the location of slurm logs.
 - **--expname**: You can always specify an experiment name, which is a
-  [NeMo-Run](https://github.com/NVIDIA/NeMo-Run) concept. This will control where
+  [NeMo-Run](https://github.com/NVIDIA-NeMo/Run) concept. This will control where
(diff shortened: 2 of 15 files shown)


nemo_skills/evaluation/math_grader.py
```diff
@@ -92,14 +92,29 @@ def math_equal(gt_answer, predicted_answer, take_modulo: int | None = None, **kw
...
 
-def extract_answer(string: str, extract_from_boxed: bool = True, extract_regex: str = r"The final answer is (.+)$"):
-    """Extract Answer String from \\boxed expression or based on regex"""
-    if not extract_from_boxed:
-        match = re.search(extract_regex, string)
-        if match:
-            return match.group(1)
-        return None
+def extract_answer(
+    string: str, extract_from_boxed: bool = True, extract_regex: str = r"The final answer is (.+)$", relaxed=True
```
Kipok: let's have relaxed=False as default here to preserve current behavior for math benchmarks, which also rely on this function

nemo_skills/prompt/config/robustness/prompt_set_config.yaml
```diff
@@ -1,46 +1,25 @@
 gpqa:
```
Kipok, line 1: might be better to use a dict here with explicit names like prompt: extract_regex: . This way you can have a dataclass inside starting script and validate names and will also make it easier to extend with more parameters in the future

Kipok (changes requested):

nemo_skills/evaluation/evaluator/mcq.py
```diff
@@ -30,14 +30,24 @@ class MCQEvaluatorConfig(BaseEvaluatorConfig):
     extract_from_boxed: bool = True
     # only used if extract_from_boxed is False
     extract_regex: str = r"The final answer is (.+)$"
+    # if relaxed is True:
+    #   extract from boxed FIRST, if not found, extract from regex
+    # if relaxed is False:
+    #   if extract_from_boxed is True -> extract from boxed{} ONLY
+    #   else extract from regex ONLY
+    relaxed: bool = False
```
Kipok: should this be True? That was the old behavior, we probably don't want to change it by default?

nemo_skills/pipeline/robust_eval.py
```diff
@@ -26,6 +29,12 @@
 LOG = logging.getLogger(get_logger_name(__file__))
 
 
+@dataclass
+class Prompt:
```
Kipok: Maybe PromptConfig? Or perhaps EvalConfig?

nemo_skills/pipeline/robust_eval.py
```diff
@@ -63,20 +88,20 @@ def robust_eval(
...
-    ```
-    <benchmark_name>:
-      - <path_to_prompt_1>
-      - <path_to_prompt_2>
-      ...
-    <another_benchmark_name>:
-        - <path_to_prompt_1>
-        - <path_to_prompt_2>
+    benchmark_name:
+        - path_to_prompt_1
```
Kipok: not up to date?

PR #1005, opened 2025-10-29
Bump vLLM version

activatedgeek:
Upgrades vLLM to `v0.10.2` by default.

### DIFF
--- a/nemo_skills/__init__.py
+++ b/nemo_skills/__init__.py
@@ -16,18 +16,12 @@ from nemo_skills.version import __version__
 
 # only used in ns setup command to initialize with defaults
 _containers = {
-    "trtllm": {
-        "default": "nvcr.io/nvidia/tensorrt-llm/release:1.0.0",
-    },
-    "vllm": {
-        "default": "vllm/vllm-openai:v0.10.2",
-    },
-    "sglang": {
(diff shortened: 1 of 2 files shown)


wedu-nvidia (approved):

dockerfiles/README.md
```diff
@@ -36,6 +36,4 @@ We directly use official `lmsysorg/sglang:v0.5.4` image.
 
 ## Building vllm image
 
-We directly use official `vllm/vllm-openai:v0.10.1.1` image.
-
-For arm64 we instead use `vllm/vllm-openai:v0.10.2` image.
+We directly use official `vllm/vllm-openai:v0.10.2` image.
```
gwarmstrong, line 39: I think we are currently experiencing some issue on v0.10.2 and v0.10.1.1 is the most stable--are we sure we want to bump this?

gwarmstrong (commented):
@activatedgeek it's a little unclear to me what this PR actually accomplishes. Can you add a description of some kind so I understand?

dockerfiles/README.md
activatedgeek: I think that's what @wedu-nvidia was currently testing, and has confirmed it works.

PR #1007, opened 2025-10-29
Add sbatch arguments to generate pipeline

samodi-nv:
Adds the new sbatch arguments parameter to the generate pipeline

### DIFF
--- a/nemo_skills/code_execution/sandbox.py
+++ b/nemo_skills/code_execution/sandbox.py
@@ -17,6 +17,7 @@ import asyncio
 import json
 import logging
 import os
+import time
 import traceback
 import uuid
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


Kipok:
yeah, agree, will separate in multiple prs in the future

about the name change - there is no big reason except just to make it clear that this is a dictionary and not a string of arguments like in other cases when things are named X_args or X_arguments. So I think it's more explicit like this and keeps things consistent. If this was a well-established feature that many people use then yeah it's not good enough reason to make a name change. But since this was added a few days ago, I thought we can just rename it quickly to keep things consistent and it's very unlikely to affect anyone. Let me know if you think otherwise

Kipok:
also there is really no functional change, it just adds missing arguments to remaining pipelines (sbatch_kwargs was missing in generate and qos was missing in convert) but otherwise this is just refactoring of internals and renaming

PR #1009, opened 2025-10-30
training_jobs_renaming

wedu-nvidia:
solve this
https://github.com/NVIDIA-NeMo/Skills/pull/788/


Kipok:
looks good, but let's merge this on Monday as this is a breaking change

Kipok (approved):
thanks!

PR #1018, opened 2025-10-31
Support Multiple Model initialization

gwarmstrong:
This PR supports having multiple models initialized in a generate command, like so:

```

from nemo_skills.pipeline.cli import wrap_arguments, generate

generate(
    ctx=wrap_arguments(""),
    cluster="ord",
    partition="interactive",
    model=["/hf_models/Qwen3-32B", "/hf_models/Qwen3-8B"],
    server_gpus=8,  # Broadcast to both models
    server_nodes=[2,1], # 2 nodes for 32B, 1 node for 8B
    server_type="vllm",
    generation_module="multiturn-conversation/multiturn_conversation.py",
    input_file="/nemo_run/code/multiturn-conversation/sample_prompts.jsonl",
    output_dir=f"/workspace/multiturn-conversation/minimal_example/output",
)
```

This generation module is not available in the branch (and likely won't due to the specificity of its implementation) but reach out to me if you want the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Here are some of the features we support:
     - [**Long-context**](https://nvidia-nemo.github.io/Skills/evaluation/long-context): e.g. [ruler](https://nvidia-nemo.github.io/Skills/evaluation/long-context/#ruler), [mrcr](https://nvidia-nemo.github.io/Skills/evaluation/long-context/#mrcr), [aalcr](https://nvidia-nemo.github.io/Skills/evaluation/long-context/#aalcr)
     - [**Tool-calling**](https://nvidia-nemo.github.io/Skills/evaluation/tool-calling): e.g. [bfcl_v3](https://nvidia-nemo.github.io/Skills/evaluation/tool-calling/#bfcl_v3)
     - [**Multilingual**](https://nvidia-nemo.github.io/Skills/evaluation/multilingual): e.g. [mmlu-prox](https://nvidia-nemo.github.io/Skills/evaluation/multilingual/#mmlu-prox), [FLORES-200](https://nvidia-nemo.github.io/Skills/evaluation/multilingual/#FLORES-200), [wmt24pp](https://nvidia-nemo.github.io/Skills/evaluation/multilingual/#wmt24pp)
+    - [**Speech & Audio**](https://nvidia-nemo.github.io/Skills/evaluation/speech-audio): e.g. [mmau-pro](https://nvidia-nemo.github.io/Skills/evaluation/speech-audio/#mmau-pro)
   - Easily parallelize each evaluation across many slurm jobs, self-host LLM judges, bring your own prompts or change benchmark configuration in any other way.
 - [Model training](https://nvidia-nemo.github.io/Skills/pipelines/training): Train models using [NeMo-RL](https://github.com/NVIDIA-NeMo/RL/) or [verl](https://github.com/volcengine/verl).
 
--- a/docs/evaluation/index.md
+++ b/docs/evaluation/index.md
@@ -10,6 +10,7 @@ We support many popular benchmarks and it's easy to add new in the future. The f
 - [**Long-context**](./long-context.md): e.g. [ruler](./long-context.md#ruler), [mrcr](./long-context.md#mrcr)
 - [**Tool-calling**](./tool-calling.md): e.g. [bfcl_v3](./tool-calling.md#bfcl_v3)
 - [**Multilingual**](./multilingual.md): e.g. [mmlu-prox](./multilingual.md#mmlu-prox), [flores-200](./multilingual.md#FLORES-200), [wmt24pp](./multilingual.md#wmt24pp)
+- [**Speech & Audio**](./speech-audio.md): e.g. [mmau-pro](./speech-audio.md#mmau-pro)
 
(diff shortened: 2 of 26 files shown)


nemo_skills/pipeline/utils/declarative.py
```diff
@@ -186,11 +186,21 @@ def __post_init__(self):
...
         if self.het_group_index is None:
-            return "127.0.0.1"  # Local fallback
-        # For heterogeneous SLURM jobs, resolve nodelist to actual hostname
-        return f"$(scontrol show hostnames $SLURM_JOB_NODELIST_HET_GROUP_{self.het_group_index} | head -n1)"
+            return "127.0.0.1"  # Local fallback for non-heterogeneous jobs
+
+        # Use our SLURM helper utility to parse the nodelist
+        var_name = f"SLURM_JOB_NODELIST_HET_GROUP_{self.het_group_index}"
+        # TODO: maybe instead  do something that is fully bash ready so it can be run in any container
+        return f"$(python3 -m nemo_skills.pipeline.utils.slurm_helpers ${var_name})"
```
Kipok: doesn't nemo-run do something like this already? Can you check the generated sbatch to see if there is already a hostname provided automatically with some env var?

nemo_skills/pipeline/utils/generation.py
```diff
@@ -188,6 +188,10 @@ def get_generation_cmd(
     wandb_parameters=None,
     with_sandbox: bool = False,
     script: str = "nemo_skills.inference.generate",
+    # Optional: for multi-model generation
+    server_addresses: list[str] | None = None,
+    model_names: list[str] | None = None,
+    num_models: int | None = None,
```
Kipok: isn't `num_models == len(model_names)` always?

nemo_skills/pipeline/generate.py
```diff
@@ -44,109 +44,231 @@
 # TODO: add num_jobs here for consistency with eval?
 
 
-def _create_commandgroup_from_config(
-    generation_cmd: str,
-    server_config: Optional[Dict],
-    with_sandbox: bool,
-    sandbox_port: Optional[int],
+def _normalize_models_config(
```
Kipok: maybe move both new funcs to a utils file as this will be shared across pipelines

nemo_skills/pipeline/generate.py
```diff
@@ -44,109 +44,231 @@
...
+    model: Optional[List[str]],
+) -> List[str]:
+    """
+    Normalize model specification to list.
+
+    Args:
+        model: List of model paths (Typer converts single values to single-element lists)
```
Kipok: I think we shouldn't assume this as people can call pipelines from Python and it bypasses all typer logic and just uses that as normal function. So we don't want code that does `model="my-model-name"` to break

nemo_skills/pipeline/generate.py
```diff
@@ -393,36 +552,52 @@ def generate(
...
-            )
-            cmd = pipeline_utils.wrap_python_path(cmd=cmd)
+            server_configs = []
+            server_addresses_resolved = []
+
+            # For n=1, we need extra_arguments from configure_client
+            # For n>1, we build server addresses ourselves
+            extra_args_from_first_server = extra_arguments_original
+
+            for idx in range(num_models):
```
Kipok: consider using model_idx instead of idx to make this a bit clearer as this is already inside two loops with their own indices

nemo_skills/pipeline/generate.py
gwarmstrong: sure--do you see any reason why moving them into the `nemo_skills.pipeline.utils.generation` file would not work?

Kipok: no reason, should be good to move there

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: good catch, looks like when there is a het job it is adding things like this to the sbatch, which is perfect:

```
export SLURM_MASTER_NODE_HET_GROUP_0=$(scontrol show hostnames $SLURM_JOB_NODELIST_HET_GROUP_0 | head -n1)
export SLURM_MASTER_NODE_HET_GROUP_1=$(scontrol show hostnames $SLURM_JOB_NODELIST_HET_GROUP_1 | head -n1)
```

nemo_skills/pipeline/generate.py
gwarmstrong: updated in [5744618](https://github.com/NVIDIA-NeMo/Skills/pull/1018/commits/57446182c778ce306a1a66b25e7452a1578a002e)

nemo_skills/pipeline/generate.py
gwarmstrong: fixed in 2bd3852

nemo_skills/pipeline/generate.py
gwarmstrong: updated in da2fdd6

nemo_skills/pipeline/utils/generation.py
gwarmstrong: updated in da2fdd6

nemo_skills/pipeline/utils/declarative.py
gwarmstrong: updated in da2fdd6

Kipok (approved):
lgtm, thanks! Feel free to merge as long as both gpu and slurm tests are passing

gwarmstrong:
Closed in favor of #1052

PR #1022, opened 2025-11-03
Adding MMAU-Pro Benchmark Evaluation and Megatron-LM Server Support

melllinia:
# MMAU-Pro Evaluation and Megatron-LM Server Support

## Summary
This PR adds configuration and scripts to enable **MMAU-Pro evaluation** for multimodal language models using the **NeMo framework** and the **Megatron-LM inference server**.

This is the first SpeechLM benchmark to be supported in NeMo Skills with the Megatron-LM server, providing a **standardized evaluation pipeline** for audio-language tasks.

## Motivation
The MMAU-Pro benchmark provides an evaluation for multimodal audio-language models. By integrating it with NeMo-Skills and the Megatron-LM server, we enable **reproducible evaluation** of model performance on audio question answering tasks. This lays the groundwork for future SpeechLM benchmarks to be supported under the same evaluation infrastructure.

## What's New
- Added MMAU-Pro evaluation support in NeMo Skills.  
- Configured Megatron-LM server for inference, incorporating modifications by Steve Huang. 
- Provided an example to run evaluation via `ns eval`.  

## Implementation Notes

The benchmark consists of three main evaluation groups, each handled with separate `test.jsonl` files and corresponding jobs:

1. **Open-ended questions** – evaluated by LLM judge (`qwen2.5-7b-instruct`).  
2. **Closed-form questions** – evaluated via embedding matching (`nv embed`).  
3. **Instruction following** – evaluated by straightforward functions.  

All scores are then combined into a **final score**.

### Preparing the Dataset
- To download `test.jsonl` data and generate subgroup `__init__.py` files, run:
```bash
python nemo_skills/dataset/mmau-pro/prepare.py
```
- To also download the audio files (~50GB), run:
```bash
python nemo_skills/dataset/mmau-pro/prepare.py --with-audio --download-dir /path/to/dir
```

## Example: Running the Evaluation

```bash
export MEGATRON_PATH="/lustre/fsw/portfolios/llmservice/users/heh/code/megatron-lm-vlm2-dev"
export CKPT_PATH="/lustre/fsw/portfolios/convai/users/mmkrtchyan/avlm_checkpoints/draco-mcore-nm_5p5_h_9b-cradio-parakeet-nemo-stage2-alm-nodes16-seq16384-bz2048-tp4-vlm2-branch-1009-tp1"
export MODEL_CFG_PATH="/lustre/fsw/portfolios/convai/users/mmkrtchyan/avlm_checkpoints/draco-mcore-nm_5p5_h_9b-cradio-parakeet-nemo-stage2-alm-nodes16-seq16384-bz2048-tp4-vlm2-branch-1009-tp1/config.yaml"

export OUTPUT_DIR="mmau-pro_eval"
export SERVER_ENTRYPOINT="$MEGATRON_PATH/examples/multimodal/run_avlm_text_generation_server.py"
export SERVER_CONTAINER="/lustre/fsw/portfolios/llmservice/users/trintamaki/workspace/containers/megatron-dev-img-05142025-pytorch-dev-te-cd37379-editable-energon-mamba-fix-vlmeval-av.sqsh"

export HF_TOKEN='YOUR_HF_TOKEN'
export WANDB='YOUR_WAND_KEY'
export NVIDIA_API_KEY='YOUR_NVIDIA_API_KEY' 

export WANDB_API_KEY=${WANDB} && \
export HF_TOKEN=${HF_TOKEN} && \
export NVIDIA_API_KEY=${NVIDIA_API_KEY} && \
export MEGATRON_PATH="$MEGATRON_PATH" && \
ns eval \
  --cluster oci_iad \
  --output_dir /path/to/output/dir/$OUTPUT_DIR \
  --benchmarks mmau-pro \
  --server_type megatron \
  --server_gpus 1 \
  --model $CKPT_PATH \
  --server_entrypoint $SERVER_ENTRYPOINT \
  --server_container $SERVER_CONTAINER \
  --data_dir="/dataset" \
  --installation_command "pip install sacrebleu" \
  ++prompt_suffix='/no_think' \
  --server_args "--inference-max-requests 1 \
                 --model-config ${MODEL_CFG_PATH} \
                 --num-tokens-to-generate 256 --temperature 1.0 --top_p 1.0"
 ```                
                 
## Notes

- **Closed-form evaluation** uses the NV-Embed model. NV-Embed requires a lower version of `transformers` than Megatron-LM and some additional packages. Transformers version is decreased but it doesn't affect the generation, because happens after it.
- The evaluation logic is organized by `judge_type` in `eval.py` to allow **GPU-based scoring**.  
- Since the benchmark runs in **subgroups**, each subgroup's **chunk size** is specified in its `__init__.py`.  
- Used custom megatron-LM container where `sacrebleu` package was missing, adding that with `--installation_command`.

### DIFF
--- a/docs/evaluation/speech-audio.md
+++ b/docs/evaluation/speech-audio.md
@@ -21,12 +21,10 @@ MMAU-Pro (Multimodal Audio Understanding - Pro) is a comprehensive benchmark for
 
 MMAU-Pro requires audio files for meaningful evaluation. **Audio files are downloaded by default** to ensure proper evaluation.
 
-!!! warning "Git Repository Safety Check"
-    By default, preparing data inside a git repository will fail to prevent accidentally packaging large data files (in this case, ~50GB audio files). You can either:
-
-    1. Specify `--data_dir` to download data outside the repository (recommended)
-    2. Use `--skip_data_dir_check` to bypass this safety check
-    3. Use `--no-audio` for text-only mode (not recommended for proper evaluation)
+!!! warning "Running without audio files"
+    If you want to evaluation without audio files (not recommended) use
+    `--no-audio` flag. In this case you can also set `--skip_data_dir_check`
+    as data is very lightweight when audio files aren't being used.
 
 ### Data Preparation
 
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -39,7 +39,7 @@ jobs:
         pip uninstall -y nemo-skills nemo_run
         pip install -e .
         pip install -r requirements/common-tests.txt
-        ns prepare_data gsm8k human-eval mbpp algebra222 mmlu ifeval math-500 amc23 aime24 --skip_data_dir_check
+        ns prepare_data gsm8k human-eval mbpp algebra222 mmlu ifeval math-500 amc23 aime24
     - name: Build Docker image
       run: |
         cd ${{ github.run_id }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,5 +42,5 @@ jobs:
         docker run --rm --network=host nemo-skills-sandbox-image &
         sleep 10
         set -o pipefail # this will make sure next line returns non-0 exit code if tests fail
-        ns prepare_data gsm8k math-500 --skip_data_dir_check
+        ns prepare_data gsm8k math-500
         python -m pytest tests/ -m "not gpu" --junitxml=pytest.xml --cov-report=term-missing:skip-covered --cov=nemo_skills --cov=pipeline --durations=30 -rs -s -vvv
--- a/nemo_skills/pipeline/prepare_data.py
+++ b/nemo_skills/pipeline/prepare_data.py
@@ -14,7 +14,6 @@
 
 import logging
 import os
-from pathlib import Path
 from typing import List
 
 import typer
@@ -32,7 +31,7 @@ LOG = logging.getLogger(get_logger_name(__file__))
 
 
 # TODO: read this from init.py
-DATASETS_REQUIRE_DATA_DIR = ["ruler", "ioi24"]
+DATASETS_REQUIRE_DATA_DIR = ["ruler", "ioi24", "mmau-pro"]
 
 
 @app.command(context_settings={"allow_extra_args": True, "ignore_unknown_options": True})
@@ -74,7 +73,9 @@ def prepare_data(
     ),
     skip_data_dir_check: bool = typer.Option(
         False,
-        help="Skip check that prevents downloading large data files into the git repository.",
+        help="Some datasets require very large input files and we will fail with error if data_dir "
+        "is not specified (to avoid accidentally uploading them to cluster with every job). "
+        "If you're running things locally or have another reason to bypass this check, set this flag.",
     ),
     sbatch_kwargs: str = typer.Option(
         "",
@@ -89,12 +90,13 @@ def prepare_data(
     extra_arguments = f"{' '.join(ctx.args)}"
     command = f"python -m nemo_skills.dataset.prepare {extra_arguments}"
 
-    if not data_dir:
+    if not data_dir and not skip_data_dir_check:
         for dataset in DATASETS_REQUIRE_DATA_DIR:
             if dataset in extra_arguments:
                 raise ValueError(
-                    f"Dataset {dataset} contains very large input data and requires a data_dir to be specified. "
-                    "Please provide --data_dir argument."
+                    f"Dataset {dataset} contains very large input data and it's recommended to have a "
+                    "data_dir to be specified to avoid accidentally uploading large data on cluster with every job. "
+                    "Please provide --data_dir argument or set --skip_data_dir_check to bypass this safety check."
                 )
 
     if data_dir and cluster is None:
@@ -113,22 +115,6 @@ def prepare_data(
         # if we use container, it will mess up permissions, so as a workaround
         # setting executor to none
         cluster_config["executor"] = "none"
-
-    # Check if preparing data inside git repository without specifying data_dir
-    # Skip check if --no-audio is used (will be generalized to other datasets in the future)
-    if not data_dir and not skip_data_dir_check and "--no-audio" not in extra_arguments:
-        current = Path.cwd()
-        while current != current.parent:
-            if (current / ".git").exists():
-                raise ValueError(
-                    f"Error: Preparing data inside git repository at {current}.\n"
-                    f"Large data files will be packaged with the repository and uploaded to the cluster.\n"
-                    f"Please either:\n"
-                    f"  1. Specify --data_dir to prepare data outside the repository\n"
-                    f"  2. Use --skip_data_dir_check to bypass this safety check\n"
-                    f"  3. Use --no-audio if you want to prepare the dataset without audio files"
-                )
-            current = current.parent
 
     if cluster_config["executor"] == "slurm" and not data_dir:
         raise ValueError(
--- a/tests/gpu-tests/test_eval.py
+++ b/tests/gpu-tests/test_eval.py
@@ -266,7 +266,6 @@ def test_prepare_and_eval_all_datasets():
         config_dir=str(config_dir),
         data_dir=str(data_dir),
         expname=f"prepare-all-datasets-{model_type}",
-        skip_data_dir_check=True,
     )
 
     eval_kwargs = dict(
--- a/tests/slurm-tests/gpt_oss_python_aime25/run_test.py
+++ b/tests/slurm-tests/gpt_oss_python_aime25/run_test.py
@@ -57,7 +57,7 @@ def main():
 
     args = parser.parse_args()
 
-    prepare_data(ctx=wrap_arguments("aime25"), skip_data_dir_check=True)
+    prepare_data(ctx=wrap_arguments("aime25"))
 
     eval_expname = eval_gpt_oss_python(
         workspace=args.workspace,
--- a/tests/slurm-tests/qwen3_4b_evals/run_test.py
+++ b/tests/slurm-tests/qwen3_4b_evals/run_test.py
@@ -142,7 +142,7 @@ def main():
 
     args = parser.parse_args()
 
-    prepare_data(ctx=wrap_arguments("bfcl_v3 aime24"), skip_data_dir_check=True)
+    prepare_data(ctx=wrap_arguments("bfcl_v3 aime24"))
 
     bfcl_expname = eval_qwen3_bfcl(
         workspace=args.workspace,
--- a/tests/slurm-tests/qwen3coder_30b_swebench/run_test.py
+++ b/tests/slurm-tests/qwen3coder_30b_swebench/run_test.py
@@ -56,7 +56,7 @@ def main():
         prepare_data_args = "swe-bench"
     else:
         prepare_data_args = f"swe-bench --container_formatter {args.container_formatter}"
-    prepare_data(ctx=wrap_arguments(prepare_data_args), skip_data_dir_check=True)
+    prepare_data(ctx=wrap_arguments(prepare_data_args))
 
     for agent_framework in ["openhands", "swe_agent"]:
         workspace = f"{args.workspace}/{agent_framework}"
--- a/tests/slurm-tests/super_49b_evals/run_test.py
+++ b/tests/slurm-tests/super_49b_evals/run_test.py
@@ -321,7 +321,6 @@ def main():
 
     prepare_data(
         ctx=wrap_arguments("gpqa mmlu-pro hle livecodebench scicode bfcl_v3 math-500 aime24 aime25"),
-        skip_data_dir_check=True,
     )
 
     setup(workspace=args.workspace, cluster=args.cluster, expname_prefix=args.expname_prefix)

### CODE REVIEW

File: nemo_skills/dataset/mmau-pro/prepare.py [hunk 1]
```diff
@@ -0,0 +1,198 @@
...
+from datasets import load_dataset
+from tqdm import tqdm
+
+from nemo_skills.dataset.utils import generate_subgroup_init_files, get_mcq_fields
+
+
+def _get_mmau_pro_subgroup_configs():
+    """Define the __init__.py configurations for each MMAU-Pro subgroup."""
+    return {
+        "closed_form": """# Closed-form questions evaluated with NV Embed similarity matching
```
Kipok: I think it should work if we have it as an actual init.py inside a subfolder, not created on the fly by prepare.py. Will be a bit more explicit and easier to manage I think

nemo_skills/dataset/mmau-pro/prepare.py
```diff
@@ -0,0 +1,198 @@
...
+METRICS_TYPE = "speechlm"
+
+# No EVAL_ARGS - evaluation happens only during judge phase with NVEmbed
+GENERATION_ARGS = "++prompt_format=openai"
+
+# Split into 10 chunks for parallel processing of large dataset
+NUM_CHUNKS = 10
```
Kipok: I think this isn't a good default, we should always use 1. People can override this if needed. Number of chunks is very model dependent, e.g. if using a small model 10 is likely an overkill

nemo_skills/dataset/mmau-pro/prepare.py
```diff
@@ -0,0 +1,198 @@
...
+            user_message["audio"] = {"path": audio_path, "duration": 10.0}
+
+    formatted_entry["messages"] = [user_message]
+    return formatted_entry
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Prepare MMAU-Pro dataset for nemo-skills")
+    parser.add_argument("--split", default="test", choices=["validation", "test"])
+    parser.add_argument("--with-audio", action="store_true", help="Download audio files (requires HF_TOKEN)")
```
Kipok: Please add a new documentation page in docs/evaluation and add this benchmark there and describe what this parameter (with-audio) is doing and generally how the benchmark works along with some example command and reference eval scores

nemo_skills/dataset/mmau-pro/prepare.py
```diff
@@ -0,0 +1,198 @@
...
+    parser.add_argument("--download-dir", help="Directory for audio files (required with --with-audio)")
```
Kipok: can we instead reuse a common data_dir parameter of prepare_data / eval pipelines?

nemo_skills/dataset/utils.py
```diff
@@ -236,3 +236,38 @@ def get_mcq_fields(question, choices):
         "options": options_text,
         **options_dict,
     }
+
+
+def generate_subgroup_init_files(output_dir, subgroup_configs):
```
Kipok: please test if it works with just putting init.py as explicit files (not generating on the fly) and if so we can remove this new function

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+    and evaluates them using two main approaches:
+    - Closed-form questions: NVEmbed similarity matching evaluation
+    - Open-ended questions: LLM as a Judge evaluation
+    - Instruction following questions: Audio Instruction Following (AIF) format evaluation
+
+    All categories maintain separate logging to track sample counts and success rates.
+    """
+    # Extract only the fields that belong to MMAUProEvaluatorConfig
+    config_fields = {"prompt_config", "embedding_model", "use_nvembed"}
+    config_kwargs = {k: v for k, v in cfg.items() if k in config_fields}
```
Kipok: Please instead subclass from base eval config and then you shouldn't need to filter the keys

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+@nested_dataclass(kw_only=True)
+class MMAUProEvaluatorConfig:
+    """Configuration for MMAU-Pro evaluation."""
+
+    # Prompt configuration
+    prompt_config: str = "eval/speechlm/mmau-pro"
```
Kipok: is this used anywhere?

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+    # NVEmbed settings
+    embedding_model: str = "nvidia/NV-Embed-v2"
+    use_nvembed: bool = True
+
+
+def eval_mmau_pro(cfg):
+    """Evaluate MMAU-Pro dataset using nemo-skills framework.
+
+    This evaluator processes JSONL files with speech/audio language model outputs
+    and evaluates them using two main approaches:
```
Kipok: is this up-to-date? I guess it should only be using the instruction following eval path, not the other two, right?

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def evaluate_sample(sample: dict[str, Any], config: MMAUProEvaluatorConfig) -> dict[str, Any]:
+    sample = sample.copy()
+    category = sample.get("category", "unknown")
+
+    # Add subset_for_metrics only for closed_form subcategories (not for open_ended or instruction_following)
+    # This creates per-category breakdowns only for closed_form (e.g., music, speech, sound)
+    if category not in ["open", "instruction following"]:
+        sample["subset_for_metrics"] = category
```
Kipok: better to do it inside prepare.py?

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+
+    if category == "instruction following":
+        sample = evaluate_instruction_following(sample, config)
+    else:
+        choices = sample.get("choices", [])
+
+        if config.use_nvembed and choices and len(choices) > 1:
+            sample = evaluate_closed_form_with_nvembed(sample, config)
```
Kipok: is this logic still used here? Or is it fully moved to the separate judge step?

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+        return sample
+
+    generation = sample.get("generation", "").strip()
+    choices = sample.get("choices", [])
+    expected_answer = sample.get("expected_answer", "")
+
+    LOG.info(f"NVEmbed evaluation for generation='{generation[:50]}'")
+
+    if not generation or not choices:
+        LOG.warning("Missing generation or choices for nvembed evaluation")
```
Kipok: can this normally happen? If not, we should probably just fail here and let users investigate why the input is broken

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+
+        is_correct = matched_choice.strip().lower() == expected_answer.strip().lower()
+
+        sample.update(
+            {"nvembed_matched_choice": matched_choice, "nvembed_confidence": confidence, "is_correct": is_correct}
+        )
+
+        return sample
+
+    except Exception as e:
```
Kipok: I'd rather just fail here, we shouldn't do blank exceptions without a very good reason

nemo_skills/evaluation/evaluator/mmau_pro.py
Kipok: same applies to other functions inside this file. It's best to not hide any potential issues, if something isn't as expected, just let the whole job fail

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def get_overall_correctness(sample: dict[str, Any], category: str) -> bool:
+    if "judgement" not in sample:
+        return sample.get("is_correct", False)
+
+    return extract_judge_result(sample.get("judgement", ""))
+
+
+def extract_judge_result(judgement_text: str) -> bool:
```
Kipok: if this has been moved to metrics, let's remove from here. I see the same method in metrics file

nemo_skills/evaluation/metrics/speechlm_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+
+import logging
+
+from nemo_skills.evaluation.metrics.base import BaseMetrics, as_int, as_percentage
+from nemo_skills.utils import get_logger_name
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+class SpeechLMMetrics(BaseMetrics):
```
Kipok: Should we have a more specific name here? I imagine not all speech lm benchmarks will use this.

Also you don't have to use a single class for all 3 subtasks in mmau, if it's more convenient, we can have 3 different metric classes for them, but up to you

nemo_skills/evaluation/metrics/speechlm_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_incorrect_sample(self, prediction: dict) -> dict:
+        """Return a sample marked as incorrect for all metrics."""
+        prediction = prediction.copy()
+        prediction["is_correct"] = False
+        prediction["judge_correct"] = False
+        if not prediction.get("generation", "").strip():
+            prediction["generation"] = None
+        return prediction
+
+    def update_common_metrics(self, agg_dict):
```
Kipok: isn't this the same as in base class? If so, no need to redefine

nemo_skills/evaluation/metrics/speechlm_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+
+            # Set success_rate based on correct field
+            if "correct" in agg_metrics:
+                agg_metrics["success_rate"] = agg_metrics["correct"]
+            elif "judge_correct" in agg_metrics:
+                agg_metrics["success_rate"] = agg_metrics["judge_correct"]
+
+        return metrics_dict
+
+    def evaluations_to_print(self):
```
Kipok: does base method work? I think it should be equivalent

nemo_skills/evaluation/metrics/speechlm_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if self.compute_no_answer:
+            base_metrics["no_answer"] = as_percentage
+
+        base_metrics["num_entries"] = as_int  # Add at end for better display order
+
+        return base_metrics
+
+
+def compute_score(combined_metrics: dict) -> dict:
```
Kipok: better to move this to a new file with mmau_pro in the name (can also place it inside dataset folder I guess)

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
-                    dependent_tasks if cluster_config["executor"] == "slurm" else all_tasks + _task_dependencies
-                ),
-                **judge_pipeline_args,
-            )
+
+            # Check for per-benchmark judge_type, fall back to global judge_type
+            benchmark_judge_type = judge_pipeline_args.pop("judge_type", judge_type)
+
+            # creating custom command for nvembed judge type
+            if benchmark_judge_type == "nvembed":
```
Kipok, line 581: could you please try to refactor this into a function as we'd have more judge types in the future and it will be convenient if all this is handled in a separate function

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
+
+                # Check if all output files and their .done markers exist (unless rerun_done is set)
+                if not rerun_done:
+                    all_done = True
+                    for output_file in output_files:
+                        done_file = f"{output_file}.done"
+                        unmounted_output = pipeline_utils.get_unmounted_path(cluster_config, output_file)
+                        unmounted_done = pipeline_utils.get_unmounted_path(cluster_config, done_file)
+
+                        # Check if files exist
```
Kipok: Can you please try to reuse the method fromnemo_skills/pipeline/utils/generation.py `get_remaining_jobs`? If not reusable directly we can refactor the common piece of checking for .done files from there?

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
+                                break
+
+                    if all_done:
+                        LOG.info(f"Skipping NVEmbed judge for {benchmark} - all output files and .done markers exist")
+                        judge_tasks = None
+                        continue
+
+                # Install required packages for NVEmbed evaluation
+                install_cmd = (
+                    "pip install -q -e /nemo_run/code && pip install -q datasets einops transformers==4.42.4 && "
```
Kipok: do we need `pip install -q -e /nemo_run/code` ? the repo might not be installable (as it's not always that this repo is being used, users can run this from inside another git repo and then we upload their repo instead). If you just cd into that folder hopefully everything should work without installation as there will always be a `nemo_skills` subfolder there

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
+                    files_to_process.append((input_file, f"{output_dir_path}/output.jsonl"))
+
+                # Create evaluation commands for all files
+                # Check generation's .done file first to ensure all chunks were merged successfully
+                eval_cmds = []
+                for src_file, dst_file in files_to_process:
+                    eval_cmd = (
+                        # First verify generation completed by checking for .done file
+                        f'if [ ! -f "{src_file}.done" ]; then '
+                        f'  echo "Error: Generation not complete: {src_file}.done not found"; '
```
Kipok: I don't think we need an explicit check, can just fail with normal error of not finding an input file in case previous job isn't finished. So I'd probably just run your command assuming it's there and if not it should be clear from error message that previous step has an issue

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
+                        f"  exit 1; "
+                        f"fi && "
+                        # Copy and evaluate only if generation succeeded
+                        f"mkdir -p {output_dir_path} && "
```
Kipok: consider moving all of this logic into a separate script and then just calling that script (including installation commands) so that we don't need to have a lot of logic build-up in the general eval.py but can just do something like `python -m nemo_skills.evaluation.evaluator.mmau_pro_nvembed_judge <parameters>` and all the rest is handled inside or something like this

nemo_skills/pipeline/eval.py
```diff
@@ -409,46 +410,160 @@ def eval(
...
+
+                run_cmd = f"{install_cmd} && {' && '.join(eval_cmds)}"
+
+                # Create task with GPU support for NVEmbed
+                judge_task = pipeline_utils.add_task(
+                    exp,
+                    cmd=run_cmd,
+                    task_name=f"{expname}-{benchmark}-nvembed-judge",
+                    log_dir=log_dir + "/judge",
+                    container=server_parameters.get("server_container"),
```
Kipok: should we pick a specific container instead? Users can provide their own containers and it's not guaranteed that they will have all dependencies you need. Maybe we just use vllm e.g. if that's what you tested this with irrespective of main server (or any other container, but I'd probably have it fixed)

Kipok: even better if we can use main nemo-skills client container as it's the most lightweights (maybe with extra on-the-fly install of whatever you need) unless it takes too much time

nemo_skills/evaluation/metrics/speechlm_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class SpeechLMMetrics(BaseMetrics):
+    """Metrics class for speech/audio language model evaluation tasks like MMAU-Pro."""
+
+    def __init__(self, compute_no_answer: bool = True, max_k: int = 1):
+        super().__init__(compute_no_answer=compute_no_answer)
+        self.max_k = max_k
+
+    def _extract_judge_result(self, judgement_text: str) -> bool:
```
Kipok: can we reuse the existing is_correct_judgement function here?

nemo_skills/pipeline/eval.py
melllinia: At first, I tried to use the main nemo-skills container, but since nv-embed requires torch, I needed to install it there. However even after installing the correct cuda-compatible version of torch, cuda still shows as unavailable.

melllinia: In the recent changes I switched to the vllm container since it already had `torch` installed, and then installed the required packages there.

nemo_skills/evaluation/evaluator/mmau_pro.py
melllinia: As you advised, I moved nvembed step completely to the new file `nvembed_judge.py` and calling it from `eval.py`. 
`mmau_pro.py` is currently handling only Instruction Following task.

docs/evaluation/index.md
```diff
@@ -6,6 +6,7 @@ We support many popular benchmarks and it's easy to add new in the future. The f
 - [**Math (formal language)**](./formal-math.md): e.g. [minif2f](./formal-math.md#minif2f), [proofnet](./formal-math.md#proofnet), [putnam-bench](./formal-math.md#putnam-bench)
 - [**Code**](./code.md): e.g. [swe-bench](./code.md#swe-bench), [livecodebench](./code.md#livecodebench)
 - [**Scientific knowledge**](./scientific-knowledge.md): e.g., [hle](./scientific-knowledge.md#hle), [scicode](./scientific-knowledge.md#scicode), [gpqa](./scientific-knowledge.md#gpqa)
+- [**Speech and Audio Understanding**](./speech-audio.md): e.g. [mmau-pro](./speech-audio.md#mmau-pro)
```
Kipok: Please also add to https://github.com/NVIDIA-NeMo/Skills/blob/main/README.md?plain=1#L17

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+#### Dataset Location
+
+- Benchmark is defined in [`nemo_skills/dataset/mmau-pro/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/mmau-pro/__init__.py)
+- Original benchmark source is hosted on [HuggingFace](https://huggingface.co/datasets/gamma-lab-umd/MMAU-Pro)
+
+## Preparing MMAU-Pro Data
+
+MMAU-Pro requires audio files for evaluation. The `--with-audio` parameter controls whether audio files are downloaded.
+
+> **Note:** Meaningful evaluation requires audio files and audio-capable models. Text-only data preparation is possible but not recommended.
```
Kipok: If that's the case, how about we make this a default behavior, and then add a `--no-audio` option to opt-out and use text only?

Kipok: for things like this it's best to use mkdocs notation (e.g. `!!! note:`) and same for warning below. Please check other doc files for an example. Also please make sure to run `mkdocs serve` and inspect that doc rendering looks good locally

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Running Evaluation
+
+> **⚠️ Warning:** MMAU-Pro evaluation currently supports only the Megatron server type (`--server_type=megatron`) with audio-capable models. Support for additional server types is planned for future releases.
+
+### Evaluation Example
+
+Complete example using Megatron-based audio-visual language models:
+
+```bash
```
Kipok: would it be better to have a Python-api example instead? If you need so many variables defined, using python is almost always a better way, so we should showcase that as an example

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+export SERVER_CONTAINER="/path/to/server_container.sqsh"
+export OUTPUT_DIR="mmau-pro-eval"
+
+# Set up keys
+export HF_TOKEN='your_huggingface_token'
+export NVIDIA_API_KEY='your_nvidia_api_key'
+export WANDB='your_wandb_key'   # not neccessary
+
+# Run evaluation with audio support
+export HF_TOKEN=${HF_TOKEN} && \
```
Kipok: these 3 lines are redundant and shouldn't be necessary as you already export above

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+
+> **⚠️ Warning:** MMAU-Pro evaluation currently supports only the Megatron server type (`--server_type=megatron`) with audio-capable models. Support for additional server types is planned for future releases.
+
+### Evaluation Example
+
+Complete example using Megatron-based audio-visual language models:
+
+```bash
+# Set up environment variables - ADJUST THESE PATHS TO YOUR SETUP
+export MEGATRON_PATH="/workspace/path/to/megatron-lm"
```
Kipok: does this require a custom megatron container? If it's public, we should also probably update our default container in dockerfiles/Dockerfile.megatron to ensure things work out-of-the-box by default

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+export CKPT_PATH=/workspace/path/to/checkpoint-tp1
```
Kipok: is there some public model we can use for a real example?

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Overall Aggregate Score:**
+
+```
+-------------------------------- mmau-pro -----------------------------------------
+evaluation_mode | avg_tokens | gen_seconds | success_rate | no_answer | num_entries
+pass@1          | 11         | 6879        | 31.44%       | 0.00%     | 5305
+```
+
+## Tips
```
Kipok: I'd remove this section

docs/evaluation/speech-audio.md
```diff
@@ -0,0 +1,215 @@
...
+export MEGATRON_PATH="/workspace/path/to/megatron-lm"
+export CKPT_PATH=/workspace/path/to/checkpoint-tp1
+export MODEL_CFG_PATH="${CKPT_PATH}/config.yaml"
+export SERVER_ENTRYPOINT="$MEGATRON_PATH/path/to/server.py"
+export SERVER_CONTAINER="/path/to/server_container.sqsh"
+export OUTPUT_DIR="mmau-pro-eval"
+
+# Set up keys
+export HF_TOKEN='your_huggingface_token'
+export NVIDIA_API_KEY='your_nvidia_api_key'
```
Kipok: one thing that could be useful to add is an example of how to run eval with a self-hosted judge model. So ideally you can show two commands, one default and another with self-hosted, you can use mkdocs tab interface for that like in here https://github.com/NVIDIA-NeMo/Skills/blob/main/docs/evaluation/natural-math.md?plain=1#L36

Kipok: there could potentially be an issue with that, not fully sure. If you see that updating judge parameters from command line makes it run judge on all 3 subsets please let me know and we should find a way to fix that

nemo_skills/dataset/mmau-pro/mmau_pro_score.py
```diff
@@ -0,0 +1,63 @@
...
+        for benchmark_name, benchmark_data in mmau_benchmarks.items():
+            if eval_mode not in benchmark_data:
+                continue
+
+            metrics = benchmark_data[eval_mode]
+            num_entries = metrics.get("num_entries", 0)
+            total_entries += num_entries
+
+            # Aggregate weighted by number of entries (metrics are already percentages)
+            if num_entries > 0:
```
Kipok: we shouldn't need this if and also please replace all .get with just [] - we want to job to fail if there is something unexpected, not hide the issue and report 0 metrics. If some of the parameters are indeed optional, we can use .get, but for anything that should always be there we should not

nemo_skills/dataset/mmau-pro/prepare.py
```diff
@@ -0,0 +1,162 @@
...
+
+    formatted_entry["messages"] = [user_message]
+    return formatted_entry
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Prepare MMAU-Pro dataset for nemo-skills")
+    parser.add_argument("--split", default="test", choices=["validation", "test"])
+    parser.add_argument("--with-audio", action="store_true", help="Download audio files (requires HF_TOKEN)")
+    parser.add_argument(
```
Kipok, line 100: you shouldn't need an extra argument for this. The main ns prepare_data has `data_dir` argument and it should handle all logic related to this directly. It might also make sense to add this benchmark to a list of the ones requiring data_dir in here https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/pipeline/prepare_data.py#L34

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,169 @@
...
+from tqdm import tqdm
+
+from nemo_skills.evaluation.evaluator.base import BaseEvaluatorConfig
+from nemo_skills.utils import get_logger_name, nested_dataclass
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+@nested_dataclass(kw_only=True)
+class MMAUProEvaluatorConfig(BaseEvaluatorConfig):
```
Kipok: if no parameters, you don't need a class, just use BaseEvaluatorConfig directly on line 43

nemo_skills/evaluation/evaluator/mmau_pro.py
```diff
@@ -0,0 +1,169 @@
...
+    generation = sample.get("generation", "").strip()
+
+    if not generation:
+        LOG.info("Empty generation detected for instruction following question")
+        sample["is_correct"] = False
+        sample["error"] = "empty_generation"
+        return sample
+
+    success = evaluate_aif_constraints(
+        generation, sample.get("task_identifier", ""), sample.get("kwargs", {}) or {}, sample
```
Kipok: again check if things are indeed optional or not. E.g. task_identifier looks like it might be required, then let's not use .get

nemo_skills/evaluation/metrics/base.py
```diff
@@ -27,8 +27,7 @@ def __init__(self, compute_no_answer: bool = True):
 
     def update_common_metrics(self, agg_dict):
         agg_dict["num_entries"] = self.total
-        if self.avg_tokens > 0:
-            agg_dict["avg_tokens"] = int(self.avg_tokens / self.total)
+        agg_dict["avg_tokens"] = int(self.avg_tokens / self.total) if self.total > 0 else 0
```
Kipok: I think we should also keep the original condition as tokens might be 0 if they aren't reported in data file and we shouldn't expose as metric in that case

docs/evaluation/speech-audio.md
melllinia: Yes, while for closed-form questions it submits the `nvembed-judge` correctly, for instruction following task it submits unnecessary llm-judge job. 
Can we just mention about this in doc and warn users that in self-hosted judge case they need to submit `mmau-pro.open_ended` benchmark separately?

nemo_skills/dataset/mmau-pro/prepare.py
melllinia: I think adding the benchmark to the list of requiring data_dir will not let the users to use `--no-audio` option.

docs/evaluation/speech-audio.md
melllinia: I asked the team, and they said that our model is not published yet, so unfortunately we don't have corresponding container either.

nemo_skills/evaluation/metrics/base.py
melllinia: Some subgroups in the closed-form group have an average token count of 0 (likely due to a model issue), which caused the job to fail when reporting final metrics. I reverted this change in `base.py` and added logic in `mmau_pro_metrics.py` to handle cases where avg_tokens is 0. Please confirm if this approach looks good.

docs/evaluation/speech-audio.md
Kipok: Yes, let's do this for now

nemo_skills/dataset/mmau-pro/prepare.py
Kipok: Yeah, could you also add an option `--skip_data_dir_check` to prepare data to bypass that? So by default it does fail to avoid accidentally putting 50gb in main repo and uploading on cluster, but in the error message we can say that if they want to bypass this, they can set that parameter

nemo_skills/evaluation/metrics/base.py
Kipok: looks good, thanks!

PR #1025, opened 2025-11-03
adding extra_metrics_arguments to pass extra arguments to the metrics…

mehrzads:
Adding extra_metrics_arguments that can be passed to summarize and eval functions. By default it does not change any functionalities

### DIFF
--- a/nemo_skills/pipeline/utils/cluster.py
+++ b/nemo_skills/pipeline/utils/cluster.py
@@ -106,7 +106,7 @@ def get_timeout_str(cluster_config, partition, with_save_delay: bool = True) -> 
     return timeout_str
 
 
-def kwargs_to_string(kwargs: str | dict | None) -> dict | None:
+def kwargs_to_string(kwargs: str | dict) -> dict:
     """
     Convert kwargs to a string.
     """


nemo_skills/pipeline/utils/cluster.py
```diff
@@ -106,6 +106,48 @@ def get_timeout_str(cluster_config, partition, with_save_delay: bool = True) ->
     return timeout_str
 
 
+def parse_extra_metrics_arguments(extra_metrics_arguments: str | dict | None) -> dict | None:
```
Kipok: let's maybe have a shared function that will be used both here and for sbatch kwargs (and any other future use-cases where we need to take a dict as an argument)? In fact there is already another one with `extra_judge_pipeline_args` in eval.py which we can use this function for

nemo_skills/pipeline/eval.py
```diff
@@ -214,6 +214,10 @@ def eval(
         "",
         help="Additional sbatch kwargs to pass to the job scheduler. Values should be provided as a JSON string or as a `dict` if invoking from code.",
     ),
+    extra_metrics_arguments: str = typer.Option(
```
Kipok: maybe we can call it extra_metrics_kwargs (or just metrics_kwargs)? To distinguish between arguments that are just string pass-through (called args/arguments) vs arguments that take in a dict (called kwargs)

nemo_skills/pipeline/eval.py
```diff
@@ -482,6 +486,8 @@ def eval(
                     command += f" --wandb_project={wandb_project} "
                 if data_dir:
                     command += f" --data_dir={data_dir} "
+                if extra_metrics_arguments:
+                    command += f" --extra_metrics_arguments='{extra_metrics_arguments}' "
```
Kipok: I guess we might need to call json.dumps if type is dict?

nemo_skills/evaluation/metrics/map_metrics.py
```diff
@@ -63,7 +63,7 @@
 }
 
 
-def get_metrics(metric_type: str):
+def get_metrics(metric_type: str, extra_arguments: dict = None):
```
Kipok: maybe just take it as **kwargs here? That would be consistent with other get_* functions we used throughout the codebase

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -106,16 +106,27 @@ def get_timeout_str(cluster_config, partition, with_save_delay: bool = True) ->
...
+def kwargs_to_string(kwargs: str | dict | None) -> dict | None:
     """
-    Parse sbatch kwargs from either a JSON string or a dictionary.
+    Convert kwargs to a string.
+    """
+    if isinstance(kwargs, dict):
+        return json.dumps(kwargs)
+    elif isinstance(kwargs, str):
+        return kwargs
+    return None
```
Kipok: should we fail with error here?

nemo_skills/pipeline/utils/cluster.py
```diff
@@ -124,28 +135,28 @@ def parse_sbatch_kwargs(sbatch_kwargs: str | dict | None, **kwargs) -> dict | No
         A dictionary of slurm kwargs, or None if no arguments are provided.
```
Kipok:
```suggestion
        A dictionary containing kwargs, or None if no arguments are provided.
```

nemo_skills/pipeline/utils/cluster.py
Kipok: sorry, I guess this one is good as we can assume this is checked in downstream code. We can revise it in the future, but should be good for now

PR #1026, opened 2025-11-03
Add SSH tunnel to `start_server` pipeline

activatedgeek:
TODO:
- [x] Verify the flexibility of the SSH tunnel command in a completely fresh node without any interfering SSH config.

### DIFF
--- a/docs/pipelines/start-server.md
+++ b/docs/pipelines/start-server.md
@@ -16,7 +16,7 @@ ns start_server \
     --log_dir=/workspace/logs/start_server \
 ```
 
-Now, the model server is available at `localhost:5000` by default.
+Now, the model server is available at `localhost:5000` by default. Note that you can also use `Qwen/Qwen3-4B-Thinking-2507` directly as the model name if you don't have it pre-downloaded.
 
 To launch a sandbox server in parallel, provide the `--with_sandbox` argument, which will be
 available at `localhost:6000` by default.


nemo_skills/pipeline/start_server.py
```diff
@@ -38,6 +48,51 @@ def get_gradio_chat_cmd(model, server_type, extra_args):
     return cmd
 
 
+def create_server_tunnel(exp: Experiment, server_port: int, wait_interval: int = 10):
+    ## Assumes first job in experiment correspond to the server.
```
Kipok:
```suggestion
    ## Assumes first job in experiment corresponds to the server.
```

nemo_skills/pipeline/start_server.py
```diff
@@ -132,9 +188,25 @@ def start_server(
...
-        exp.run(detach=False, tail_logs=True)
-        # TODO: seems like not being killed? If nemorun doesn't do this, we can catch the signal and kill the server ourselves
+
+        exp.run(detach=True, tail_logs=True)
+
+        ## NOTE: Use ctrl + c twice to cancel all experiment jobs.
+        signal.signal(signal.SIGINT, signal.default_int_handler)
+        try:
+            tunnel_proc = None
+            if create_tunnel:
```
Kipok, line 219: let's fail with clear message if executor isn't slurm

nemo_skills/pipeline/start_server.py
```diff
@@ -132,9 +188,25 @@ def start_server(
...
+                tunnel_proc = create_server_tunnel(exp, server_config["server_port"])
```
Kipok: we should do the same for sandbox port

nemo_skills/pipeline/start_server.py
```diff
@@ -38,6 +48,51 @@ def get_gradio_chat_cmd(model, server_type, extra_args):
...
+        "-N",
+        "-o",
+        "StrictHostKeyChecking=accept-new",
+        "-J",
+        f"{job.executor.tunnel.user}@{job.executor.tunnel.host}",
+        job_node_list[0],
+        "-L",
+        f"{server_port}:localhost:{server_port}",
+    ]
+    LOG.info(f"SSH tunnel command: {' '.join(ssh_tunnel_args)}")
```
Kipok, line 105: would be nice to print a message for the final host/port for the server to access locally. host is always localhost, but port can be random, so would be nice to print here so that it's easy to know where to connect. Can even add what exactly should be set in ++server_address argument e.g. Alternatively we can accept server/sandbox local port to map to

nemo_skills/pipeline/start_server.py
activatedgeek: done

nemo_skills/pipeline/start_server.py
activatedgeek: The `tail_logs=True` actually generates many lines of logs that this message will effectively disappear. But I've added arguments for `server_tunnel_port` and `sandbox_tunnel_port`.

activatedgeek: I've added these details to the docs.

docs/pipelines/start-server.md
```diff
@@ -0,0 +1,47 @@
+# Launching a Remote Model Server
+
+!!! info
+
+    This pipeline starting script is [nemo_skills/pipeline/start_server.py](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/pipeline/start_server.py)
+
```
Kipok, line 6: Can you provide a local example first please so that people know they can use it to spin up server locally not just on slurm? And then maybe we put the tunnel one into a separate subsection?

docs/pipelines/start-server.md
```diff
@@ -0,0 +1,47 @@
...
+## Sandbox Server
+
+To launch a sandbox server, provide the `--with_sandbox` argument.
+
+When the `--create_tunnel` argument is set, the sandbox server is available at `localhost:6000` by default.
+
+## Remote and Tunnel Ports
+
+To avoid port conflicts on the remote hosts, use `--get_random_port` to randomly assign ports to launched server.
```
Kipok: Please mention that this is only necessary if using partial node jobs (e.g. requesting 4 gpus for slurm clusters with 8 gpus per node)

docs/pipelines/start-server.md
```diff
@@ -0,0 +1,47 @@
+# Launching a Remote Model Server
+
+!!! info
+
+    This pipeline starting script is [nemo_skills/pipeline/start_server.py](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/pipeline/start_server.py)
+
+This pipeline provides a convenient way to host models remotely on a Slurm cluster and access them locally.
```
Kipok: Another thing that can be very useful to describe is that instead of doing `ns generate` or `ns eval` with num_gpus parameter, people can do start server first and then change eval/generate to use server_address to avoid spinning up server every time when doing development. Can also mention that can remove cluster parameter to even avoid launching containers for even faster iteration (assuming all deps are installed)

mkdocs.yml
```diff
@@ -70,6 +70,7 @@ nav:
     - Training (NeMo-RL): pipelines/training.md
     - Training (verl): pipelines/training-verl.md
     - Arbitrary commands: pipelines/run-cmd.md
+    - Remote Server: pipelines/start-server.md
```
Kipok: Maybe `Starting a Server` ? It doesn't have to be remote

docs/pipelines/start-server.md
activatedgeek: Ah yes, forgot about local. Testing these commands and updating docs.

docs/pipelines/start-server.md
activatedgeek: Testing the `server_address` argument locally.

docs/pipelines/start-server.md
activatedgeek: Updated.

docs/pipelines/start-server.md
activatedgeek: Added an example with `ns eval`.

docs/pipelines/start-server.md
```diff
@@ -0,0 +1,76 @@
...
+```bash
+ns start_server \
+    --cluster=local \
+    --server_type=vllm \
+    --model=/workspace/models/Qwen3-4B-Thinking-2507 \
+    --server_gpus=1 \
+    --log_dir=/workspace/logs/start_server \
+```
+
+Now, the model server is available at `localhost:5000` by default.
```
Kipok:
```suggestion
Now, the model server is available at `localhost:5000` by default. Note that you can also use `Qwen/Qwen3-4B-Thinking-2507` directly as the model name if you don't have it pre-downloaded.
```

docs/pipelines/start-server.md
```diff
@@ -0,0 +1,76 @@
...
+
+To launch a sandbox server in parallel, provide the `--with_sandbox` argument, which will be
+available at `localhost:6000` by default.
+
+## Remote Server
+
+It is also possible to remotely start the server on a slurm cluster and access it locally.
+This approach is especially useful for quick debugging or when your local workstation does not have all the required compute resources available.
+
+Consider a vLLM model at path `/workspace/models/Qwen3-4B-Thinking-2507` that requires 16 GPUs.
```
Kipok:
```suggestion
Consider a vLLM model at path `/workspace/models/Qwen3-235B-A22B-Thinking-2507` that requires 16 GPUs.
```

PR #1029, opened 2025-11-04
ENH skip scicode data download in CI

gwarmstrong:


dockerfiles/Dockerfile.sandbox
```diff
@@ -62,9 +62,15 @@ ENV PATH="/lean4/my_project:$PATH"
...
-RUN mkdir /data && pip install gdown && \
-    python -c "import gdown; url = f'https://drive.google.com/uc?id=17G_k65N_6yFFZ2O-jQH00Lh6iaw3z-AW'; gdown.download(url, '/data/test_data.h5', quiet=False)"
+# For scicode eval - create data directory and download test data
+# Set SKIP_SCICODE_DATA_DOWNLOAD=1 build arg to skip download (useful for CI when download fails)
+# If skipped, data will be downloaded at runtime on first use
+ARG SKIP_SCICODE_DATA_DOWNLOAD=0
+RUN mkdir -p /data && pip install gdown && \
+    if [ "$SKIP_SCICODE_DATA_DOWNLOAD" != "1" ]; then \
+        python -c "import gdown; url = 'https://drive.google.com/uc?id=17G_k65N_6yFFZ2O-jQH00Lh6iaw3z-AW'; gdown.download(url, '/data/test_data.h5', quiet=False)" || \
+        echo "Warning: Failed to download scicode test data during build. Will retry at runtime."; \
```
Kipok: one thing to consider is to just fail here. Users can enable the arg if they want, but it might be hard to notice that it was skipped and could result in a container without this baked in which might not be what people want

nemo_skills/evaluation/evaluator/scicode.py
```diff
@@ -125,6 +125,35 @@ async def install_packages():
...
+            )
+            result, _ = await sandbox.execute_code(download_cmd, language="shell", timeout=300.0)
+
+            if result["process_status"] != "completed":
+                error_msg = result.get("stderr", result.get("stdout", "Unknown error"))
+                LOG.error(f"Failed to download scicode test data: {error_msg}")
+                LOG.error(
+                    "You can manually download the file from "
+                    "https://drive.google.com/file/d/17G_k65N_6yFFZ2O-jQH00Lh6iaw3z-AW/view "
+                    "and mount it to the sandbox at /data/test_data.h5"
```
Kipok:
```suggestion
                    "and mount it to the sandbox at /data/test_data.h5. "
                    "If you do this, rerun with `--keep_mounts_for_sandbox`"
```

dockerfiles/Dockerfile.sandbox
gwarmstrong: yeah I think if our published containers are still getting the data downloaded, then this is overkill. I'll set it to fail here.

PR #1031, opened 2025-11-05
ENH allow generation passthrough of hydra args and config

gwarmstrong:
# Add Config File Support for NeMo-Skills Pipelines

## Problem

When using `ns generate` and other NeMo-Skills pipelines, certain configuration parameters with special characters (like `end_reasoning_string='</think>'`) cause shell escaping issues. This leads to errors like:

```
hydra.errors.OverrideParseException: LexerNoViableAltException: ++end_reasoning_string=\</think\>
```

## Solution

This PR adds support for using YAML config files with NeMo-Skills pipelines, enabling a users to use a combination of config parameters and command-line overrides.

## Changes

* Had to update the hydra command construction to move all the hydra arguments to the front of the command
* Documentation update for using config files
* Tests for the hydra argument re-ordering utilities

## Usage Example

**Before** (problematic):
```bash
ns generate \
    --cluster=slurm \
    ... \
    ++end_reasoning_string='"\"</think>\""'
```

**After** (clean):

1. Create `/workspace/configs/reasoning_config.yaml`:
```yaml
end_reasoning_string: '</think>'
```

2. Use it:
```bash
ns generate \
    --cluster=slurm \
    ..
    --config-path=/workspace/configs \
    --config-name=reasoning_config \
```

Or in Python:
```python
from nemo_skills.pipeline.cli import generate, wrap_arguments

generate(
    wrap_arguments(
        "--config-path /workspace/configs",
        "--config-name reasoning_config",
    ),
    cluster="local",
    server_type="openai",
    model="meta/llama-3.1-8b-instruct",
    server_address="https://integrate.api.nvidia.com/v1",
    output_dir="/workspace/test-generate",
    input_file="/workspace/input.jsonl",
)
```

### DIFF
--- a/docs/pipelines/index.md
+++ b/docs/pipelines/index.md
@@ -192,13 +192,16 @@ parallel_thinking:
         wrap_arguments(
             "--config-path /workspace/configs "
             "--config-name reasoning_config "
+            "++prompt_config=generic/math-base "
+            "++inference.temperature=0.7 "
+            "++inference.tokens_to_generate=2048 "
         ),
-        cluster="local",
-        server_type="openai",
-        model="meta/llama-3.1-8b-instruct",
-        server_address="https://integrate.api.nvidia.com/v1",
-        output_dir="/workspace/test-generate",
-        input_file="/workspace/input.jsonl",
+        cluster="slurm",
+        server_type="vllm",
+        model="Qwen/QwQ-32B-Preview",
+        server_gpus=4,
+        output_dir="/workspace/reasoning-output",
+        input_file="/workspace/math-problems.jsonl",
     )
     ```
 


docs/pipelines/generation.md
```diff
@@ -464,4 +464,82 @@ We support three methods for automatic trimming of generation budget or context:
     ```bash
         ++server.enable_soft_fail=True
         ++server.context_limit_retry_strategy=reduce_prompt_from_end
-    ```
\ No newline at end of file
+    ```
+
+### Using Config Files
```
Kipok: Can you please move it to docs/pipelines/index.md? That's probably a better place since this is applicable to most pipelines. At least to both generate and eval, so I'd put it there

nemo_skills/pipeline/utils/generation.py
```diff
@@ -175,6 +175,112 @@ def get_remaining_jobs(cluster_config, output_dir, random_seeds, chunk_ids, reru
...
+        "--config-path",
+        "--config-name",
+        "--config-dir",
+        "--package",
+        "--experimental-rerun",
+        "--cfg",
+        "--info",
+    }
+
+    args_parts = extra_arguments.split()
```
Kipok: is this reliable? I imagine we could have cases when one of the parameter values might have a space inside? Not fully sure if this is going to cause problems

gwarmstrong: good point. maybe `shlex.split` is more appropriate?

Kipok: might be, as long as it works correctly

docs/pipelines/generation.md
gwarmstrong: done in 28ef504

nemo_skills/pipeline/utils/generation.py
gwarmstrong: seems fine, have pretty extensive tests for it--can always update and add new tests for the utility if we find a syntax isn't working well.

docs/pipelines/index.md
```diff
@@ -126,3 +126,80 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+```
+
+Common error:
+```
+hydra.errors.OverrideParseException: LexerNoViableAltException: ++end_reasoning_string=\</think\>
+```
+
+### The Solution
+
+**1. Create a config file** (`/workspace/reasoning_config.yaml`):
```
Kipok: Maybe we should go with `/nemo_run/code` instead of /workspace, so that this can be packaged and uploaded? Feels like a more common use-case instead of keeping it as a standalone file. Not fully sure, but at least could mention that as an option (and add reference to packaging docs to ensure people commit this when applicable)

docs/pipelines/index.md
```diff
@@ -126,3 +126,80 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+    --config-name=reasoning_config \
+    ++prompt_config=generic/math-base \
+    ++inference.temperature=0.7 \
+    ++inference.tokens_to_generate=2048
+```
+
+**How it works:**
+- `--config-path=/workspace/configs`: Directory containing your config file
+- `--config-name=reasoning_config`: Config filename without `.yaml` extension
+- Command-line `++` args can still override config file values if needed
```
Kipok: Maybe add a link to hydra docs where people can learn more

docs/pipelines/index.md
```diff
@@ -126,3 +126,80 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+- This works with any pipeline script with generation (`ns generate`, `ns eval`, etc.)
+
+### Python API
+
+When using the Python API, leverage config files with `wrap_arguments`:
```
Kipok: Maybe better to just have a tabular switch between the two, so that it doesn't require a separate section and doesn't take space on the page, similar to how we do it here https://github.com/NVIDIA-NeMo/Skills/blob/main/docs/evaluation/natural-math.md?plain=1#L36

(I even created an issue to duplicate all commands throughout docs with both cli and python https://github.com/NVIDIA-NeMo/Skills/issues/984)

docs/pipelines/index.md
```diff
@@ -126,3 +126,80 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+
+```python
+from nemo_skills.pipeline.cli import generate, wrap_arguments
+
+# Assume config is mounted to /workspace/configs/reasoning_config.yaml
+generate(
+    wrap_arguments(
+        "--config-path /workspace/configs ",
```
Kipok:
```suggestion
        "--config-path /workspace/configs "
```

docs/pipelines/index.md
```diff
@@ -126,3 +126,80 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+        "--config-name reasoning_config ",
```
Kipok:
```suggestion
        "--config-name reasoning_config "
```

docs/pipelines/generation.md
```diff
@@ -464,4 +464,10 @@ We support three methods for automatic trimming of generation budget or context:
     ```bash
         ++server.enable_soft_fail=True
         ++server.context_limit_retry_strategy=reduce_prompt_from_end
-    ```
\ No newline at end of file
+    ```
+
+!!! tip "Using Config Files"
```
Kipok: Let's move this higher, currently it's hard to find. Maybe right below #run-generation section? Also would be good to duplicate inside evaluation docs as well as it's maybe even more relevant there

gwarmstrong: done in 071dc56

docs/pipelines/index.md
gwarmstrong: done in c0e6eb5

docs/pipelines/index.md
gwarmstrong: done in b07d3b6

docs/pipelines/index.md
gwarmstrong: done in b207578 
linked to packaging in: 9578e90

docs/pipelines/index.md
```diff
@@ -126,3 +126,85 @@ All of the scripts that involve LLM data generation accept a common set of param
...
+        --output_dir=/workspace/reasoning-output \
+        --input_file=/workspace/math-problems.jsonl \
+        --config-path=/nemo_run/code/main_arg_configs \
+        --config-name=reasoning_config \
+        ++prompt_config=generic/math-base \
+        ++inference.temperature=0.7 \
+        ++inference.tokens_to_generate=2048
+    ```
+
+=== "python interface"
```
Kipok, line 186: I guess would be a little better if we use exactly the same parameters for both, so that people can see how things translate from cmdline to python

gwarmstrong: good idea, added in aefcc8a

PR #1041, opened 2025-11-13
Enable parse_reasoning by default for SciCode

jubick1337:


gwarmstrong (approved):
LGTM once all checks pass

nemo_skills/inference/eval/scicode.py
```diff
@@ -117,6 +121,17 @@ async def process_single_datapoint(self, data_point, all_data):
             total_generated_tokens += llm_output.get("num_generated_tokens", 0)
             if self.cfg.parse_reasoning:
                 parse_reasoning(llm_output, "generation", self.cfg.end_reasoning_string)
+            
+            # Warn once on first generation if reasoning detected but not being parsed
```
Kipok: could also consider doing this in main generate module as this is something that's probably useful for any benchmark / generation, not just scicode (although for scicode it's particularly critical of course)

nemo_skills/inference/generate.py
```diff
@@ -543,6 +546,16 @@ async def postprocess_single_output(self, output, original_data_point):
                 self.cfg.end_reasoning_string,
             )
 
+        # Warn once if reasoning detected but not being parsed
```
gwarmstrong, line 549: I might suggest moving this into the __init__ and then we can simplify the logic a bit, right? `_reasoning_warning_shown` flag can be dropped. Right? Not sure if this check actually needs to be dependent on a real generation--isn't the fact that `parse_reasoning` is not set and `end_reasoning_string` is set enough to warrant showing the warning?

jubick1337: Good idea, but there's a tricky bit: `end_reasoning_string` defaults to `</think>`, so it's always set even for non-reasoning models. If we checked in `__init__`, we'd warn on every run regardless of whether reasoning is actually being used.
Current approach waits to see if reasoning tokens actually show up in the output, so we only warn when there's a real issue. The flag just makes sure we don't spam it on every generation.
Let me know if you see a better way to handle this!

gwarmstrong: Ah I see. Yeah then we would have to start changing defaults to get this working properly--less than ideal. I think this check will still miss cases where `end_reasoning_string` is misconfigured, but it's probably about as good as we can do.

PR #1042, opened 2025-11-13
Request to vLLM with audio message 

karpnv:
## Release Notes

* **New Features**
  * Added audio transcription and evaluation support with multicriteria scoring for correctness, relevance, completeness, and clarity.
  * Audio content encoding for model inference requests.

* **Documentation**
  * Expanded audio evaluation guides with server setup examples and evaluation workflows.
  * Updated result formats and benchmark descriptions.

* **Tests**
  * Added audio generation and evaluation test coverage.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -406,12 +406,11 @@ class GenerationTask:
 
         # Audio wrapper (preprocesses messages before they reach the model)
         # Auto-enable for audio benchmarks on vLLM (eval_type=audio OR dataset_group=speechlm)
-        should_enable_audio = (
-            self.cfg.audio is not None or 
-            (self.cfg.server.get("server_type", "").lower() == "vllm" and 
-             (self.cfg.eval_type == "audio" or self.cfg.dataset_group == "speechlm"))
+        should_enable_audio = self.cfg.audio is not None or (
+            self.cfg.server.get("server_type", "").lower() == "vllm"
+            and (self.cfg.eval_type == "audio" or self.cfg.dataset_group == "speechlm")
         )
-        
+
         if should_enable_audio:
             audio_supported_servers = {"vllm"}
             server_type = self.cfg.server.get("server_type", "").lower()
@@ -420,10 +419,10 @@ class GenerationTask:
                     f"Audio processing is not supported for server_type='{server_type}'. "
                     f"Supported server types: {audio_supported_servers}"
                 )
-            
+
             # Use provided config or create default
             audio_config = self.cfg.audio if self.cfg.audio is not None else AudioProcessorConfig()
-            
+
             llm = AudioProcessor(
                 llm,
                 audio_config,
--- a/nemo_skills/pipeline/utils/eval.py
+++ b/nemo_skills/pipeline/utils/eval.py
@@ -131,12 +131,12 @@ def get_benchmark_args_from_module(
     if eval_args:
         generation_args = f"{eval_args} {generation_args}"
     generation_args += f" ++eval_config.split={split} "
-    
+
     # Pass dataset_group from benchmark config if defined
     dataset_group = get_arg_from_module_or_dict(benchmark_module, "DATASET_GROUP", None, override_dict=override_dict)
     if dataset_group:
         generation_args += f" ++dataset_group={dataset_group} "
-    
+
     requires_sandbox = get_arg_from_module_or_dict(benchmark_module, "REQUIRES_SANDBOX", False, override_dict)
     keep_mounts_for_sandbox = get_arg_from_module_or_dict(
         benchmark_module, "KEEP_MOUNTS_FOR_SANDBOX", False, override_dict


nemo_skills/inference/model/vllm.py
```diff
@@ -24,8 +26,16 @@
...
+    """Encodes an audio file into a base64 string."""
+    with open(audio_file_path, "rb") as audio_file:
+        audio_content = audio_file.read()
+        return base64.b64encode(audio_content).decode("utf-8")
+
+
 class VLLMModel(BaseModel):
-    def __init__(self, **kwargs):
+    def __init__(self, data_dir: str = "", **kwargs):
+        self.data_dir = data_dir
```
gwarmstrong: Can we move this out of the VLLModel class? E.g., if we format the audio file path in around here: https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/inference/generate.py#L579

Also, would it be safe to make an assumption like `paths` are relative to the `input_file`? So we could reduce a parameter here?

nemo_skills/inference/model/vllm.py
```diff
@@ -99,6 +109,28 @@ def _build_completion_request_params(
             "extra_body": self._build_request_body(top_k, min_p, repetition_penalty, extra_body=extra_body),
         }
 
+    def content_text_to_list(self, message):
```
gwarmstrong: Is any of this logic able to be used outside of vllm? E.g., does it also work for openai APIs?

nemo_skills/inference/generate.py
```diff
@@ -525,6 +529,20 @@ def dump_outputs(self, outputs, data_points, fout):
...
+        if "messages" not in output:
+            return
+
+        for message in output["messages"]:
+            # Skip if content is not a list (e.g., string content in system messages)
+            if not isinstance(message.get("content"), list):
+                continue
+
+            # Filter out audio_url items from list-style content
+            message["content"] = [content for content in message["content"] if content.get("type") != "audio_url"]
```
gwarmstrong: can you make "audio_url" here a list that is specified in this config? that way it can be configurable and other fields can be included/excluded if desired

nemo_skills/inference/model/vllm.py
gwarmstrong: Even if not, I think it may have more utility if we make a hook for it in `generate.py` rather than here

tests/gpu-tests/test-local.yaml
```diff
@@ -17,6 +17,7 @@ executor: local
 containers:
   trtllm: nvcr.io/nvidia/tensorrt-llm/release:1.0.0
   vllm: vllm/vllm-openai:v0.10.1.1
+  vllm-audio: nvcr.io/nvidian/ac-aiapps/vllm-openai-audio:v1.0.0
```
gwarmstrong: Is this org/image gated? That will cause a failure if it is used in CI. I don't actually see this being used though--would it be used as the server container in the gpu tests?

melllinia: I am pretty sure this image is org‑gated. Yes, I am using it in the GPU tests. What’s the recommended workaround in this case?

gwarmstrong: This will be built into the base vllm image now--no need to pull nvcr container

nemo_skills/inference/model/vllm.py
Jorjeous: Please check again, if not outdated, than no - we should have it here as now we need to chunk audio

Jorjeous:
Need to fix import of audio_file_to_base64

nemo_skills/dataset/mmau-pro/prepare.py
```diff
@@ -75,22 +75,27 @@ def format_entry(entry, with_audio=False):
...
+        # Prepend /dataset/mmau-pro/ to make paths absolute for cluster
+        if len(audio_path) == 1:
+            user_message["audio"] = {"path": f"/dataset/mmau-pro/{audio_path[0]}"}
+        else:
+            user_message["audios"] = [{"path": f"/dataset/mmau-pro/{path}"} for path in audio_path]
+
+    # Don't use /no_think for open-ended questions to allow reasoning
+    system_content = "You are a helpful assistant."
+    if category != "open":
+        system_content += " /no_think"
```
Kipok, line 96: this is specific to qwen models, I don't think we should be doing it globally here?

nemo_skills/inference/generate.py
```diff
@@ -393,6 +404,32 @@ def setup_llm(self):
         else:
             llm = get_model(**self.cfg.server, tokenizer=self.tokenizer)
 
+        # Audio wrapper (preprocesses messages before they reach the model)
+        # Auto-enable for audio benchmarks on vLLM (eval_type=audio OR dataset_group=speechlm)
+        should_enable_audio = self.cfg.audio is not None or (
```
Kipok, line 409: I'd rather configure this through prepare.py of relevant benchmarks. Otherwise this logic is too coupled with datasets but this module is very general. So instead of dataset_group, I'd add a flag should_enable_audio and you can just set it as default parameter in the init of those benchmarks. You can still have a check for vllm to be used as a server and other logic for wrapping, just let's remove the dataset_group based behaviour

nemo_skills/inference/generate.py
```diff
@@ -519,12 +556,45 @@ def fill_prompt(self, data_point, data):
...
+        Override this method to add custom preprocessing logic.
+        Audio conversion is handled by the AudioProcessor wrapper.
+        """
+        return prompt
+
     def dump_outputs(self, outputs, data_points, fout):
         for output in outputs:
             fout.write(json.dumps(output) + "\n")
 
+    def drop_binary_data(self, output):
```
Kipok, line 582: maybe just drop_data or better drop_fields_from_messages or something like this as it's not limited to binary but just checks what to drop from the parameter?

nemo_skills/pipeline/utils/eval.py
```diff
@@ -131,6 +131,12 @@ def get_benchmark_args_from_module(
     if eval_args:
         generation_args = f"{eval_args} {generation_args}"
     generation_args += f" ++eval_config.split={split} "
+
+    # Pass dataset_group from benchmark config if defined
```
Kipok, line 135: so I'd remove this as well

nemo_skills/dataset/mmau-pro/prepare.py
Jorjeous: @melllinia ?

Jorjeous: Imho, this is not problem for other models

nemo_skills/inference/generate.py
Jorjeous: Got your point, when moving to separate class PR #1157 this would be auto-resolved

nemo_skills/inference/generate.py
Jorjeous: Ok, why not

nemo_skills/pipeline/utils/eval.py
Jorjeous: No harm from one additional field, but makes a but more convenient to separate datasets
Still want to remove?

nemo_skills/dataset/mmau-pro/prepare.py
Kipok: It certainly introduces a little mismatch, whether it's significant or not for accuracy is unclear, but I think it's much safer not to have it. What's the reason why we need to hardcode it here?

PR #1044, opened 2025-11-14
Allowing top_p to be set as null

Davood-M:
In this case, vllm will use the default top_p value from generation config of model


gwarmstrong (commented):
Looks good overall--please fix the failing check

Kipok:
closing as this will be fixed with gym migration

PR #1046, opened 2025-11-18
Port ICPC changes to IOI

SeanNaren:
Updates IOI with fixes from ICPC.

## Summary by CodeRabbit

* **New Features**
  * IOI evaluation now supports per-input test cases with dedicated input files
  * Enhanced IOI metrics with submission clustering and detailed per-subtask scoring

* **Documentation**
  * Updated IOI evaluation workflow with new command syntax and output formats

* **Bug Fixes**
  * Improved error handling and diagnostics for IOI evaluation

* **Chores**
  * CI workflow improvements for test infrastructure

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -52,7 +52,15 @@ jobs:
         cd ${{ github.run_id }}
         nvidia-smi
         set -o pipefail # this will make sure next line returns non-0 exit code if tests fail
-        ./tests/gpu-tests/run_qwen.sh
+        # Run heartbeat in background, capture its PID, and ensure cleanup
+        (while true; do sleep 60; echo "[HEARTBEAT] $(date '+%Y-%m-%d %H:%M:%S') - still running..."; done) &
+        HEARTBEAT_PID=$!
+        # Run tests and capture exit code
+        EXIT_CODE=0
+        ./tests/gpu-tests/run_qwen.sh || EXIT_CODE=$?
+        # Kill heartbeat and exit with test result
+        kill $HEARTBEAT_PID 2>/dev/null || true
+        exit $EXIT_CODE
     - name: Cleanup
       if: always()
       run: |
--- a/docs/evaluation/speech-audio.md
+++ b/docs/evaluation/speech-audio.md
@@ -2,8 +2,10 @@
 
 This section details how to evaluate speech and audio benchmarks, including understanding tasks that test models' ability to reason about audio content (speech, music, environmental sounds) and ASR tasks for transcription.
 
-!!! note
-    Currently supports only Megatron server type (`--server_type=megatron`).
+!!! warning "Running without audio files"
+    If you want to evaluation without audio files (not recommended) use
+    `--no-audio` flag. In this case you can also set `--skip_data_dir_check`
+    as data is very lightweight when audio files aren't being used.
 
 ## Supported benchmarks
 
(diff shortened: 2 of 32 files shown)


.github/workflows/tests.yml
```diff
@@ -39,7 +39,10 @@ jobs:
         NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
       run: |
-        docker run --rm --network=host nemo-skills-sandbox-image &
+        # Default shared runtime directory
+        sudo mkdir -p /nemo_run
+        sudo chmod 777 /nemo_run
+        docker run --rm --network=host -v /nemo_run:/nemo_run nemo-skills-sandbox-image &
```
SeanNaren, line 91: @gwarmstrong we need /nemo_run to be available just like it is when you run it in the container, is this ok to add globally to the tests?

docs/evaluation/code.md
```diff
@@ -207,24 +207,30 @@ ns eval \
...
+    --extra_eval_args="++eval_config.test_file=<PATH_TO_METADATA_TEST_DIR>/ioi24_metadata.json" \
     ++inference.temperature=0.6 \
     ++inference.top_p=0.95 \
     ++inference.tokens_to_generate=65536
 ```
 
 ##### Verifying Results
 
-After all jobs are complete, you can check the results in `<OUTPUT_DIR>/eval-results/ioi24/metrics.json`. You can also take a look at `<OUTPUT_DIR>/eval-results/ioi24/summarized-results/main_*`. They should look something like this:
+After all jobs are complete, you can check the results in `<OUTPUT_DIR>/eval-results/ioi24/ioi/metrics.json`. You can also take a look at `<OUTPUT_DIR>/eval-results/ioi24/ioi/summarized-results/main_*`. They should look something like this:
```
SeanNaren, line 224: @mehrzads is this naming convention ok? I realised that summarization requires knowing the dataset name and preserving the order of folders, so ioi has to be nested in order to map to the correct metrics class

SeanNaren: for people like yourself that handle this one level higher, you don't need to specify the eval_subfolder.

.github/workflows/tests.yml
gwarmstrong: What do you actually get by doing this? Or where does it cause an errorif you don't do it? It's actually not clear to me what this does other than make sure the directory exists in the container.

docs/evaluation/code.md
mehrzads: @SeanNaren yes this is good. We will handle it in higher level

.github/workflows/tests.yml
SeanNaren: you can see it here: https://github.com/NVIDIA-NeMo/Skills/actions/runs/19473949430/job/55728662396#step:6:5034

nemo_skills/evaluation/evaluator/ioi.py
```diff
@@ -88,76 +89,60 @@ def _precompile_grader(
         wait_for_sandbox(sandbox)
         sandbox._owner_tid = threading.get_ident()
 
-    pre_dir = f"/tmp/ioi_pre_{problem_name}_{os.getpid()}"
-    # Build shell script to create files and invoke compile.sh.
-    creation_cmds = [
-        f"mkdir -p {pre_dir}/graders",
-    ]
-    # Dump grader related files
+    pre_dir = f"/nemo_run/ioi_pre_{problem_name}_{os.getpid()}"
```
gwarmstrong, line 97: What's the reason for writing to the `/nemo_run` dir? Rather than creating a new mount like `/workspace/ioi_problems/`? I realize this `/nemo_run/ioi_pre_{problem_name}_{os.getpid()}` is not a directory that likely is made by `nemo-run` at any point--but I get nervous writing to `/nemo_run` as it is kind of supposed to be a static snapshot of what was run and this would be modifying that directory at runtime.

gwarmstrong: This is probably a minor concern.

.github/workflows/tests.yml
gwarmstrong: Addressed in comment below.

PR #1048, opened 2025-11-19
SWE-bench: install packages only once per job and share between instances

ludwig-n:
This PR makes SWE-bench evaluation less prone to network issues by downloading files & dependencies only once per job, instead of separately in each instance container.

Currently, SWE-agent/OpenHands and the SWE-bench evaluation harness are downloaded and installed inside of every single instance container. When thousands of instances are run simultaneously on a cluster, these downloads can slow down drastically or get stuck due to the strain on the network, causing errors and GPU idling. We've seen this particularly often recently with conda downloads.

With this PR, all files & dependencies required for the evaluation are only downloaded & installed once at the beginning of each job, and then mounted into the child containers so that they can re-use them. This cuts down the overall download size by a factor of 50 for our typical setup (50 instances per job). Also, conda is no longer used, which speeds up the OpenHands setup process.

The Slurm tests for SWE-bench pass with no regression in quality.

### DIFF
--- a/nemo_skills/inference/eval/swebench.py
+++ b/nemo_skills/inference/eval/swebench.py
@@ -183,6 +183,7 @@ class SweBenchGenerationTask(GenerationTask):
 
         self.should_run_evaluation = False
         self.evaluator = None
+        self._reasoning_warning_shown = False
 
         # Set up output folder,
         # making sure it is different for each random seed if we're running with --benchmarks=swe-bench:N


nemo_skills/inference/eval/swebench.py
```diff
@@ -182,6 +183,105 @@ def __init__(self, cfg: SweBenchGenerationConfig):
 
         self.should_run_evaluation = False
         self.evaluator = None
+        self._reasoning_warning_shown = False
```
gwarmstrong: does it not make sense to run `super().__init__(cfg)` here instead? This would cut down on some of the other boilerplate above too.

gwarmstrong (commented):
LGTM and tests pass on my end as well!

PR #1049, opened 2025-11-19
token count fix

Davood-M:


Kipok:
will need to be addressed in gym

PR #1051, opened 2025-11-21
Add proof-verification paper recipe

smahdavi4:


recipes/proof-gen-verification/README.md
```diff
@@ -1 +1,70 @@
...
+## Prerequisites
+
+To get started with NeMo-Skills, follow the instructions [here](https://nvidia-nemo.github.io/Skills/basics). Once you have set up your cluster configs, you can proceed with the steps below.
+
+## Dataset Preparation
+
+First, generate the dataset files. Run the following command for each dataset you need:
+
+```bash
+python3 nemo_skills/dataset/{dataset}/prepare.py
```
gwarmstrong: I think this might be more clear if you show using it with an argument? I also typically use it through the `ns prepare_data` pipeline, but that might be a nit pick

Issue #829, opened 2025-09-19
Add Evaluator Classes for async eval


gwarmstrong:
The evaluation functions should all eventually be converted to evaluation classes that support `eval_full` and optional `eval_single`.

_Originally posted by @Kipok in https://github.com/NVIDIA/NeMo-Skills/pull/825#discussion_r2364225888_

PR #1093, opened 2025-12-10
Audiometrics unification

Jorjeous:
This PR is part of bulk improvements proposed introduced in https://github.com/NVIDIA-NeMo/Skills/pull/1072

Current one for unification of all audio-related metrics 

Now all tasks related to audio can be evaluated with one evaluator, making it easier to add new datasets

## Summary by CodeRabbit

* **New Features**
  * Introduced audio evaluation module with support for multiple audio tasks including automatic speech recognition (ASR), translation, character error rate, and hallucination detection.
  * Added audio metrics computation capabilities for comprehensive evaluation of audio-based results.


nemo_skills/evaluation/evaluator/audio.py
```diff
@@ -0,0 +1,395 @@
...
+        "pc_rate": round(pc_rate, 3),
+        "punct_precision": round(punct_precision, 3),
+        "punct_recall": round(punct_recall, 3),
+        "punct_f1": round(punct_f1, 3),
+        "cap_accuracy": round(cap_accuracy, 3),
+        "is_correct": pc_rate > 0.5,
+    }
+
+
+def eval_audio(cfg):
```
gwarmstrong: can you write this as a `BaseEvaluator` implementation? This will allow evaluation to be performed interleaved with inference--this will help reduce GPU idle time at the end of the job, make better use of the CPU during generation, and allow evaluations to be performed concurrently so wall time spent on evaluation is probably lower.

It should be a pretty straightforward application of your `evaluate_sample` in the `eval_single` method.

gwarmstrong (changes requested):
Overall looks pretty good, just a note to please try to use the `BaseEvaluator`. It is an open issue to convert other evaluations (#829 ), but I think it will be helpful to have it here from the start, if possible.

nemo_skills/evaluation/evaluator/audio.py
Jorjeous: Ok, done

PR #1102, opened 2025-12-12
Evaluation support for SWE-rebench

wasiahmad:
Eval harness for swe-rebench with local exec enabled: https://github.com/wasiahmad/SWE-rebench

## Summary by CodeRabbit

* **New Features**
  * Added SWE-rebench dataset module with default evaluation configuration and customizable settings.
  * Introduced utility tool for converting Docker containers to compatible container images.
  * Added dataset preparation script for processing and exporting SWE-rebench data with date-based splits.

* **Improvements**
  * Enhanced container name resolution in evaluation to support multiple configuration approaches.

### DIFF
--- a/nemo_skills/dataset/swe-rebench/__init__.py
+++ b/nemo_skills/dataset/swe-rebench/__init__.py
@@ -18,3 +18,4 @@ DATASET_GROUP = "code"
 METRICS_TYPE = "swe-bench"
 # evaluation is fused with generation for efficiency
 GENERATION_MODULE = "nemo_skills.inference.eval.swebench"
+GENERATION_ARGS = "++eval_harness_repo=https://github.com/wasiahmad/SWE-rebench.git "


nemo_skills/inference/eval/swebench.py
```diff
@@ -342,9 +342,14 @@ async def _execute_local_command(self, command, timeout=None):
 
     async def _execute_container_command(self, data_point, command, expected_file_pattern, mode, timeout=100000):
         """Execute a command in an Apptainer container with retry logic."""
-        container_name = data_point["container_formatter"].format(
-            instance_id=data_point["instance_id"].replace("__", "_1776_")
-        )
+        if "container_name" in data_point:
```
ludwig-n: Still think we should maybe reuse `container_formatter` here instead of introducing a new field with fallback logic. Then we wouldn't have to duplicate the `dump_images.py` script for Rebench, the user can just reuse the one from the `swe-bench` folder. And this file can stay exactly the same.

Also, if a swe-bench-style dataset (non-Rebench) has an existing column called `container_name`, this code will read from that column, leading to unexpected behavior. If we use `container_formatter` for all datasets, this can't happen, because the prepare script will overwrite it regardless.

The only downside I can think of is the name being misleading, since it wouldn't actually be a formatter in this case.

nemo_skills/dataset/swe-rebench/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+EVAL_SPLIT = "default"
+DATASET_GROUP = "code"
+METRICS_TYPE = "swe-bench"
+# evaluation is fused with generation for efficiency
+GENERATION_MODULE = "nemo_skills.inference.eval.swebench"
```
ludwig-n, line 20: Ideally we should make it so that `eval_harness_repo` and `eval_harness_commit` are set to your fork of the Rebench harness by default, right? Based on [this code](https://github.com/NVIDIA-NeMo/Skills/blob/7205c431fc5462ad76c0a00acc1c5a210d531385/nemo_skills/pipeline/utils/eval.py#L130), seems like it could be done by setting EVAL_ARGS here. Though there is a note saying it's deprecated, not sure what the recommended approach is then.

wasiahmad: Good point.

nemo_skills/inference/eval/swebench.py
wasiahmad: But @ludwig-n the instance_id does not match in container name. But I remember you mentioned that it won't affect both swe-bench inference and dump_images.py. Is my understanding right?

ludwig-n: Yeah, my idea is to simply store the Rebench container names in `container_formatter` instead of `container_name`, without using `{instance_id}`. Then the logic in swebench.py and dump_images.py can stay the same as it is now, since the `.format(` operation will be a no-op.

wasiahmad:
Reproduced numbers:

#### Model name: Qwen3-Coder-30B-A3B-Instruct
- Split: 2025-08 to 2025-11
- Number of samples: 200
- Framework: SWE-Agent
- Pass@1: [run1] 26.5%, [run2] 24.0%, [run3] 25.6%
- From leaderboard: Pass@1: 23.2%

PR #1106, opened 2025-12-12
Add RULERv2

hsiehjackson:
## New Features
* Added RULER2 benchmark with multiple dataset types (NIAH, MMLU, QA) at configurable difficulty levels
* New RULER2 evaluator with scoring capabilities for benchmark assessment
* Introduced tokenizer abstraction supporting multiple backend providers
* Added automated dataset preparation workflows for benchmark setup

### DIFF
--- a/docs/evaluation/other-benchmarks.md
+++ b/docs/evaluation/other-benchmarks.md
@@ -11,3 +11,77 @@ More details are coming soon!
 
 - Benchmark is defined in [`nemo_skills/dataset/arena-hard/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/arena-hard/__init__.py)
 - Original benchmark source is [here](https://github.com/lmarena/arena-hard-auto).
+
+### AA-Omniscience
+
+This is a benchmark developed by AA to measure hallucinations in LLMs and penalize confidently-false answers.
+
+- Benchmark is defined in [`nemo_skills/dataset/omniscience/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/omniscience/__init__.py)
+- Original benchmark and leaderboard are defined [here](https://artificialanalysis.ai/evaluations/omniscience), and data is [here](https://huggingface.co/datasets/ArtificialAnalysis/AA-Omniscience-Public)
+
+#### Eval Results:
+|        Model        | Accuracy | Omni-Index | Hallucination Rate |
+| ------------------- | -------- | ---------- | ------------------ |
+| Qwen3-8B (Reported) | 12.73%   | -66        | 90.36%             |
+| Qwen3-8B (Measured) | 15.17%   | -64.83     | 94.30%             |
+
+#### Notes:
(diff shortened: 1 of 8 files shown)


nemo_skills/dataset/ruler2/tokenizer.py
```diff
@@ -0,0 +1,127 @@
...
+from tenacity import (
+    retry,
+    stop_after_attempt,
+    wait_fixed,
+    wait_random,
+) 
+
+
+def select_tokenizer(tokenizer_type, tokenizer_path):
+    if tokenizer_type == 'nemo':
```
Kipok: do we really need this? I think all recent nemotrons use normal hf tokenizers

nemo_skills/evaluation/evaluator/ruler.py
```diff
@@ -16,6 +16,7 @@
 import logging
 import os
 import re
+import editdistance
```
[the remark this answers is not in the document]
Kipok: that seems reasonable, or is editdistance already installed in some other package? Still could explicitly add it @hsiehjackson

nemo_skills/dataset/ruler2/prepare.py
```diff
@@ -0,0 +1,577 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import argparse
+import concurrent.futures
+import subprocess
+from pathlib import Path
+
+DEFAULT_SETTINGS = """
```
Kipok, line 21: If this is fixed (e.g. doesn't depend on user-provided arguments), then it's best to just create subfolders and put this directly as init.py in there. See https://github.com/NVIDIA-NeMo/Skills/tree/main/nemo_skills/dataset/mmau-pro as an example

nemo_skills/dataset/ruler2/ruler2_score.py
```diff
@@ -0,0 +1,47 @@
...
+        "qa_hard",
+    ]
+    setup = list(metrics.keys())[0].rsplit(".", 1)[0]
+    metrics[setup] = {}
+
+    for aggregation in metrics[f"{setup}.mk_niah_basic"]:
+        metrics[setup][aggregation] = {
+            "accuracy": sum(
+                metrics[f"{setup}.{task}"][aggregation].get(
+                    "accuracy", (metrics[f"{setup}.{task}"][aggregation].get("symbolic_correct", 0))
```
Kipok: do we need .get? If symbolic_correct is always expected to be present, we should just use ["symbolic_correct"] instead to not hide unexpected issues

nemo_skills/dataset/ruler2/prepare.py
hsiehjackson: We don't create init.py directly since this is similar to RULERv1 that all data are depending on the length which we will create different folder name based the setup argument.

nemo_skills/dataset/ruler2/ruler2_score.py
hsiehjackson: Fixed

Kipok (approved):

Issue #1108, opened 2025-12-13
Fix `AssertionError: no app_id collisions expected` when running multiple tasks with `executor="none"`


bzantium:
**Description**
When running a pipeline using `cluster_config["executor"] = "none"` (local execution without docker), if a job step contains multiple tasks (commands), the execution fails with the following error:

```
AssertionError: no app_id collisions expected since uuid4 suffix is used
```

This occurs in the `torchx` local scheduler integration within `nemo_run`, specifically when submitting multiple bundled scripts as a single job entry in the experiment DAG.

**Proposed Change**
To resolve this, we need to separate tasks into individual jobs when the executor is set to `"none"`.

1.  **Update `nemo_skills/pipeline/utils/exp.py`**:

      * Check if `cluster_config["executor"] == "none"`.
      * If true, iterate through the commands and add them to the experiment as separate jobs (calling `exp.add` individually for each) rather than bundling them into a single list of `run.Script` objects.

2.  **Update `nemo_skills/pipeline/eval.py`**:

      * Modify the `eval` function to handle the case where `pipeline_utils.add_task` returns a list of tasks (jobs) instead of a single job object.
      * Ensure `prev_tasks` and `all_tasks` are updated correctly (using `extend` for lists).

**Error Traceback**

```
  File "/nemo_run/code/nemo_skills/pipeline/eval.py", line 733, in eval
    pipeline_utils.run_exp(exp, cluster_config, dry_run=dry_run)
  File "/nemo_run/code/nemo_skills/pipeline/utils/exp.py", line 728, in run_exp
    exp.run(detach=False, tail_logs=True, sequential=False)
...
  File "/home/user/.local/lib/python3.10/site-packages/nemo_run/run/torchx_backend/schedulers/local.py", line 106, in schedule
    app_id = super().schedule(dryrun_info=dryrun_info)
  File "/home/user/.local/lib/python3.10/site-packages/torchx/schedulers/local_scheduler.py", line 791, in schedule
    app_id not in self._apps
AssertionError: no app_id collisions expected since uuid4 suffix is used
```

PR #1109, opened 2025-12-13
fix(pipeline): separate tasks for local executor to prevent app_id collisions

bzantium:
This PR fixes a `AssertionError: no app_id collisions expected` crash that occurs when running pipelines locally (`executor="none"`) with multiple tasks in a single job step.

**Problem**
When `cluster_config["executor"]` is set to `"none"`, the `nemo_run` local scheduler (via `torchx`) does not support bundling multiple commands into a single job entry the same way Slurm/remote executors do. Submitting them as a bundled list causes internal App ID collisions during the `schedule` phase.

**Changes**

  * **`nemo_skills/pipeline/utils/exp.py`**: Added logic to check if the executor is `"none"`. If so, it now iterates through the commands and adds them as individual jobs to the experiment (returning a list of jobs) instead of bundling them.
  * **`nemo_skills/pipeline/eval.py`**: Updated the `eval` function to handle `add_task` returning a list of jobs. It now checks the type of `new_task` and uses `.extend()` for lists or `.append()` for single objects when updating the task tracking lists.

**Related Error Traceback**

```
  File "/home/user/.local/lib/python3.10/site-packages/nemo_run/run/torchx_backend/schedulers/local.py", line 106, in schedule
    app_id = super().schedule(dryrun_info=dryrun_info)
  File "/home/user/.local/lib/python3.10/site-packages/torchx/schedulers/local_scheduler.py", line 791, in schedule
    app_id not in self._apps
AssertionError: no app_id collisions expected since uuid4 suffix is used
```

Fixes: #1108 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task dependency tracking to properly handle batch task submissions as multiple dependent tasks.
  * Improved multi-command execution handling for independent cluster mode by submitting each command separately with correct dependency management.

### DIFF
--- a/nemo_skills/pipeline/eval.py
+++ b/nemo_skills/pipeline/eval.py
@@ -547,8 +547,12 @@ def eval(
                     installation_command=installation_command,
                     skip_hf_home_check=skip_hf_home_check,
                 )
-                prev_tasks = [new_task]
-                all_tasks.append(new_task)
+                if isinstance(new_task, list):
+                    prev_tasks = new_task
+                    all_tasks.extend(new_task)
+                else:
+                    prev_tasks = [new_task]
+                    all_tasks.append(new_task)
                 # only last dependent job will be here, which is what we want
                 job_id_to_tasks[idx] = prev_tasks
         # scheduling judge jobs if needed
--- a/nemo_skills/pipeline/utils/exp.py
+++ b/nemo_skills/pipeline/utils/exp.py
@@ -691,15 +691,26 @@ def add_task(
     else:
         if heterogeneous:
             executors[0].het_group_indices = het_group_indices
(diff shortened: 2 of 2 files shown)


bzantium:
@gwarmstrong  Apologies, but I realized that the approach in this PR is a workaround rather than a complete fix.
After further investigation, I identified that the root cause lies within the upstream library (`Nemo/Run`). Specifically, when handling a `JobGroup` with multiple executables, the logic fails to store `dryrun_info` for each executable individually. Instead, it overwrites them, preserving only the last one, which leads to the `app_id` collision.
I will close this PR and submit a patch to the `Nemo/Run` repository to address the core issue properly. Thanks for your time!

PR #1117, opened 2025-12-16
replace raise error with LOG.warning in AA LCR dataset prepare

anowaczynski-nvidia:
## Summary by CodeRabbit

* **Improvements**
  * Enhanced dataset preparation robustness to gracefully handle token count mismatches, allowing processing to continue seamlessly instead of halting.

### DIFF
--- a/tests/gpu-tests/test_eval.py
+++ b/tests/gpu-tests/test_eval.py
@@ -44,7 +44,6 @@ EXCLUDED_DATASETS = {
     "mbpp",
     "mmau-pro",
     "asr-leaderboard",
-    "aalcr",  # Has tokenization mismatch issues
     "audiobench",
     "librispeech-pc",
 }


nemo_skills/dataset/aalcr/prepare.py
```diff
@@ -187,7 +187,7 @@ def write_data_to_file(output_file, data, txt_file_folder, max_context_window, t
                     continue
 
             if n_tokens != entry["input_tokens"]:  # check if the n_tokens exactly match the input_tokens in the entry
-                raise ValueError(f"n_tokens: {n_tokens} != input_tokens: {entry['input_tokens']}")
+                LOG.warning(f"n_tokens: {n_tokens} != input_tokens: {entry['input_tokens']}")
```
gwarmstrong, line 190: can you also remove it from the excluded datasets? Then we can run gpu tests to confirm our integration is fixed: https://github.com/NVIDIA-NeMo/Skills/blob/main/tests/gpu-tests/test_eval.py#L47

gwarmstrong (changes requested):

PR #1124, opened 2025-12-17
Add ComputeEval Dataset Support

blahblahasdf:
### Overview
This PR integrates ComputeEval - a benchmark for evaluating LLMs on CUDA code generation - into NeMo-Skills.

### ComputeEval features:

1. 406 (and growing) CUDA programming challenges
2. Functional correctness evaluation via compilation and test execution
3. Multiple release versions (2025-1, 2025-2, 2025-3)

### Changes
1. Dataset Preparation - download ComputeEval problems from HuggingFace (gated dataset)
2. Generation Task - implement `GenerationTask` to generate CUDA solutions using ComputeEval's reference generation logic
3. Evaluator - implement `BaseEvaluator` using ComputeEval's evaluation logic
4. Metrics - compute pass@k metrics

### Dependencies
Adds compute-eval package (`compute-eval @ git+https://github.com/NVIDIA/compute-eval.git@2d14770`)
Note: compute-eval is a public repository but has not been published to PyPI yet

## Release Notes

* **New Features**
  * Added support for compute-eval dataset from HuggingFace with optional release specifications.
  * Introduced compute-eval evaluator for CUDA code assessment with automatic CTK version detection.
  * Added compute-eval metrics for computing evaluation scores including pass-at-k performance.
  * Integrated compute-eval generation task into the inference pipeline.
  * Registered compute-eval as a new evaluation type across the framework.

### DIFF
--- a/docs/evaluation/code.md
+++ b/docs/evaluation/code.md
@@ -178,6 +178,65 @@ all you need to do is replace `openhands` with `swe_agent` in the command above.
 !!! note
     For evaluation, we use a [custom fork](https://github.com/Kipok/SWE-bench) of the SWE-bench repository that supports running evaluation inside of an existing container. It may not always have the latest updates from the upstream repo.
 
+### compute-eval
+
+- Benchmark is defined in [`nemo_skills/dataset/compute-eval/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/compute-eval/__init__.py)
+- Original benchmark source is [here](https://github.com/NVIDIA/compute-eval).
+
+ComputeEval is a benchmark for evaluating Large Language Models on CUDA code generation tasks. It features handcrafted CUDA programming challenges that test an LLM's capability at writing reliable CUDA code. The benchmark includes functional correctness evaluation through compilation and execution against held-out test suites.
+
+**Prerequisites:** NVIDIA GPU with CUDA Toolkit 12 or greater must be installed, and `nvcc` must be available in your PATH.
+
+#### Data Preparation
+
+First, prepare the dataset by running the `ns prepare_data` command. You can optionally specify a release version:
+
+```bash
+ns prepare_data compute-eval --release 2025-1
+```
+
+If no release is specified, the default release will be downloaded. This will generate an `eval.jsonl` file in the `nemo_skills/dataset/compute-eval/` directory.
+
+**Note:** You need to set the `HF_TOKEN` environment variable because the dataset requires authentication.
+
+#### Running the Evaluation
+
+Once the data is prepared, you can run the evaluation. Replace `<...>` placeholders with your cluster and directory paths.
+
+This command runs an evaluation of [OpenReasoning-Nemotron-32B](https://huggingface.co/nvidia/OpenReasoning-Nemotron-32B) on a Slurm cluster:
+
+```bash
+ns eval \
+    --cluster=<CLUSTER_NAME> \
+    --model=nvidia/OpenReasoning-Nemotron-32B \
+    --server_type=vllm \
+    --server_args="--async-scheduling" \
+    --server_nodes=1 \
+    --server_gpus=8 \
+    --benchmarks=compute-eval \
+    --data_dir=<DATA_DIR> \
+    --output_dir=<OUTPUT_DIR> \
+    ++inference.temperature=0.6 \
+    ++inference.top_p=0.95 \
+    ++inference.tokens_to_generate=16384
+```
+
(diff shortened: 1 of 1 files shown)


nemo_skills/dataset/compute-eval/prepare.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def _get_hf_token():
+    token = os.getenv("HF_TOKEN")
+    if token:
+        return token
+
+    # noinspection PyBroadException
+    try:
+        api = HfApi()
```
gwarmstrong: what is the purpose of this?

nemo_skills/inference/eval/compute_eval.py
```diff
@@ -0,0 +1,117 @@
...
+        return
+
+    async def process_single_datapoint(self, data_point, data):
+        try:
+            problem = _TYPE_ADAPTER.validate_python(data_point["problem"])
+            solution = await asyncio.to_thread(
+                generate_model_completions,
+                system_prompt=self._system_prompt,
+                problem=problem,
+                model=self._model,
```
gwarmstrong: Looks like this is applying this function:
https://github.com/NVIDIA/compute-eval/blob/main/compute_eval/generate_completions.py#L31-L84

Which appears to use the ComputeEval clients? 
What I would recommend we do instead, is write a version of that function that uses NeMo-Skills' `process_single_datapoint`--so this probably looks something like 

```
    result = super().process_sinlge_datapoint(data_point, data)
    ...

    completion = ...
    return FileSolution(
        task_id=problem.task_id,
        files=_parse_solution(completion),
        **debug_info,
    )
```

this will make sure the prompt logic is consistent with typical NeMo-Skills and that the eval prompt logic is the same as normal generations.

gwarmstrong: Then the above methods that are overwritten would just use the default methods rather than override to return nothing.

blahblahasdf: Thanks for the quick feedback! Your suggestion is to call `process_sinlge_datapoint` and then use the ComputeEval post processing to convert the model's response to a `FileSolution`?  The "ComputeEval clients" are just a pass through to the OpenAI client.  I think as is if you were to configure to a local model (or a model hosted somewhere else) it would just work as is but I agree it doesn't fit super cleanly with the NeMo-Skills abstractions.

Honestly, the generation side of the ComputeEval repo is generally intended to be a reference implementation for how other groups/companies would set up a generator scaffold to solve the ComputeEval problems.  The reference implementation's prompt and post processing logic are pretty coupled.

nemo_skills/dataset/compute-eval/prepare.py
blahblahasdf: This allows a user to access the HuggingFace gated (you have to be allow-listed to have access to the ComputeEval data today) data either via an environment variable (`HF_TOKEN`) or by using the HuggingFace CLI to login.

blahblahasdf: Actually, after looking closer we don't need to access the API at all for this to work.  I'll just clean this up.

gwarmstrong: yeah, we should just need a `load_dataset` call and it will handle the token if it's already in your env variables, like this one is gated: https://github.com/blahblahasdf/Skills/blob/cb7829094799f9b5967e260d30e39c2535f67bab/nemo_skills/dataset/flores200/prepare.py#L49

nemo_skills/inference/eval/compute_eval.py
gwarmstrong: The logic for calling the openai client is here, right? https://github.com/NVIDIA/compute-eval/blob/main/compute_eval/models/model_interface.py#L100-L154
Seems pretty straightforward and is pretty much a subset of the functionality our clients have. And if do it the way it's currently being done, you lose a lot of things we support in NeMo-Skills (like tool-calling, pretty much all sampling parameters not directly specified in the compute-eval function, as well as the supported server types would be quite limited).

blahblahasdf: Thanks for the pointers and happy New Year!  I see where you're coming from on fitting this in with the NemoSkills generation more cleanly.  What is the expectation on how to document this`GenerationTask`s content format?  I was able to get this working locally but there is definitely some coupling between the expected data format and `system` and `user` prompts required in a `++prompt_config`.  I would guess that has to be normal and just documented?

gwarmstrong: Yeah that's pretty typical. We don't really have it documented on a case-by-case basis--it's pretty easy to infer from the prompt config itself, and we have a generic section on understanding the interaction between data and prompt_config here: https://nvidia-nemo.github.io/Skills/basics/prompt-format/#prompt-api

blahblahasdf:
Closing this PR in favor of one directly from a branch on NeMo Skills.

PR #1132, opened 2025-12-18
BIRD Benchmark (Text-to-SQL)

redoctopus:
Added BIRD's SQL execution accuracy metric to code generation benchmarks.
Also added some documentation on how to download/process data and run the benchmark.

Original dataset website: https://github.com/AlibabaResearch/DAMO-ConvAI/tree/main/bird
SQL execution from BIRD: https://github.com/AlibabaResearch/DAMO-ConvAI/blob/main/bird/llm/src/evaluation.py

## Summary by CodeRabbit

* **New Features**
  * Added BIRD benchmark evaluation framework for text-to-SQL tasks with difficulty-categorized accuracy metrics and database validation testing.

* **Documentation**
  * Updated evaluation documentation with BIRD benchmark instructions and data preparation guides; added BIRD to code evaluation references.

### DIFF
--- a/nemo_skills/evaluation/evaluator/bird.py
+++ b/nemo_skills/evaluation/evaluator/bird.py
@@ -118,7 +118,7 @@ class BirdEvaluator(BaseEvaluator):
             return "SELECT 1"  # No-op filler
 
         # Remove comments first
-        ans = re.sub(r"--.*?|/\*.*?\*/", "", code_matches[-1], flags=re.DOTALL)  # Use last match
+        ans = re.sub(r"--.*?$|/\*.*?\*/", "", code_matches[-1], flags=re.DOTALL)  # Use last match
         # Collapse whitespace
         ans = re.sub(r"\s+", " ", ans)
         # Remove miscellaneous headers that snuck in


docs/index.md
```diff
@@ -16,7 +16,7 @@ Here are some of the features we support:
     - Evaluate your models on many popular benchmarks.
         - [**Math (natural language**)](./evaluation/natural-math.md): e.g. [aime24](./evaluation/natural-math.md#aime24), [aime25](./evaluation/natural-math.md#aime25), [hmmt_feb25](./evaluation/natural-math.md#hmmt_feb25)
         - [**Math (formal language)**](./evaluation/formal-math.md): e.g. [minif2f](./evaluation/formal-math.md#minif2f), [proofnet](./evaluation/formal-math.md#proofnet), [putnam-bench](./evaluation/formal-math.md#putnam-bench)
-        - [**Code**](./evaluation/code.md): e.g. [swe-bench](./evaluation/code.md#swe-bench), [livecodebench](./evaluation/code.md#livecodebench)
+        - [**Code**](./evaluation/code.md): e.g. [swe-bench](./evaluation/code.md#swe-bench), [livecodebench](./evaluation/code.md#livecodebench), [bird](./evaluation/code.md#BIRD)
```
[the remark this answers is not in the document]
redoctopus: ?? Looks like the same indentation level as the other benchmark bullet points, at least to me...
Someone please let me know if I'm missing something here, otherwise I'll mark as resolved.

docs/evaluation/code.md
```diff
@@ -316,6 +316,48 @@ Due to variance between runs, you can automatically repeat the evaluation and av
...
+
+```bash
+ns eval \
+     --cluster=<CLUSTER_NAME> \
+     --server_type='sglang' \
+     --server_gpus=8 \
+     --model=Qwen/Qwen3-8B \
+     --benchmarks=birdbench \
+     --output_dir=<OUTPUT_DIR> \
+     --installation_command="pip install func_timeout" \
```
gwarmstrong: this shouldn't be needed once we have the updated requirements merged, so we can probably remove it here

docs/evaluation/code.md
```diff
@@ -316,6 +316,48 @@ Due to variance between runs, you can automatically repeat the evaluation and av
...
+     ++inference.tokens_to_generate=10000 \
+     ++inference.temperature=0.6 \
+     ++inference.top_p=0.95 \
+     ++inference.top_k=20 \
+     ++max_concurrent_requests=1024 \
+     ++eval_config.dev_json_filepath=<DATA_DIR>/dev.json \
+     ++eval_config.db_path=<DATA_DIR>/dev_databases
+```
+!!! note
+    BIRD evaluation requires the `func_timeout` library, so you will need to add the installation argument `--installation_command` seen in this example if it is not included in your container.
```
gwarmstrong: again probably unnecessary once we merge this

nemo_skills/evaluation/evaluator/bird.py
```diff
@@ -0,0 +1,128 @@
...
+import asyncio
+import json
+import logging
+import os
+import re
+from pathlib import Path
+
+from func_timeout import FunctionTimedOut, func_timeout
+
+from nemo_skills.dataset.birdbench.evaluation import execute_sql
```
gwarmstrong: the other file is just one function--maybe we can move it into this file?

nemo_skills/evaluation/evaluator/bird.py
```diff
@@ -0,0 +1,128 @@
...
+        # Remove comments first
+        ans = re.sub(r"--.*", "", code_matches[-1])  # Use last match
+        # Collapse whitespace
+        ans = re.sub(r"\s+", " ", ans)
+        # Remove miscellaneous headers that snuck in
+        ans = re.sub(r"^\*\*.*\*\*", "", ans).strip()
+
+        return ans
+
+    async def eval_full(self):  # type: ignore[override]
```
gwarmstrong: are we able to remove this method and simply rely on the `super()`'s method, which should basically do exactly this--apply `eval_single` over each line

nemo_skills/evaluation/evaluator/bird.py
```diff
@@ -0,0 +1,128 @@
...
+        db_id = data_point["db_id"]
+
+        # Retrieve pred and gt
+        predicted_sql = self._extract_answer(data_point["generation"])
+        ground_truth = data_point["gt_sql"]
+        db_place = str(Path(self.eval_config.db_path, db_id, db_id + ".sqlite"))
+
+        try:
+            # Wait for result with timeout as set
+            res = func_timeout(self.eval_config.timeout, execute_sql, args=(predicted_sql, ground_truth, db_place))
```
gwarmstrong, line 140: Is this a hard requirement to use this `func_timeout` library? Or can we write something simple to do it ourselves? I would prefer if we didn't have to introduce a new dependency for what ideally should be a really simple utility. Let me know if you see something I don't with this.

nemo_skills/evaluation/evaluator/bird.py
```diff
@@ -0,0 +1,128 @@
...
+class BirdEvaluatorConfig(BaseEvaluatorConfig):
+    timeout: int = 30
+
+    # Answer format can be "BOXED", "CODEBLOCK", or "USE_REGEX", the last of
+    # which uses the given regex in the extraction_regex arg.
+    answer_format: str = "CODEBLOCK"
+    extraction_regex: str | None = None
+    regex_dotall: bool = False
+
+    db_path: str = ""
```
gwarmstrong: I think we can always treat this relative to the dataset path, right? So we shouldn't have to specify it in the config--see our logic for getting the dataset path here: https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/prepare.py#L35-L52

gwarmstrong: Also, I think making this empty rather than having no default will lead to unexpected errors if someone doesn't know to update db_path

nemo_skills/dataset/birdbench/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+GENERATION_ARGS = (
+    "++prompt_config=generic/text_to_sql "
+    "++eval_type=bird "
+    "++inference.tokens_to_generate=10000 "
+    "++inference.temperature=0.6 "
+    "++inference.top_p=0.95 "
+    "++inference.top_k=20 "
+    "++max_concurrent_requests=1024 "
+    # Note: Override this path via ++eval_config.db_path
+    "++eval_config.db_path=/sql_data/BIRD/dev_databases"
```
gwarmstrong: It would be great if we could infer this from the dataset directory rather than hard code a bad value--I think this will cause unexpected issues where the failure that happens if someone doesn't specify this will be of the nature of `FileNotFound`, rather than the user knowing the option they should be updating

docs/evaluation/code.md
```diff
@@ -316,6 +316,48 @@ Due to variance between runs, you can automatically repeat the evaluation and av
...
+
+The arguments for evaluation also include paths for `++eval_config.dev_json_filepath` and `++eval_config.db_path`. These are used for SQL execution. You should point these to the unmodified `dev.json` file and `dev_databases` directory found in the original downloaded and unzipped data (i.e. `.../dev_20240627/dev.json`, etc.). If using a Slurm cluster, you may need to upload and mount the data directory first.
```
gwarmstrong: If these paths are relative to the `dataset`, I would prefer we infer it from the dataset, rather than have to always specify these paths

nemo_skills/dataset/birdbench/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+DATASET_GROUP = "code"
+METRICS_TYPE = "bird"
+EVAL_SPLIT = "dev"
```
gwarmstrong, line 18: is there a strong reason to call this split `dev` rather than something more canonical like `test` ?

gwarmstrong (changes requested):
Thanks @redoctopus ! a few comments and suggestions here

docs/evaluation/code.md
redoctopus: Oh good catch, thanks. Removing these references to `installation_command`.

docs/evaluation/code.md
redoctopus: The `dev_databases` directory isn't inherently tied to the dataset, so I don't think we can infer it directly.

Example use case: BIRD data processing is done locally, but then generation and evaluation are done on either locally or using a cluster--to have the flexibility to do either we need to be able to dynamically specify the location of the database files.

redoctopus: (Side note: I removed `dev_json_filepath` in my last refactor so at least that's cleaner now :)

gwarmstrong: Is it the same database content locally and on the cluster? How big are the databases?

nemo_skills/dataset/birdbench/__init__.py
redoctopus: Yep--they have a separate `test` set that isn't publicly available, but is labeled as such on their leaderboard. The set we're using to evaluate is what they officially call `dev`.

nemo_skills/dataset/birdbench/__init__.py
redoctopus: Given the above (can't necessarily couple it with the dataset manifest), is there a better way to handle this? Maybe comment this out and then add a check for the value in evaluation that gives a more informative error?

nemo_skills/evaluation/evaluator/bird.py
redoctopus: Referring to the example I mentioned in the other comment where the database directory can be located in different places depending on where the code is executed, in the cluster execution case I believe it _should_ actually be specifically decoupled as it should be uploaded separately first--it's a somewhat large directory that shouldn't be copied to the cluster for each run.

> Also, I think making this empty rather than having no default will lead to unexpected errors if someone doesn't know to update db_path

Related to other comment--I can add a check for empty `db_path` to be more informative if that's helpful.

nemo_skills/evaluation/evaluator/bird.py
redoctopus: Ooh yes, I think so! Testing this out now to make sure I didn't break anything by updating this (and `eval_single`'s return value).

nemo_skills/evaluation/evaluator/bird.py
redoctopus: Ahhhh this! I spent a few days trying to get this to work with any of the other multiprocessing/multithreading/signal handler override timeout methods I could find that wouldn't need an extra dependency, to no avail.

I found that the main issue for multi* in python is that apparently all the timeouts just alert the parent that the child has timed out, but lets the child continue until finished in the background. There are no functions to abort the child at this point; the only stop functions I could find were to cancel unrun jobs.

The built-in timeout for sqlite3 doesn't interrupt reaaaaaally long IO operations like the one this timeout is supposed to catch (i.e. super inefficient queries that take forever to run). Overwriting the signal handler for SIGALRM didn't work either for I think the same reason--the signal isn't handled until the read finishes, which defeats the purpose.

I did find some additional libraries that modified the multi* APIs to automatically kill timed-out children, but then that's the same issue as what we have now--a new dependency. In the end I just stuck with this timeout library because that's what the official BIRD repo's eval script was using.

However, if I am missing anything I would be happy to try it out! It would be great to remove this extra dependency just for one function in one benchmark.

nemo_skills/evaluation/evaluator/bird.py
gwarmstrong: if you leave it as:
```
    db_path: str
```
won't hydra enforce it is specified? Rather than have to implement the check

nemo_skills/evaluation/evaluator/bird.py
gwarmstrong: sounds good, if it works well and is consistent with how the upstream BIRD handles it, I think it is fine

docs/evaluation/code.md
redoctopus: Yep, should be the same. 1.8GB unzipped, 331MB zipped.

nemo_skills/evaluation/evaluator/bird.py
redoctopus: (It works!)

nemo_skills/evaluation/evaluator/bird.py
redoctopus: Oh, I didn't know that! I'll update that and remove the manual check.

docs/evaluation/code.md
gwarmstrong: Can we use the `data_dir` argument instead, then?

https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/pipeline/prepare_data.py#L34-L50

Then we just have to pass around that same argument between prepare_data and eval, but don't need to go separately prepare it on the cluster and then get the path back. Makes it easier to keep things consistent.

redoctopus: I think I might be missing something here but I'm still not quite seeing how that would work--if we ran prepare_data locally and then wanted to run evaluation on the cluster, at prepare_data runtime `data_dir` would only have the local path to the directory and not the cluster's path, right? Additionally, if we don't want to have NeMo-Skills upload the dataset for every run, we can't guarantee it'll be in any given relative path location e.g. `nemo_skills/datasets/birdbench/...` since the user could upload it anywhere else.

(And a quick clarification: currently we don't need to prepare the data separately, the pipeline is more like (1) run prepare_data locally, (2) upload that directory to desired cluster, (3) run evaluation with that directory mounted & pass in that path.)

gwarmstrong: Right, so the way we typically do exactly that is with `data_dir`. See the `livecodebench` for `pypy3` docs: https://nvidia-nemo.github.io/Skills/evaluation/code/#for-pypy3-evaluation


In this case then, (1) becomes optional. (2) is done with `ns prepare_data  birdbench --cluster=<cluster name> --data_dir=<data_dir>`, then you can do any amount of eval with (3) using `ns eval --cluster=<cluster name> --data_dir=<data_dir>` without it uploading--it should find the data in `data_dir` rather than uploading. 

This helps us keep things consistent, and gives you the interface to upload with, rather than us having one-off  copy workflows and arguments for each dataset.

redoctopus: Ah, so in this case we're forgoing local data prep in favor of doing it all on the cluster. Gotcha, that's fine with me! Even if we want to set up on multiple clusters the data prep is pretty quick to run.

docs/evaluation/code.md
```diff
@@ -316,6 +316,43 @@ Due to variance between runs, you can automatically repeat the evaluation and av
...
+     --benchmarks=birdbench \
+     --data_dir=<DATA_DIR> \
+     --output_dir=<OUTPUT_DIR> \
+     ++inference.tokens_to_generate=10000 \
+     ++inference.temperature=0.6 \
+     ++inference.top_p=0.95 \
+     ++inference.top_k=20 \
+     ++max_concurrent_requests=1024 \
+```
+You should specify: `<CLUSTER_NAME>`, which should match your cluster config name; `<DATA_DIR>`, which should be the path to your base `nemo_skills/datasets` directory on the cluster; and `<OUTPUT_DIR>`.
```
gwarmstrong:
```suggestion
You should specify: `<CLUSTER_NAME>`, which should match your cluster config name and `<OUTPUT_DIR>`.
```

gwarmstrong: can you also maybe link to here? https://nvidia-nemo.github.io/Skills/evaluation/#using-data-on-cluster

docs/evaluation/code.md
```diff
@@ -316,6 +316,43 @@ Due to variance between runs, you can automatically repeat the evaluation and av
...
+
+
+First, the data must be downloaded and prepared, which you can do by running:
+```bash
+ns prepare_data birdbench
+```
+
+This will download and unpack a file into `<output_directory>/dev_20240627`, which contains the BIRD dev manifest, table information, and database schemas. By default, `output_directory` will be under `nemo_skills/dataset/birdbench/`, though this can be changed via command line argument.
+
+The script will also process the original manifest into `<output_directory>/dev.jsonl`, which will be the input for evaluation.
```
gwarmstrong: is this needed now that we have `data_dir`? The rest of this is pretty consistent across evaluations.

redoctopus: Removed!

nemo_skills/evaluation/evaluator/bird.py
```diff
@@ -0,0 +1,149 @@
...
+        if dotall:
+            code_matches = re.findall(regex, text, flags=re.DOTALL)
+        else:
+            code_matches = re.findall(regex, text)
+
+        if not code_matches:
+            return "SELECT 1"  # No-op filler
+
+        # Remove comments first
+        ans = re.sub(r"--.*", "", code_matches[-1])  # Use last match
```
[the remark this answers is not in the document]
redoctopus: Editing, but using regex `r"--.*?|/\*.*?\*/"` instead as the suggested regex omits the `?` for the first pattern, which needs to be lazy instead of greedy in order to not capture non-comments.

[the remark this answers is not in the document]
redoctopus: Whoops, scratch that--should be `r"--.*?$|/\*.*?\*/"` instead (added end of string delimiter). So it's not too lazy.

PR #1135, opened 2025-12-19
generation.py to respect separate server type for the client

vmendelev:
This is needed when we want to untie model used by a client from the server type.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server configuration now respects user-specified settings. The system detects when users provide custom server type configurations and preserves those overrides instead of unconditionally applying defaults, for both hosted and remote deployment scenarios.


gwarmstrong (commented):
Do you think it could make more sense to swap the order of extra args and the configured overrides here? It might simplify this code and result in an overall more predictable experience.
e.g., change

```
f"{extra_arguments} ++server.server_type={server_type} ++server.host=127.0.0.1 "
f"++server.port={server_port} ++server.model={model} "
```

to 

```
f"++server.server_type={server_type} ++server.host=127.0.0.1 "
f"++server.port={server_port} ++server.model={model} {extra_arguments} "
```

PR #1137, opened 2025-12-19
move audio processing into model

gwarmstrong:
#### 1. `AudioProcessorConfig` (nested dataclass)
```python
class AudioProcessorConfig:
    data_dir: str = ""              # Base directory for audio files
    enable_chunking: bool = True    # Whether to chunk long audio
    chunk_task_types: list[str] | None = None
    chunk_threshold_sec: int = 30   # Chunk audio longer than this
```

#### 2. `AudioProcessor` (wrapper class)
```python
class AudioProcessor:
    def __init__(self, model, config, eval_config=None, eval_type=None):
        self.model = model
        # Resolves data_dir from config or eval_config
        ...
    
    async def generate_async(self, prompt, task_type=None, **kwargs):
        if isinstance(prompt, list):
            # Check if chunking needed
            # Convert audio files to base64
            # Handle chunking if audio is long
        return await self.model.generate_async(prompt=prompt, **kwargs)
```

#### 3. Integration in `generate.py`
```python
# In GenerateSolutionsConfig:
audio: AudioProcessorConfig | None = None  # Nested config, None = disabled

# In setup_llm():
if self.cfg.audio is not None:
    audio_supported_servers = {"vllm"}
    if server_type not in audio_supported_servers:
        raise ValueError(f"Audio not supported for {server_type}")
    llm = AudioProcessor(llm, self.cfg.audio, eval_config=..., eval_type=...)
```

### 1. Clean Separation of Concerns
- `VLLMModel`: ~150 lines, pure VLLM API client
- `AudioProcessor`: ~350 lines, all audio logic in one place

### 2. Composability
Audio processing wraps the base model, so it works with any underlying model:
```python
# Works with VLLM
llm = AudioProcessor(VLLMModel(...), config)

# Could work with our tool calling modules
llm = AudioProcessor(ToolCallingWrapper(VLLMModel(...)), config)

```

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -405,6 +405,13 @@ class GenerationTask:
 
         # Audio wrapper (preprocesses messages before they reach the model)
         if self.cfg.audio is not None:
+            audio_supported_servers = {"vllm"}
+            server_type = self.cfg.server.get("server_type", "").lower()
+            if server_type not in audio_supported_servers:
+                raise ValueError(
+                    f"Audio processing is not supported for server_type='{server_type}'. "
+                    f"Supported server types: {audio_supported_servers}"
+                )
             llm = AudioProcessor(
                 llm,
                 self.cfg.audio,


karpnv (approved):
LGTM

PR #1141, opened 2025-12-23
Fixing issues with HumanEval-Infilling

wasiahmad:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code evaluation by normalizing code completions with leading newlines. The system now correctly handles LLM outputs with leading newlines before fenced code blocks, ensuring consistent code processing and more reliable evaluation results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 *.json
+!greptile.json
 *.tar.gz
 *.tar
 *.npy
--- a/greptile.json
+++ b/greptile.json
@@ -1,0 +1,28 @@
+{
+  "strictness": 1,
+  "statusCheck": false,
+  "triggerOnUpdates": false,
+  "includeBranches": ["main", "chtruong/*"],
+  "instructions": "Review code following the guidelines in CONTRIBUTING.md.",
+  "commentTypes": ["logic", "syntax", "style"],
+  "summarySection": {
+    "visible": true,
+    "collapsed": true
+  },
+  "issuesTableSection": {
+    "visible": true,
+    "collapsed": false
+  },
+  "confidenceScoreSection": {
+    "visible": true,
+    "collapsed": true
+  },
+  "sequenceDiagramSection": {
+    "visible": false,
+    "collapsed": true
+  },
+  "fixWithAI": true,
+  "customContext": {
+    "files": ["CONTRIBUTING.md"]
+  }
+}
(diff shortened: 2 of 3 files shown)


nemo_skills/evaluation/evaluator/code.py
```diff
@@ -461,7 +461,15 @@ def postprocess_code(sample):
                 raise ValueError(f"All samples should have the same split, but got {data_split} and {sample['split']}")
 
             sample = preprocess_code(sample, language="python", strip_whitespace=False)
-            sample["original_completion"] = sample["completion"]
+            # ---------------------------------------------------------
+            # remove one leading "\n" from the completion
+            # because LLM generated solutions are in the form: ```python\n<fill_in_the_middle>```
+            completion = sample["completion"]
+            if completion.startswith("\n"):
+                completion = completion[1:]
```
gwarmstrong, line 469: is there a strong reason to do this, rather than use `strip_whitespace=True` in`preprocess_code`?

nemo_skills/inference/generate.py
```diff
@@ -291,7 +291,12 @@ def __init__(self, cfg: GenerateSolutionsConfig):
             cfg: GenerateSolutionsConfig object with the configuration parameters or subclass.
         """
         self.cfg = cfg
-        self.cfg.inference.extra_body = dict(self.cfg.inference.extra_body)
+        if OmegaConf.is_config(self.cfg.inference.extra_body):
```
gwarmstrong: can you add a comment about what exactly is going on here and the issues it prevents?

nemo_skills/evaluation/evaluator/code.py
wasiahmad: Yes @gwarmstrong because FIM is a separate case and we cannot do `strip_whitespace=True`, we only check for preeceeding newlines (I noticed a few LLMs start with a newline by default).

nemo_skills/inference/generate.py
wasiahmad: Sure.

Kipok: @wasiahmad do you want to do this, so that we can merge this?

Kipok: should we maybe remove this change? It doesn't look relevant for human eval fixing and I also don't fully understand the logic and why it's needed @wasiahmad

PR #1142, opened 2025-12-25
Fix fail to run Nemo Skills when need trust remote code

fzyzcjy:
I think Nemo-Skills is run by users only internally w/ control, thus not make it configurable. Please ping me if this is needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tokenizer initialization to enable remote code execution support when loading tokenizers from string identifiers.


Kipok:
thanks @fzyzcjy! Re-created here https://github.com/NVIDIA-NeMo/Skills/pull/1146 from a branch to be able to run ci and also added this parameter in a few more places

PR #1147, opened 2025-12-27
Fix RULER incompatible with thinking models

fzyzcjy:
## Summary by CodeRabbit

* **New Features**
  * Added `--use_chat_completions` CLI flag to toggle between different data preparation modes based on model requirements.
  * Chat completions mode changes how answer prefixes are processed and integrated during dataset preparation for improved compatibility.
  * Dual workflow support provides flexible configuration options for formatting model training data to match various completion API standards.

### DIFF
--- a/nemo_skills/dataset/ruler/prepare.py
+++ b/nemo_skills/dataset/ruler/prepare.py
@@ -66,8 +66,8 @@ def prepare_task_for_ns(task, data_dir, setup, use_chat_completions):
         short_name = task.split("_")[0]
         init_file.write(
             DEFAULT_SETTINGS.format(
-                tokens_to_generate=TOKENS_TO_GENERATE[short_name],
                 match_type=MATCH_TYPE[short_name],
+                tokens_to_generate=TOKENS_TO_GENERATE[short_name],
                 extra_generation_args="" if use_chat_completions else TEXT_COMPLETIONS_EXTRA_ARGS,
             )
         )


nemo_skills/dataset/ruler/prepare.py
```diff
@@ -28,31 +28,38 @@
...
     new_path = Path(__file__).parent / setup / task / "test.jsonl"
     Path(new_path).parent.mkdir(parents=True, exist_ok=True)
     with open(original_path, "r", encoding="utf-8") as fin, open(new_path, "w", encoding="utf-8") as fout:
         for line in fin:
             original_entry = json.loads(line)
+            answer_prefix = original_entry["answer_prefix"].strip()
             new_entry = {
                 "index": original_entry["index"],
-                "question": original_entry["input"],
+                "question": original_entry["input"] + (answer_prefix if use_chat_completions else ""),
```
hsiehjackson, line 57: I think we can remove answer prefix when we do chat completion mode.

Kipok:
is this not supported in that PR? E.g. with "default" data format, it will mimic the paper very closely (which is the current behavior on main) and with "chat" data format, it will be almost identical to your version with the only difference that answer prefix isn't appended to the prompt, which is probably reasonable (as model is unlikely to follow it anyway and will be burning many tokens on thinking)? Or do you think that not appending answer prefix is not the right way to run it through chat completions? Or is there some other difference I'm missing?

PR #1156, opened 2026-01-07
Comet metrics for machine translation

AlexGrinch:
Adding Comet evaluator to compute xCOMET-XXL score for generated translations.

## Summary by CodeRabbit

* **New Features**
  * Added xCOMET-XXL based evaluation for machine translation outputs
  * Support for single-file and batch directory processing
  * Automatic package installation and model loading with CUDA support
  * Configurable batch processing for scoring results


Kipok (commented):
thanks @AlexGrinch ! Looks like quite a few of automatic comments are real issues that need to be fixed.

More generally, where is this used? Should this be exposed as an option in some benchmarks?

Kipok (approved):

PR #1162, opened 2026-01-12
add ray template support

wedu-nvidia:
Add ray template support

## Summary by CodeRabbit

* **New Features**
  * Added `ray_template` CLI option to SFT and GRPO training commands. Users can now specify a Slurm Ray template that will be applied during training execution.

### DIFF
--- a/nemo_skills/pipeline/nemo_rl/grpo.py
+++ b/nemo_skills/pipeline/nemo_rl/grpo.py
@@ -340,9 +340,8 @@ def grpo_nemo_rl(
     cluster_config = get_cluster_config(cluster, config_dir)
     cluster_config = resolve_mount_paths(cluster_config, mount_paths)
 
-    # Read ray_template from env_vars
-    env_variables = get_env_variables(cluster_config)
-    ray_template = env_variables.get("RAY_TEMPLATE", None)
+    # Read ray_template from cluster config
+    ray_template = cluster_config.get("ray_template", None)
 
     if log_dir is None:
         log_dir = output_dir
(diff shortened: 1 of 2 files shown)


Kipok (changes requested):
let's do this through cluster config instead. So nemo_run_ray_template would be a top-level variable there, so that this can be set one time and not for each job. Will need to read that and set appropriately in the code

@hemildesai do you think we can support this through a parameter on nemo-run side? Otherwise we'd have to set env var and then roll it back to not contaminate other jobs that could be launched on different cluster within same python session

nemo_skills/pipeline/nemo_rl/grpo.py
```diff
@@ -327,6 +327,11 @@ def grpo_nemo_rl(
         None,
         help="Max position embeddings to use for conversion. If not specified, will be inferred from the model config.",
     ),
+    ray_template: str = typer.Option(
```
Kipok: can we read this from cluster config instead? So define ray_template variable directly in the config and read it

wedu-nvidia: yes, sure, I will update

nemo_skills/pipeline/nemo_rl/grpo.py
```diff
@@ -340,6 +340,10 @@ def grpo_nemo_rl(
     cluster_config = get_cluster_config(cluster, config_dir)
     cluster_config = resolve_mount_paths(cluster_config, mount_paths)
 
+    # Read ray_template from env_vars
+    env_variables = get_env_variables(cluster_config)
```
Kipok: instead of taking it from env variables, let's just read it from the config itself. So just cluster_config.get["RAY_TEMPLATE"]. Using env variables might have unintended effects of persisting between clusters if launched from the same job

wedu-nvidia: Sure, just updated as above suggestions.

PR #1166, opened 2026-01-13
Fix RULER to be compatible with Base/Chat/CoT models

hsiehjackson:
* Added `--data_format` to select RULER data format from `default`, `base`, or `chat`. 
  * `default`: we add answer prefix and use text completion mode.
  * `base`: we add answer prefix with a newline separator and use text completion mode.
  * `chat`: we remove answer prefix and use chat completion mode.
* We reduce 500 samples to 100 samples.
* We remove the default hf tokenizer to prepare RULER data.

## Summary by CodeRabbit

* **New Features**
  * Added `--data_format` option for ruler data preparation supporting default, base, and chat formats.

* **Chores**
  * Refined data preparation workflow and configuration handling.
  * Optimized sampling parameters in the preparation pipeline.

### DIFF
--- a/docs/evaluation/code.md
+++ b/docs/evaluation/code.md
@@ -238,6 +238,67 @@ The benchmark reports:
 - **pass@1**: Same as accuracy for single-solution generation
 - **pass@k**: Success rate when generating k solutions per problem (if configured)
 
+### swe-bench-multilingual
+
+- Benchmark is defined in [`nemo_skills/dataset/swe-bench-multilingual/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/swe-bench-multilingual/__init__.py)
+- Original benchmark source is [here](https://www.swebench.com/multilingual.html).
+
+SWE-bench Multilingual uses mostly the same logic as regular SWE-bench, so most of the [SWE-bench docs](#swe-bench) apply to it as well. The differences are as follows:
+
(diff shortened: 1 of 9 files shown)


nemo_skills/dataset/ruler/prepare.py
```diff
@@ -200,12 +222,21 @@ def prepare_task(task):
             "--tokenizer_path meta-llama/Llama-3.1-8B-Instruct --max_seq_length 131072"
         )
         exit(0)
-    print(f"Preparing RULER dataset for tasks: {args.tasks} with additional arguments: {ruler_prepare_args}")
+
+    if "--tokenizer_type" not in ruler_prepare_args:
```
Kipok: shouldn't this work automatically? If it's in the default command and provided again by a user, the user's option should be used

nemo_skills/dataset/ruler/prepare.py
```diff
@@ -188,6 +199,17 @@ def prepare_task(task):
         default=None,
         help="Directory to store intermediate data. If not provided, a temporary directory will be created.",
     )
+    parser.add_argument(
+        "--data_format",
```
Kipok, line 203: Please update the docs to mention this new parameter and when it should be set to chat

nemo_skills/dataset/ruler/prepare.py
hsiehjackson: Fixed

nemo_skills/dataset/ruler/prepare.py
hsiehjackson: Updated

Kipok (approved):

PR #1170, opened 2026-01-16
Add support for custom environments per generation module

Kipok:
We have some conflicting requirements, e.g. compute-eval needs tree_sitter that's different version from bfcl_eval. To address the problem, this pr adds an on-the-fly requirement installation that can be defined per generation module

## Summary by CodeRabbit

* **New Features**
  * Generation modules can now declare their specific package requirements for isolated execution.
  * Enhanced generation pipeline with automatic virtual environment setup and per-module dependency management.

### DIFF
--- a/nemo_skills/inference/eval/bfcl.py
+++ b/nemo_skills/inference/eval/bfcl.py
@@ -310,7 +310,7 @@ class ServerMessageParser:
 
         # Use model output if not a tool call
         if isinstance(output_dict["message"], dict):
-            message_content = output_dict["message"].get("content", "")
+            message_content = output_dict["message"]["content"]
         else:
             message_content = output_dict["message"].content
         output_dict["generation"] = generation if generation else [message_content]
@@ -322,7 +322,7 @@ class ServerMessageParser:
 
     def get_response_text(self, message):
         if isinstance(message, dict):
-            return message.get("content")
+            return message["content"]
         return message.content
 
     def set_response_text(self, message, response_text):


nemo_skills/pipeline/utils/generation.py
```diff
@@ -102,6 +104,51 @@ def normalize_parameter(
...
+        f"{requirements_content}\n"
+        "NEMO_SKILLS_REQS_EOF\n"
+        "fi && "
+        f"VENV_ROOT={venv_root_arg} && VENV_DIR={venv_dir_arg} && "
+        'READY_FILE="$VENV_DIR/.nemo_skills_ready" && '
+        'LOCK_DIR="$VENV_DIR.lock" && '
+        'mkdir -p "$VENV_ROOT" && '
+        'if [ ! -f "$READY_FILE" ]; then '
+        '  if mkdir "$LOCK_DIR" 2>/dev/null; then '
+        '    if ! uv venv --system-site-packages "$VENV_DIR"; then rmdir "$LOCK_DIR"; exit 1; fi; '
```
gwarmstrong, line 139: as I understand it, this copies the system packages into the venv. I thought there was conflict with compute_eval dependencies and bfcl dependencies. Why does that not cause an issue here? Is it because it is not trying to satisfy the dependencies simultaneously?

nemo_skills/dataset/bfcl_v3/prepare.py
```diff
@@ -17,11 +17,40 @@
...
+def ensure_bfcl_eval_installed():
+    try:
+        import bfcl_eval  # noqa: F401
+
+        return
+    except (ModuleNotFoundError, ImportError, AttributeError):
+        logging.getLogger(__name__).info("bfcl_eval not found. Installing runtime dependency...")
+        try:
+            subprocess.run(
+                [sys.executable, "-m", "pip", "install", BFCL_EVAL_PIP_PACKAGE],
```
gwarmstrong: what would cause the pip install to fail? why are we trying both options and not sticking with one?

nemo_skills/pipeline/utils/generation.py
Kipok: Yes, we don't ever need both of them to be active for the same generation module. So currently the parent environment has correct versions for compute_eval and the bfcl-specific environment should override them with different versions, breaking compute_eval in that venv

nemo_skills/dataset/bfcl_v3/prepare.py
Kipok: let me clean this up, we shouldn't need two

nemo_skills/inference/eval/bfcl.py
```diff
@@ -243,34 +282,61 @@ def construct_input_dict(self, messages: list[dict], tools: list[dict]):
...
+        output_dict["generation"] = generation if generation else [message_content]
         output_dict["tool_calls"] = tool_calls
         output_dict["tool_call_ids"] = tool_call_ids
         output_dict["num_generated_tokens"] = output_dict.get("num_generated_tokens", 0)
 
         return output_dict
 
     def get_response_text(self, message):
+        if isinstance(message, dict):
+            return message.get("content")
```
gwarmstrong: should we use `.get()` or `message["content"]` here?

Kipok: fixed, thanks

PR #1180, opened 2026-01-22
Allow SERPAPI for BFCLv4

Glorf:


nemo_skills/inference/eval/bfcl_web_search.py
```diff
@@ -22,6 +23,7 @@
 from bs4 import BeautifulSoup
 from ddgs import DDGS
 from ddgs.exceptions import DDGSException
+from serpapi.google_search import GoogleSearch
```
Kipok: since this is a somewhat optional requirement, let's maybe import it inside the relevant method? Just to avoid breaking things for people using older container

Kipok (commented):
just a small comment, otherwise lgtm

nemo_skills/inference/eval/bfcl_web_search.py
Glorf: Good point - maybe we can do dynamic imports for both ddgs and serpapi, so that only one of them is required to be installed. Let me change that

Glorf: @Kipok  done, I also added additional check for whether these modules are available

PR #1183, opened 2026-01-23
New contribution guidelines

Kipok:
## Summary by CodeRabbit

* **Documentation**
  * Expanded contribution guidelines with comprehensive general guidelines and development best practices.
  * Added new sections on benchmarking, code elegance principles, and pre-commit hooks setup instructions.
  * Enhanced documentation to support improved developer collaboration and streamlined development workflows.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,8 @@ The following things are required when adding new benchmarks
 - Avoid data loss because of evaluation mistakes. Our current design overrides the original
   generation files with new evaluation-specific keys. Make sure to do all computation before
   re-opening the files for writing to avoid accidental data loss if there is a bug and code fails before writing is complete.
-- Run GPU tests in the CI (or locally). To run in CI, we need to set "run GPU tests" label.
+- Run GPU tests in the CI (or locally). To run in CI, we need to set "run GPU tests" label
+  (toggle it off and back on if rerunning after changes).
   By default all datasets will be prepared and evaluated on a few samples in the CI. You can
   remove your dataset from the test explicitly if it requires very heavy data preparation or
   has another reason why we can't use it. But try to avoid that if possible!


CONTRIBUTING.md
```diff
@@ -2,23 +2,100 @@
...
+The following things are required when adding new benchmarks
+- Add it to a corresponding place in the documentation.
+  Make sure to add an example command for how to run evaluation and expected results for any model you tested it with.
+  Describe any details that are specific to this dataset. Any special arguments to prepare data?
+  Any non-standard inference arguments? Any other things to pay attention to?
+- Don't forget to run `mkdocs serve` and visually check that the documentation renders properly in the browser.
+- Avoid data loss because of evaluation mistakes. Our current design overrides the original
+  generation files with new evaluation-specific keys. Make sure to do all computation before
+  re-opening the files for writing to avoid accidental data loss if there is a bug and code fails before writing is complete.
+- Run GPU tests in the CI (or locally). To run in CI, we need to set "run GPU tests" label.
```
agronskiy: `nit`: IIRC, the label needs to be explicitly toggled off/on to rerun tests. Was not obvious initially.

CONTRIBUTING.md
```diff
@@ -2,23 +2,100 @@
...
-## Code Quality
+## General guidelines
+
+Applies to both humans and AI agents! Make sure the agent you use has access to these guidelines when implementing or reviewing Nemo-Skills features.
+
+### Don't be overly defensive!
+
+Being overly defensive can lead to silent misbehavior for our code which is a lot worse than a quick runtime failure. Here are some examples to avoid
+
+- Don't use `.get` for accessing dictionary keys if the code expects them to be present.
```
agronskiy, line 13: `thinking loud`: the "code needs something to be present in a dict" is a hard-to-reason-about factology that requires an explicit assert or similar failing check. Advanced LSPs (ty/pyright) might bark at the code that both does not use .get and tries to access smth.

`suggestion`: add an instruction to start guarding those things by checks.

Kipok: I'm not fully sure I understand. If we access a field with data[key_name], that means it has to be present and if it's not, the code will fail with a clear error. Do you think we need an extra assert or a check for such cases? The proposal here is to just not use .get for things where the code has to fail, which is a very common issue in ai generated code these days

CONTRIBUTING.md
Kipok: thanks, updated!

CONTRIBUTING.md
Jorjeous: Sometimes this field is just helpfull, not required
Example: "duration" key - many models jsut do not need it to properly run, therefore some may even not include this  in manifest.

Jorjeous: @Kipok about comment above: It's about possibility that AI guards/checks/ other CICD stuff would trigger on any accessing without get [and suggest .get addition]

and reasonable suggestion to tune our guards and other tests to check that get is ont present (may be drop warning it tests f.e.)

Kipok: Of course if field is optional than we should use .get. Only for fields that are required and should always be present unless there was some error, we shouldn't hide this and should just let the code fail. Let's see if the current wording is clear for ai agents and if not we will refine it

agronskiy: Sry folks I was unclear: e.g. pyright with `strictDictionaryInference` has the ability to trace what was assigned etc, I erroneously assumed that it would flag things like `x = d["non-existent-key"]`, annoying everyone. In this case, _explicitly_ expecting AI / developer to write `assert "key" in d` would make it both runtime- and static-analysis-time proof.

Kipok: I still don't fully understand. The code before should always be assigning that key, so it should be `x = d["existing-key"]`. We just want to make sure we don't do `x = d.get("esisting-key", "random_default")` to avoid this causing silent failures if there are bugs that cause key to not be there or previous job crashing because of cluster issues or things like this. Is extra assert still needed here? Feels like it's just going to bloat the code and also not super visible anyway (as we don't want to have it in all possible places this key is used, so it will be in some random place in the code, not necessarily where the things are going to fail). And it feels very clear that if you use d[<>], then <> must exist, so extra assert shouldn't be adding anything?

Kipok: Let me merge as is, but we can continue discussing here (or offline) and can create a follow up pr with updated guidelines

PR #1184, opened 2026-01-26
 Add proper token counting to code execution model

i-vainn:
## Changes                                                                                                                                          
- `get_code_execution_model` now passes `require_tokenizer=True` to ensure tokenizer availability                                                   
- `CodeExecutionWrapper` now uses tokenizer to count tokens for:                                                                                    
  - Code execution output added to prompts                                                                                                          
  - Streaming segment token counts                                                                                                                  
- Added RuntimeError with clear message if tokenizer is unavailable                                                                                 
- Fixed `_initialize_tokenizer` to catch exceptions and allow fallback to server endpoint  

## Release Notes

* **Bug Fixes**
  * Enhanced tokenizer initialization with improved error handling and fallback mechanisms
  * Improved token counting accuracy during code execution and generation to ensure proper token budget tracking
  * Added runtime validation to ensure tokenizer availability during generation processes

### DIFF
--- a/cluster_configs/example-local.yaml
+++ b/cluster_configs/example-local.yaml
@@ -15,9 +15,9 @@
 executor: local
 
 containers:
-  trtllm: nvcr.io/nvidia/tensorrt-llm/release:1.0.0
-  vllm: vllm/vllm-openai:v0.10.1.1
-  sglang: lmsysorg/sglang:v0.5.4
+  trtllm: nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc1
+  vllm: dockerfile:dockerfiles/Dockerfile.vllm
+  sglang: lmsysorg/sglang:v0.5.8
   # dockerfile: for now can only specify relative to repo root
   megatron: dockerfile:dockerfiles/Dockerfile.megatron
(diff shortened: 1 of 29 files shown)


nemo_skills/inference/model/base.py
```diff
@@ -202,7 +204,11 @@ def _initialize_tokenizer(self, tokenizer: str | None) -> WrapperAutoTokenizer |
         if tokenizer is None:
             return None
         if isinstance(tokenizer, str):
-            return WrapperAutoTokenizer(tokenizer)
+            try:
+                return WrapperAutoTokenizer(tokenizer)
+            except (OSError, Exception) as e:
```
Kipok: why do we need this? If we were asked to get tokenizer, but it failed, we should just fail completely, no?

i-vainn: This is intentional - see https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/inference/model/base.py#L177-L195 which tries to initialize the tokenizer from the string, otherwise initialize the tokenizer endpoint.

Without this catch, server model path crashes before reaching the server endpoint fallback.

PR #1186, opened 2026-01-26
support structured outputs in hle judge for optional AA compatibility

anowaczynski-nvidia:
**Why?**
Enable optional AA-compatibility in HLE benchmark implementation.

**What?**
Support structured output in generations. See https://platform.openai.com/docs/guides/structured-outputs

**How?**
- new file `nemo_skills/inference/structured_outputs.py` with predefined response formats for structured outputs
- new parameter in GenerationTaskConfig: `structured_output` str, default None
- in `process_single_datapoint` before generation: add response_format to generation_params based on `self.cfg.structured_output` if not None
- in `postprocess_single_output` prase generation and extract `correct` field in the expected judgement format

## Summary by CodeRabbit

* **New Features**
  * Added structured output support to inference generation, enabling formatted JSON responses from compatible models
  * Automatic response parsing and field extraction for improved result handling
  * Response format configuration with consistent API across supported model providers

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -691,7 +691,7 @@ class GenerationTask:
             "stop_phrases": [self.cfg.stop_phrase] if self.cfg.stop_phrase else None,
         }
 
-        if self.cfg.structured_output in STRUCTURED_OUTPUTS:
+        if self.cfg.structured_output is not None:
             generation_params["response_format"] = STRUCTURED_OUTPUTS[self.cfg.structured_output]
 
         if self.cfg.code_execution:
--- a/tests/test_generation.py
+++ b/tests/test_generation.py
@@ -206,7 +206,7 @@ def test_judge_generations_with_structured_output(tmp_path):
         f"    --judge_server_address=https://integrate.api.nvidia.com/v1 "
         f"    --judge_server_type=openai "
         f"    --metric_type=hle-aa "
-        f'    --extra_judge_args="++structured_output=HLE_JUDGE_AA"'
+        f'    --extra_judge_args="++structured_output=HLE_JUDGE_AA" '
         f"    ++max_samples=2 "
         f"    ++inference.tokens_to_generate=1024 "  # to make test go fast
     )


nemo_skills/inference/generate.py
```diff
@@ -630,6 +633,14 @@ async def postprocess_single_output(self, output, original_data_point):
         # all of the original data to the output file alongside the new generations
         output[self.cfg.generation_key] = output.pop("generation")
 
+        if self.cfg.structured_output == "HLE_JUDGE_AA":
```
jiacheng-xu: It is not be a good idea to hard code HLE_JUDGE_AA in `generate.py`.
Can we build a function to handle that like https://github.com/NVIDIA-NeMo/Skills/blob/54c8bc0fa980b248a977475e2ddf00df73aa81f6/nemo_skills/inference/generate.py#L658 ?

Kipok: @anowaczynski-nvidia can we move this logic into metrics? Why does it need to be in the generation?

anowaczynski-nvidia: Reasons I added if with postprocessing here:
- to enable AA-compatible HLE judge, `++structured_output=HLE_JUDGE_AA` needs to be added only in one place (judge generations pipeline command)
- with the current version `summarize_results` command and pipeline logic for aggregating hle judge outputs into metrics doesn't require any modifications (the same command + code handles both: default and AA-compatible judges)

I am aware this code is fundamental to the entire package, all generations pass through it.

Regarding moving this to metrics: I see the possibility to create hleaa_metrics.py in evaluation/metrics, inherit from MathMetrics, and override only `_get_score_dict`, such that postprocessing of judgement (parsing into json etc) is applied before `is_correct_judgement`. Do you approve this plan?

Kipok: Yes, either that or we can just have this as option for main math metrics, so that any dataset, not just HLE can be evaluated in this setup. The one problem is I am not fully sure if metrics are currently customizable, but I guess if not, then we should enable customization in a similar way to how it's done for eval / generation parameters. Let me know if you need help with the design on that, happy to discuss in more details

nemo_skills/inference/structured_outputs.py
```diff
@@ -0,0 +1,28 @@
...
+class HLEJudgeAAResponseFormat(BaseModel):
+    extracted_final_answer: str
+    reasoning: str
+    correct: Literal["yes", "no"]
+    confidence: int
+
+
+STRUCTURED_OUTPUTS = {
+    "HLE_JUDGE_AA": HLEJudgeAAResponseFormat,
+}
```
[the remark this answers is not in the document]
anowaczynski-nvidia: it's supported.
There are many examples in the openai docs: https://platform.openai.com/docs/guides/structured-outputs
And verified empirically as well.

nemo_skills/inference/generate.py
anowaczynski-nvidia: @Kipok I tried the hard way first, but nothing I created was correct and convincing, so I pushed one commit with `class HLEAAMetrics(MathMetrics)` solution as it was conceptually much simpler. The main downside is that I had to add `metric_type` to eval command. It doesn't look right here. It doesn't compose with eval on multiple benchmarks idea. Can you take a look? If we're doing Metrics Config idea, I need a sync how to approach it.

Kipok: I think this is the right approach. When doing eval on multiple benchmarks you can't really customize anything except maybe inference parameters. E.g. doing prompt change or eval arguments will also break things, so I think adding metric_type is a good change. An alternative would be to add this as an argument to MathMetrics and then you can reuse existing metric_kwargs parameter to customize it. But adding metric_type is a good change anyway given that we support metric_kwargs already.

If the current implementation fully works for you, I think it LGTM as well and we can merge it. But do let me know if you have any concerns or think we should do things differently

Kipok: It's probably a good idea to add a new test for this in test_generation.py, but only if models on build.nvidia.com support this response_format argument

anowaczynski-nvidia: added `test_judge_generations_with_structured_output` but it takes 10 minutes to complete even with max_samples=2, obviously this can't be merged, but where do we go from here?

Kipok: thanks @anowaczynski-nvidia - pushed a change to limit max tokens (since we aren't checking generation correctness anyway), seems to finish very fast now!

PR #1192, opened 2026-01-27
Fixes for slurm tests

Kipok:
1. Removed bfcl from llama-nemotron as its parser is no longer supported
2. Adjust bfcl install to be non-editable (was failing in cron job)
3. Improve stability for scicode download by switching to HF path (was failing with rate limits)
4. Adjust different constraints / fix typos after some api changes

## Summary by CodeRabbit

* **Chores**
  * Updated dataset preparation to use HuggingFace dataset loading for SciCode.
  * Improved virtual environment setup with deterministic seeding for test reproducibility.
  * Removed editable installation flag from BFCL runtime setup.

* **Tests**
  * Updated evaluation metric ranges and validation thresholds.
  * Refined tool-calling evaluation configuration and removed redundant checks.
  * Adjusted model evaluation framework for improved testing workflows.

### DIFF
--- a/nemo_skills/inference/eval/bfcl.py
+++ b/nemo_skills/inference/eval/bfcl.py
@@ -49,11 +49,12 @@ LOG = logging.getLogger(get_logger_name(__file__))
 
 BFCL_REQUIREMENTS = [
     # Source: https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/pyproject.toml
+    # we made changes listed in comments to speed up installation time
     "requests",
     "tqdm",
     "numpy==1.26.4",
     "pandas",
-    "huggingface_hub",
+    "huggingface_hub<1",  # pin to <1 to satisfy a requirement from another package
     "pydantic>=2.8.2",
     "python-dotenv>=1.0.1",
     "tree_sitter==0.21.3",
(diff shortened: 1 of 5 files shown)


nemo_skills/dataset/scicode/prepare.py
```diff
@@ -13,47 +13,27 @@
...
 
+    split_mapping = {"validation": "dev", "test": "test"}
+    all_data = []
+    for hf_split, output_split in split_mapping.items():
+        output_file = data_dir / f"{output_split}.jsonl"
         with open(output_file, "wt", encoding="utf-8") as fout:
-            for entry in data:
+            for entry in dataset[hf_split]:
                 fout.write(json.dumps(entry) + "\n")
+                all_data.append(entry)
```
gwarmstrong: probably a very minor optimization, but can't we write to `test_aai_file` at the same time rather than accumulating and appending? not sure how large the data is, so might not make a big difference in the end

tests/slurm-tests/super_49b_evals/run_test.py
```diff
@@ -142,31 +142,10 @@ def eval_reasoning_on(workspace, cluster, expname_prefix, wandb_project):
         wandb_name=f"{expname_prefix}-super_49b-eval-reasoning-on",
     )
 
-    # BFCL (Reasoning ON)
-    eval(
```
gwarmstrong, line 146: Is there a specific reason to remove?

Kipok: yes, latest vllm doesn't support llama-nemotron reasoning parser anymore, so that test fails and it's not really worth fixing as we should just switch to a different model

nemo_skills/dataset/scicode/prepare.py
Kipok: updated

gwarmstrong (approved):
LGTM

PR #1200, opened 2026-01-30
Add CritPt benchmark

jiacheng-xu:
Add CritPt benchmark 

Two-Turn Generation Flow:
Turn 1: Model generates step-by-step physics solution with LaTeX formatting
Turn 2: Model populates solution into provided code template

Key Components:

- Dataset (dataset/critpt/): Downloads from HuggingFace, formats with problem descriptions and code templates
- Inference (inference/eval/critpt.py): Custom generation task with ChatML parsing, tracks tokens per turn
- Evaluator (evaluation/evaluator/critpt.py): Submits predictions to Artificial Analysis API, caches responses
- Metrics (evaluation/metrics/critpt_metrics.py): Reports accuracy, tokens, generation time

## Summary by CodeRabbit

* **New Features**
  * CritPt benchmark integration with a two-turn generation workflow and Hydra entrypoint
  * Batch submission to an external evaluation API with caching and aggregated accuracy reporting
  * New metrics registration for CritPt-style aggregate scoring
  * Prompt templates for physics problem-solving with strict formatting and final-answer rules
  * Export utility to produce CritPt dataset entries for evaluation
* **Tests**
  * Excluded CritPt from preparable GPU tests
* **Chores**
  * Added configurable defaults for CritPt evaluation and generation (overridable via CLI)


jiacheng-xu:
> @jiacheng-xu can you show an example `ns generate`/`ns eval` script that you are using? I suspect the issue would be if you are setting `++code_execution=True`--I suspect that's not what you actually want here, as that is an older way of calling python execution before we had tool parsers/structured tool calls integrated.

Yes I am using code_execution=True.
```
{
  "benchmarks": "critpt:1",
  "split": "test",
  "num_chunks": 1,
  "server_type": "vllm",
  "server_gpus": 4,
  "server_args": "--async-scheduling",
  "model": "/hf_models/gpt-oss-20b",
  "with_sandbox": true,
  "expname": "critpttest-ghb-model_gpt_oss_20b-oci-debug",
  "cluster": "oci",
  "output_dir": "/workspace/critpttest-ghb-model_gpt_oss_20b-oci-debug",
  "wandb_project": "critpt",
  "wandb_name": "012918-critpttest-ghb-model_gpt_oss_20b-oci",
  "__ctx_args": "++max_samples=10 ++inference.temperature=1.0 ++inference.tokens_to_generate=65536 ++code_tags=gpt-oss ++server.code_execution.max_code_executions=100 ++inference.endpoint_type=text --config-path=/nemo_run/code/nemo_skills_stem/configs --config-name=gpt_oss ++chat_template_kwargs.builtin_tools=[python] ++chat_template_kwargs.reasoning_effort=high ++parse_reasoning=True ++code_execution=true"
}
```
Could you share what's the recommended way for tool calling, one for endpoint='text' (like gpt oss) and one for endpoint='chat'? Thanks!

jiacheng-xu:
@gwarmstrong 
Tool calling on HLE with GPT OSS doesn't work in this case:
4`"++inference.temperature=1.0 ++inference.tokens_to_generate=65536 ++inference.extra_body.reasoning_effort=high ++tool_modules=[\"nemo_skills.mcp.servers.python_tool::PythonTool\"] ++chat_template_kwargs.enable_thinking=True"`

Example: "generation": "import sys, subprocess, json, pkgutil, textwrap, os, importlib, types, inspect, platform, math",
Example:  "generation": "{\"code\":\"import math\\nn=25\\nfactorial=math.factorial(n)\\n\\n# generate partitions of n as nonincreasing tuples\\n\\n# .......

I tried to turn on parse_reasoning, then the generation are all empty.

nemo_skills/prompt/config/eval/critpt/code_output.yaml
```diff
@@ -0,0 +1,6 @@
+
+user: |-
+  Populate your final answer into the code template provided below.
+  This step is purely for formatting/display purposes. No additional reasoning or derivation should be performed.
+  Do not import any modules or packages beyond what is provided in the template.
+  ```python\n{code_template}\n```
```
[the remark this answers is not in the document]
Kipok: @jiacheng-xu please fix this

nemo_skills/prompt/config/eval/critpt/solve_problem.yaml
```diff
@@ -0,0 +1,30 @@
...
+  2. **Mathematical Typesetting** - Use LaTeX for all mathematics:  `$...$` for inline expressions, `$$...$$` for display equations.
+
+  3. **Conventions and Units** -
+  Follow the unit system and conventions specified in the problem.
+
+  4. **Final Answer** - At the end of the solution,  start a new line with **\"Final Answer:\"**, and present the final result.
+
+  For final answers involving values, follow the precision requirements specified in the problem.
+  If no precision is specified:
+  - If an exact value is possible, provide it (e.g., \\$\\sqrt(2)\\$, \\$\\pi/4\\$).
```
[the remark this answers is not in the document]
Kipok: @jiacheng-xu please fix this

nemo_skills/evaluation/evaluator/critpt.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
+        # Distribute correctness based on aggregate accuracy
+        # Since API doesn't provide per-example results, we assign first n examples as correct
+        total_examples = len(data_points)
+        num_correct = int(accuracy * total_examples)
+
+        LOG.info(
+            f"Marking first {num_correct} out of {total_examples} examples as correct based on {accuracy:.2%} accuracy"
+        )
```
Kipok: agree with ai reviewers, I think this is confusing and will cause problems. Maybe instead we write aggregate_accuracy to all rows, so that it's always the same? Can even name it full_dataset_accuracy or something like this so that it's extra clear. And then just handle appropriately in metrics

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,193 @@
...
+    messages = []
+    for role, content in pattern.findall(prompt):
+        messages.append({"role": role.strip(), "content": content.strip()})
+    return messages
+
+
+# Like nemo_skills.inference.generate.InferenceConfig, except most parameters are not passed by default
+# because they may not be supported by all LLM servers.
+@nested_dataclass(kw_only=True)
+class CritPtInferenceConfig:
```
Kipok, line 41: let's maybe skip this for now for consistency? I agree this is better, but I'd rather update our main inference config than have some evaluation benchmarks treating it differently unless this is absolutely necessary. I will push a separate PR updating defaults in main config if that's ok

nemo_skills/dataset/critpt/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+
+DATASET_GROUP = "math"
+METRICS_TYPE = "critpt"
+GENERATION_ARGS = "++prompt_config=eval/critpt/solve_problem ++eval_type=critpt"
```
Kipok: you don't have to set prompt config if custom generation module is used where it's already default

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,193 @@
...
+        # ===== Turn 1: Generate solution for the problem =====
+        # In this round, we still use the NS prompt filling pipeline.
+
+        # Since process_single_datapoint did not return the prompt, we fill it here manually.
+        # turn1_prompt contains the system message, and the user message.
+        turn1_prompt = self.fill_prompt(data_point, all_data)
+        LOG.debug(f"Turn 1 prompt: {turn1_prompt}")
+
+        if isinstance(turn1_prompt, str):
+            turn1_prompt = parse_chatml(turn1_prompt)
```
Kipok: could you please explain why we need this?

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,193 @@
...
+            LOG.debug(f"Turn 1 prompt (parsed): {turn1_prompt}")
+
+        turn1_result = await super().process_single_datapoint(data_point, all_data)
+        LOG.debug(f"Turn 1 result: {turn1_result}")
+        #
+        if self.cfg.end_reasoning_string in turn1_result[self.cfg.generation_key]:
```
Kipok: this should only be done if parse_reasoning is set to true

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,193 @@
...
+        solution_turn1 = turn1_result[self.cfg.generation_key]
+        LOG.debug(f"Solution: {solution_turn1}")
+
+        assistant_msg: dict = {"role": "assistant", "content": solution_turn1}
+
+        LOG.debug(f"Assistant: {assistant_msg}")
+
+        # ===== Turn 2: Generate code using template =====
+        # Build prompt that includes code template
+        turn2_prompt_template = load_config(self.cfg.prompt_config_turn2)["user"]
```
Kipok: can we call get_prompt here?

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,193 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
Kipok, line 1: I feel that this implementation is a little more complex than necessary. I think we should try to avoid having a `_process_single_completion` which is almost exactly a copy-paste from the base class. And the current management of prompts is a bit confusing as there is a bunch of logic to work-around the restrictions on the base class. For such cases, I think it's much better to just operate with prompt_format=openai by default, so that you can completely sidestep our prompt filling logic. You can still read the prompt config if you want and use it inside your custom process datapoint function, but that way the function itself will just expect a messages list, so you can hopefully reuse it directly without the current second copy. Do you think you could do another version with the above changes? Happy to discuss more if something isn't clear

nemo_skills/inference/eval/critpt.py
jiacheng-xu: Yes the base InferenceConfig needs to be updated for gpt (both number models and o models)/gemini/claude. I can remove this and the SWEBench one might also be removed.

nemo_skills/inference/eval/critpt.py
jiacheng-xu: it's for gpt-oss and/or text completion endpoint. 
```
def parse_chatml(prompt: str) -> List[dict]:
    pattern = re.compile(r"<\|start\|>(.*?)<\|message\|>(.*?)<\|end\|>", re.DOTALL)

    messages = []
    for role, content in pattern.findall(prompt):
        messages.append({"role": role.strip(), "content": content.strip()})
    return messages
```

I am a little confused by available prompt_format. 
If "openai", we don't need this at all. `parse_chatml` converts gpt oss string to conversation.

nemo_skills/inference/eval/critpt.py
jiacheng-xu: I ran into cases where parse_reasoning is set to be true but there is no end reasoning string in generation ( eg length reached max or just model behavior). If running parse_reasoning, the generation will be empty. 
Can we modify the "parse_reasoning" logic? I just want to avoid empty generation if the parsing fails.

nemo_skills/inference/eval/critpt.py
jiacheng-xu: do you want something like:
```
prompt = get_prompt(prompt_config=self.cfg.prompt_config_turn2)
turn2_prompt = prompt.build_user_message(input_dict=data_point)
```
I remember I ran into an issue where code_tags polluting the user prompt (adding "I am ChatGPT ... Tool ..." in the user prompt) but now I reviewed the implementation of get_prompt, this shouldn't be an issue. 
Just want to confirm if the above code is what you are looking for.

nemo_skills/inference/eval/critpt.py
jiacheng-xu: I agree with you. `_process_single_completion` is an almost duplicate function. The issue I had was: we have "system", "user", "assistant", "user" for the second turn. I wish I could just make it a prompt, but for one YAML file it can not take two "user" keys. Any thoughts on this?

One solution I have in mind: implementing "fill_prompt" and set an if-condition for turn 1 and turn 2. Does it work?

nemo_skills/inference/eval/critpt.py
Kipok: but that's expected behavior - if model hasn't produced end of thinking tag, we want to return empty generation

nemo_skills/inference/eval/critpt.py
Kipok: yes, but even better to do the refactoring I mentioned here and then this isn't a concern anymore https://github.com/NVIDIA-NeMo/Skills/pull/1200#discussion_r2776581375

nemo_skills/inference/eval/critpt.py
Kipok: if you set prompt_format=openai, then fill prompt isn't going to do anything and you can have any logic you need before calling super method for process single completion. Let me know if this isn't clear

nemo_skills/inference/eval/critpt.py
Kipok: so that's only needed for gpt-oss with tools. Maybe we don't add support for that into main and if you need to eval to get a baseline you can just do it with a one-time custom change on your side? I just don't want to add something like this into the core evaluation code that's super specific to a particular model and a particular inference strategy. All other models / use-cases shouldn't need this

nemo_skills/inference/eval/critpt.py
Kipok: Let's keep swebench for now as that, I will take care of that as well in another pr

nemo_skills/inference/eval/critpt.py
jiacheng-xu: Thanks! Please check the new `process_single_datapoint` after refactoring. 
`CritPtInferenceConfig` is still there since I am testing and running the code to make sure the number remains the same before refactoring. Other smaller changes can be resolved once we are good with this larger refactoring.

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,171 @@
...
+    def fill_prompt(self, data_point, data):
+        """Build messages list for turn 1, or return pre-built messages for turn 2."""
+        # If messages are already provided (turn 2), just return them
+        if "messages" in data_point:
+            return data_point["messages"]
+
+        # Turn 1: Build messages from prompt config
+        messages = []
+
+        # Add system message if present
```
Kipok: can this be replaced with a call to self.prompt_config_turn1.fill ?

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,171 @@
...
+        LOG.debug(f"Solution: {solution_turn1}")
+
+        # Get the turn1 messages by calling fill_prompt (it will rebuild them)
+        turn1_messages = self.fill_prompt(data_point, all_data)
+
+        # ===== Turn 2: Generate code using template =====
+        # Build messages for turn 2: turn1_messages + assistant response + turn2_user_message
+        turn2_messages = turn1_messages + [
+            {"role": "assistant", "content": solution_turn1},
+            {"role": "user", "content": self.turn2_prompt_template.format(**data_point)},
```
Kipok: can we use prompt.fill logic here directly as well? So you'd load both prompts as normal configs but call their .fill method manually. Not super important I guess, but a bit nicer to go through a single api point unless this causes some issues. If it makes code much more complicated or has other problems, then you can use your current way

nemo_skills/inference/eval/critpt.py
Kipok: yes, I think the current structure is pretty much what i had in mind. Left a few small comments and there are a few more not finished from previous review

nemo_skills/inference/eval/critpt.py
jiacheng-xu: I will keep CritPtInferenceConfig as is so the code can run. Removing `CritPtInferenceConfig` should be easy.

nemo_skills/inference/eval/critpt.py
jiacheng-xu: in `fill`, I don't see anyway to first add "assistant" and then add "user".
This is what I do for now:
```
        turn2_prompt_instance = get_prompt(prompt_config=self.cfg.prompt_config_turn2)
        turn2_user_messages:List[dict] = turn2_prompt_instance.fill(input_dict=data_point)

        turn2_messages = turn1_messages + [{"role": "assistant", "content": solution_turn1}] + turn2_user_messages
```

nemo_skills/inference/eval/critpt.py
jiacheng-xu: please take a look at the current version of code.

Kipok:
we do want to have some reference command / numbers in the documentation though

nemo_skills/evaluation/metrics/critpt_metrics.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def _get_score_dict(self, prediction: dict) -> dict[str, bool | int | float]:
+        """Extract score from CritPt evaluation results.
+
+        The evaluator sets is_correct based on aggregate accuracy distribution.
+        Since CritPt API doesn't provide per-example correctness, the first n examples
+        are marked as correct where n = int(aggregate_accuracy * total_examples).
+        """
+        return {"full_dataset_accuracy": prediction["full_dataset_accuracy"]}
```
Kipok:
```suggestion
        return {"accuracy": prediction["full_dataset_accuracy"]}
```
maybe at metrics level we call it just "accuracy" for clarity and consistency with other datasets?

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,166 @@
...
+
+        solution_turn1 = turn1_result[self.cfg.generation_key]
+        LOG.debug(f"Solution: {solution_turn1}")
+
+        # Get the turn1 messages by calling fill_prompt (it will rebuild them)
+        turn1_messages = self.fill_prompt(data_point, all_data)
+
+        # ===== Turn 2: Generate code using template =====
+        # Build messages for turn 2: turn1_messages + assistant response + turn2_user_message
+        turn2_prompt_instance = get_prompt(prompt_config=self.cfg.prompt_config_turn2)
```
Kipok: it's best to call get_prompt outside of this function. Otherwise it's being re-done for every single example

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,166 @@
...
+
+        turn2_messages = turn1_messages + [{"role": "assistant", "content": solution_turn1}] + turn2_user_messages
+        LOG.debug(f"Turn 2 messages: {turn2_messages}")
+
+        # Use a data point with turn 2 messages
+        # fill_prompt will detect the "messages" key and return them directly
+        turn2_data_point = {"messages": turn2_messages}
+        turn2_result = await super().process_single_datapoint(turn2_data_point, all_data)
+
+        if self.cfg.end_reasoning_string in turn2_result[self.cfg.generation_key]:
```
Kipok: `if self.cfg.parse_reasoning:` should be used here as well?

nemo_skills/prompt/config/eval/critpt/code_output.yaml
```diff
@@ -0,0 +1,8 @@
+
```
Kipok: reference link?

nemo_skills/prompt/config/eval/critpt/solve_problem.yaml
```diff
@@ -0,0 +1,30 @@
...
+  **Output (Markdown format)**
+  1. **Step-by-Step Derivation** -
+  Show every non-trivial step in the solution.  Justify steps using relevant physical laws, theorems, or mathematical identities.
+
+  2. **Mathematical Typesetting** - Use LaTeX for all mathematics:  `$...$` for inline expressions, `$$...$$` for display equations.
+
+  3. **Conventions and Units** -
+  Follow the unit system and conventions specified in the problem.
+
+  4. **Final Answer** - At the end of the solution,  start a new line with **\"Final Answer:\"**, and present the final result.
```
Kipok: do we need a `\` here? It will be rendered as is, not used to escape quotes

nemo_skills/inference/eval/critpt.py
jiacheng-xu: When adding parse_reasoning for turn2_result, all the generation becomes empty. Even though the original `turn2_result` contains </think> and the correct answer.
Can you confirm if `postprocess_single_output` in `GenerationTask` handles parse_reasoning? If true, we don't need parse_reasoning for turn2. 

response from cc:
The parse_reasoning call for turn2 was redundant because postprocess_single_output in the parent GenerationTask (line 647 of generate.py) already calls parse_reasoning on the final output when self.cfg.parse_reasoning is True. The turn2 call was stripping </think> first, and then the parent's call couldn't find it anymore, resulting in the generation being set to empty.

nemo_skills/evaluation/metrics/critpt_metrics.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def _get_score_dict(self, prediction: dict) -> dict[str, bool | int | float]:
+        """Extract score from CritPt evaluation results.
+
+        The evaluator sets is_correct based on aggregate accuracy distribution.
+        Since CritPt API doesn't provide per-example correctness, the first n examples
+        are marked as correct where n = int(aggregate_accuracy * total_examples).
+        """
+        return {"accuracy": prediction["accuracy"]}
```
Kipok: sorry, maybe i wasn't clear last time - can we save the aggregate accuracy as "full_dataset_accuracy" or something like that in each row. But then here, we re-map it to just "accuracy" as this field is a normal aggregate accuracy, not per-problem anymore. So something like this with an appropriate change to the evaluator class


```suggestion
        return {"accuracy": prediction["full_dataset_accuracy"]}
```

nemo_skills/evaluation/metrics/critpt_metrics.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def update(self, predictions):
+        """Update metrics with new predictions.
+
+        Args:
+            predictions: List of prediction dictionaries with is_correct field.
+        """
+        super().update(predictions)
```
Kipok, line 48: does this work? Don't we need to call `self._compute_pass_at_k(predictions)` or something like this?

nemo_skills/inference/eval/critpt.py
Kipok: this makes sense, so we should remove this if condition for turn 2 completely then?

nemo_skills/evaluation/metrics/critpt_metrics.py
jiacheng-xu: In evaluator, 
```

        for data_point in data_points:
            # Add evaluation results to data point
            data_point["evaluation_metadata"] = {
                "aggregate_accuracy": accuracy,
                "timeout_rate": timeout_rate,
                "server_timeout_count": server_timeout_count,
                "judge_error_count": judge_error_count,
            }
            data_point["full_dataset_accuracy"] = accuracy
```

In metrics, 
```
return {"accuracy": prediction["full_dataset_accuracy"]}
```

nemo_skills/evaluation/metrics/critpt_metrics.py
jiacheng-xu: Yes. I fixed this in my local.

nemo_skills/inference/eval/critpt.py
```diff
@@ -0,0 +1,166 @@
...
+        # Build messages for turn 2: turn1_messages + assistant response + turn2_user_message
+        turn2_user_messages: List[dict] = self.turn2_prompt_instance.fill(input_dict=data_point)
+
+        turn2_messages = turn1_messages + [{"role": "assistant", "content": solution_turn1}] + turn2_user_messages
+        LOG.debug(f"Turn 2 messages: {turn2_messages}")
+
+        # Use a data point with turn 2 messages
+        # fill_prompt will detect the "messages" key and return them directly
+        turn2_data_point = {"messages": turn2_messages}
+        turn2_result = await super().process_single_datapoint(turn2_data_point, all_data)
```
jiacheng-xu, line 131: @Kipok  there is no if condition for the second turn now.

PR #1204, opened 2026-01-31
Fix mkdocs warnings

Kipok:
## Summary by CodeRabbit

* **Documentation**
  * Corrected a link anchor in the evaluation documentation
  * Added Vision Language Models evaluation section to the documentation navigation


Kipok:
@activatedgeek I thought we had some new ci check to catch issues like these? Or do we not have it?

Kipok:
after looking more I don't think we changed ci, we just added `strict: true` parameter, right? If so, then `mkdocs serve` still worked fine with it, didn't fail with error. But ideally what we should do is we should add a new ci step that will just build the docs, not deploy them, so that we can run it in ci and hopefully configure to fail on any warnings, so that people can fix them before merging

PR #1206, opened 2026-02-02
Fix infinite wait in sandbox.wait_for_sandbox

i-vainn:
Fixed a bug where `wait_for_sandbox()` could hang indefinitely if the sandbox never became ready

## Release Notes

* **Bug Fixes**
  * Improved sandbox initialization reliability with enhanced timeout handling and better error messaging when startup delays occur.

* **Refactor**
  * Refined sandbox wait mechanism for more robust polling with clearer timeout boundaries.

### DIFF
--- a/nemo_skills/evaluation/evaluator/code.py
+++ b/nemo_skills/evaluation/evaluator/code.py
@@ -53,7 +53,7 @@ class CodeExecEvaluator(BaseEvaluator):
             f"max_output_characters={self.eval_config.max_output_characters}"
         )
         self.sandbox = get_sandbox(self.eval_config.sandbox)
-        self.sandbox.wait_for_sandbox(wait_timeout=50)
+        self.sandbox.wait_for_sandbox(wait_timeout=240)
 
     async def eval_single(self, data: dict):
         """Evaluate single code during generation."""


nemo_skills/code_execution/sandbox.py
```diff
@@ -271,8 +271,11 @@ def _check_ready(self, timeout: float = 5.0) -> bool:
         except httpx.HTTPError:
             return False
 
-    def wait_for_sandbox(self, timeout: int = 5):
-        while not self._check_ready(timeout=timeout):
+    def wait_for_sandbox(self, wait_timeout: int = 240, http_timeout: int = 5):
+        start_time = time.time()
+        while not self._check_ready(timeout=http_timeout):
+            if time.time() - start_time >= wait_timeout:
+                raise RuntimeError(f"Sandbox at {self.host}:{self.port} did not start within {wait_timeout} seconds")
```
Kipok, line 278: should we do the same for llm wait as well? @i-vainn

gwarmstrong: @Kipok are you saying like waiting for the llm server?

Kipok: yes, I think we should have the same timeout logic for wait for llm server as we do for sandbox server. Currently llm will also block forever if something is wrong with the address

gwarmstrong: how long do you think we should poll for? some models can take over 30 minutes to be ready--we could make the wait time configurable, but sometimes there is a decent amount of variance in the load times. Would we rather preemptively kill jobs that are ok, or fail misconfigured jobs with a job reaper (which will have a delay, but otherwise be more accurate signal).

Kipok: we can make it configurable. I guess by default we should set a fairly high timeout, like 2 hours. But just thinking that we should have a consistency in logic, if we support timeout for sandbox, we should also support it for llm. But this isn't very high priority

PR #1209, opened 2026-02-03
New slurm customization parameters (account, containers)

Kipok:
## Summary by CodeRabbit

* **New Features**
  * Added a global --account option to specify a Slurm account for job submissions.
  * Added container override options (--main-container, --sandbox-container, --judge-container, --judge-server-container, --container) to select non-default images; these overrides propagate across all job/task creation flows.

* **Tests**
  * Updated generation tests to accept the new account parameter.

### DIFF
--- a/tests/test_generation.py
+++ b/tests/test_generation.py
@@ -29,13 +29,14 @@ def test_eval_gsm8k_api(tmp_path):
     cmd = (
         f"ns eval "
         f"    --server_type=openai "
-        f"    --model=nvidia/nvidia/Nemotron-3-Nano-30B-A3B "
+        f"    --model=nvidia/nvidia/nemotron-nano-30b-v3 "
         f"    --server_address=https://inference-api.nvidia.com/v1/ "
         f"    --benchmarks=gsm8k "
         f"    --output_dir={tmp_path} "
         f"    ++max_samples=2 "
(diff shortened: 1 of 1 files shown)


nemo_skills/pipeline/eval.py
```diff
@@ -352,6 +364,12 @@ def eval(
     server_container: str = typer.Option(
         None, help="Override container image for the hosted server (if server_gpus is set)"
     ),
+    main_container: str = typer.Option(None, help="Override container image for the main evaluation client"),
+    sandbox_container: str = typer.Option(None, help="Override container image for the sandbox"),
+    judge_container: str = typer.Option(None, help="Override container image for GPU-based judges (comet, nvembed)"),
```
gwarmstrong, line 195: I think it's a little bulky to have separate override arguments for each container everywhere. Not sure that there is a better solution though. If we wanted to have overrides like we do for tools, e.g.,
```
++container_overrides.sandbox = "..."
++container_overrides.judge = "..."
```

But then the choice of key is unclear--since our "job components", e.g., Judge, main, sandbox, ... don't map cleanly to a container name (e.g., "judge" -> containers[judge_server_type], main -> containers["nemo-skills"], sandbox -> containers["sandbox"]).

So I think with the current structure, what you've done the best choice, but maybe we can eventually work toward something a little more general here.

PR #1210, opened 2026-02-04
Update promt_config to working with openai format + inline setup

Jorjeous:
Example
ns eval ... ++prompt_config=/path/to/prompt.yaml
ns eval ... ++user_message="Transcribe this audio"
ns eval ... ++prompt_config=/path/to/prompt.yaml

++prompt_config=... — Path or name of a prompt config YAML (e.g. eval/speechlm/numb3rs_text). Used for both NS and openai format; with openai, the template is filled and audio from the data is merged into the generated messages.
++user_message=... — Overrides only the user message. For openai-format data with a single user message, this is a quick way to try a different instruction without adding a YAML (e.g. for audio benchmarks).
++system_message=... — Overrides the system message

## Summary by CodeRabbit

* **New Features**
  * prompt_config now accepts templates, inline configs, or paths for flexible prompt construction.
  * Added user_message override and audio_field support; prompts can reference/attach audio metadata and now merge audio from input.
  * Improved system_message handling and optional prompt_suffix behavior across prompt modes.

* **Tests**
  * Expanded coverage for prompt_config paths, user/system/user_message overrides, prompt_suffix, and audio merging across formats.

### DIFF
--- a/docs/tutorials/posts/noc-reasoning-agent.md
+++ b/docs/tutorials/posts/noc-reasoning-agent.md
@@ -1,5 +1,5 @@
 ---
-date: 2026-02-13
+date: 2026-02-27
 readtime: 30
 hide:
   - toc
@@ -84,6 +84,8 @@ Key orchestration features used throughout this tutorial:
 | **<a href="https://slurm.schedmd.com/" target="_blank">Slurm</a> with <a href="https://github.com/NVIDIA/pyxis" target="_blank">NVIDIA/pyxis</a>** (optional) | Cluster job scheduler for distributed workloads |
 | **Python 3.10+** | Required Python version |
 | **NeMo-Skills CLI** | Main interface for running pipelines |
+
+
 ## Setup
 
 ### Install Nemo-Skills


nemo_skills/inference/generate.py
```diff
@@ -576,6 +589,54 @@ def skip_completed_samples(self, data):
...
+                        prompt[0]["audio"] = data_point["audio"]
+                    if "audios" in data_point:
+                        prompt[0]["audios"] = data_point["audios"]
+                else:
+                    prompt = self.cfg.new_prompt
+            
+            # Apply prompt_suffix if configured
+            if self.cfg.prompt_suffix:
+                if isinstance(prompt, list):
+                    prompt[-1]["content"] += self.cfg.prompt_suffix
```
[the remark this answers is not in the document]
Jorjeous: It must have content.

nemo_skills/inference/generate.py
```diff
@@ -576,6 +589,54 @@ def skip_completed_samples(self, data):
...
+                else:
+                    # No existing messages, use new_prompt and add audio from data_point
+                    prompt = deepcopy(self.cfg.new_prompt)
+                    for msg in prompt:
+                        if msg.get("role") == "user":
+                            if "audio" in data_point:
+                                msg["audio"] = data_point["audio"]
+                            if "audios" in data_point:
+                                msg["audios"] = data_point["audios"]
+                            break
```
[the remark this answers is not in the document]
Jorjeous: fixed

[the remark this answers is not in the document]
Jorjeous: fixed

nemo_skills/inference/generate.py
```diff
@@ -576,6 +589,54 @@ def skip_completed_samples(self, data):
...
+                        prompt[0]["audio"] = data_point["audio"]
+                    if "audios" in data_point:
+                        prompt[0]["audios"] = data_point["audios"]
+                else:
+                    prompt = self.cfg.new_prompt
+
+            # Apply prompt_suffix if configured
+            if self.cfg.prompt_suffix:
+                if isinstance(prompt, list):
+                    prompt[-1]["content"] += self.cfg.prompt_suffix
```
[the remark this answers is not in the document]
Jorjeous: ok. error is fine. why should someone add emtpy?

nemo_skills/inference/generate.py
```diff
@@ -576,6 +589,54 @@ def skip_completed_samples(self, data):
...
+                # String format - build simple message with audio if present
+                if "audio" in data_point or "audios" in data_point:
+                    prompt = [{"role": "user", "content": self.cfg.new_prompt}]
+                    if "audio" in data_point:
+                        prompt[0]["audio"] = data_point["audio"]
+                    if "audios" in data_point:
+                        prompt[0]["audios"] = data_point["audios"]
+                else:
+                    prompt = self.cfg.new_prompt
+
```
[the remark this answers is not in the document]
Jorjeous: error

nemo_skills/inference/generate.py
```diff
@@ -576,8 +589,64 @@ def skip_completed_samples(self, data):
...
+                else:
+                    # No existing messages, use new_prompt and add audio from data_point
+                    prompt = deepcopy(self.cfg.new_prompt)
+                    for msg in prompt:
+                        if msg["role"] == "user":
+                            if "audio" in data_point:
+                                msg["audio"] = data_point["audio"]
+                            if "audios" in data_point:
+                                msg["audios"] = data_point["audios"]
+                            break
```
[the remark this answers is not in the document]
Jorjeous: new version

Kipok (commented):
I'm confused about the motivation. Why can't we use an existing prompt_config for this?

nemo_skills/inference/generate.py
```diff
@@ -573,11 +575,44 @@ def skip_completed_samples(self, data):
 
         return remaining_data
 
+    def _merge_audio_from_data(self, messages, data_point):
```
Kipok: can this be handled at the prompt level, similar to how images are handled? Also why do we need two cases here for where audio is specified?

nemo_skills/inference/generate.py
```diff
@@ -573,11 +575,44 @@ def skip_completed_samples(self, data):
...
     # TODO: data will not include any samples skipped after restart
     def fill_prompt(self, data_point, data):
         """Passing in full data in case it's needed to fill the prompt in subclasses."""
-        if self.cfg.prompt_format == "openai":
-            if self.cfg.prompt_suffix:
+        if self.cfg.prompt_format == "openai" and self.prompt is None:
+            # Pure openai path -- messages come from the data
+            if self.cfg.user_message:
+                for msg in data_point["messages"]:
+                    if msg["role"] == "user":
```
Kipok: Can we maybe add an assert that there is only a single user message? I don't think we ever want to change all user messages in the conversation to the same string

nemo_skills/inference/generate.py
```diff
@@ -586,6 +621,25 @@ def fill_prompt(self, data_point, data):
                     data_point["messages"][0]["content"] = self.cfg.system_message
             return data_point["messages"]
 
+        if self.cfg.prompt_format == "openai" and self.prompt is not None:
+            # openai path with prompt_config template -- build prompt from template, merge audio from data
+            data_point = deepcopy(data_point)
+            filled_prompt = self.prompt.fill(
```
Kipok, line 670: could you please share an example of when we need this kind of functionality?

nemo_skills/prompt/utils.py
```diff
@@ -292,7 +296,15 @@ def fill(
         else:
             user_content = user_text
 
-        messages.append({"role": "user", "content": user_content})
+        user_message_dict = {"role": "user", "content": user_content}
+
+        # For audio: attach audio metadata to user message (model layer handles base64 conversion)
+        if self.config.audio_field and self.config.audio_field in input_dict:
+            user_message_dict["audio"] = input_dict[self.config.audio_field]
```
Kipok: instead of always having these two cases (audio vs auidos), should we just convert "audio" to a single element list at some initial point and then everywhere else we assume it's always audios which is a list (but can have only one element)?

nemo_skills/inference/generate.py
Jorjeous: When we testing diff promt variants to find the best one. f.e.

nemo_skills/prompt/utils.py
Jorjeous: Ok, will do

nemo_skills/inference/generate.py
Jorjeous: I guess no, this works with how our audio benchmarks is implemented, no audio data at top level

nemo_skills/inference/generate.py
Jorjeous: yes

nemo_skills/inference/generate.py
```diff
@@ -579,11 +581,37 @@ def skip_completed_samples(self, data):
...
+            if self.cfg.user_message:
+                user_msgs = [m for m in data_point["messages"] if m["role"] == "user"]
+                assert len(user_msgs) == 1, (
+                    f"user_message override expects exactly 1 user message, found {len(user_msgs)}"
+                )
+                user_msgs[0]["content"] = self.cfg.user_message
+            if self.cfg.prompt_suffix and data_point["messages"]:
                 data_point["messages"][-1]["content"] += self.cfg.prompt_suffix
             if self.cfg.system_message:
                 if data_point["messages"][0]["role"] != "system":
```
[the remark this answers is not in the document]
Jorjeous: ok

nemo_skills/inference/generate.py
```diff
@@ -579,11 +581,35 @@ def skip_completed_samples(self, data):
 
         return remaining_data
 
+    def _merge_audio_from_data(self, messages, data_point):
```
Kipok: can we do this automatically inside prompt/utils? If this is called after prompt.fill anyway, maybe we can do it inside, so that anywhere prompt.fill is used, this functionality is available?

nemo_skills/inference/generate.py
```diff
@@ -579,11 +581,35 @@ def skip_completed_samples(self, data):
...
+        """Copy audio metadata from original data messages into template-generated messages.
+
+        Normalizes to a single "audios" list on each message. Openai-format data always
+        has messages; top-level audio is not used in practice.
+        """
+        if "messages" not in data_point or not isinstance(data_point["messages"], (list, ListConfig)):
+            return
+        original_messages = data_point["messages"]
+        for i, msg in enumerate(messages):
+            if i < len(original_messages):
```
Kipok: shouldn't we just assert that len(messages) == len(original_messages)? Maybe we can also rename "messages" to "template_filled_messages" or something like this for clarity?

nemo_skills/inference/generate.py
```diff
@@ -579,11 +581,35 @@ def skip_completed_samples(self, data):
...
+        if self.cfg.prompt_format == "openai" and self.prompt is None:
+            # Pure openai path -- messages come from the data
+            data_point = deepcopy(data_point)
+            if self.cfg.user_message:
+                user_msgs = [m for m in data_point["messages"] if m["role"] == "user"]
+                assert len(user_msgs) == 1, (
+                    f"user_message override expects exactly 1 user message, found {len(user_msgs)}"
+                )
+                user_msgs[0]["content"] = self.cfg.user_message
+            if self.cfg.prompt_suffix and data_point["messages"]:
```
```suggestion
            if self.cfg.prompt_suffix:
```
it must be there if prompt_format == 'openai'

nemo_skills/inference/generate.py
```diff
@@ -606,9 +651,9 @@ def fill_prompt(self, data_point, data):
             format_as_string=(self.cfg.inference.endpoint_type == EndpointType.text),
         )
         if self.cfg.prompt_suffix:
-            if isinstance(filled_prompt, list):
+            if isinstance(filled_prompt, list) and filled_prompt:
                 filled_prompt[-1]["content"] += self.cfg.prompt_suffix
-            else:
+            elif isinstance(filled_prompt, str):
```
Kipok, line 702: what's the other possible case? Why do we need elif?

nemo_skills/inference/generate.py
```diff
@@ -606,9 +651,9 @@ def fill_prompt(self, data_point, data):
             format_as_string=(self.cfg.inference.endpoint_type == EndpointType.text),
         )
         if self.cfg.prompt_suffix:
-            if isinstance(filled_prompt, list):
+            if isinstance(filled_prompt, list) and filled_prompt:
```
Kipok: can filled_prompt be empty? If not, let's remove extra check

nemo_skills/prompt/utils.py
```diff
@@ -106,6 +106,9 @@ class PromptConfig:
     image_field: str | None = None
     # Whether to put image before or after the text in multimodal content
     image_position: str = "before"  # "before" or "after"
+    # Audio support: field name from input_dict containing audio metadata (dict or list of dicts)
+    # When set, audio metadata is attached to the user message as "audios" list for model processing
+    audio_field: str | None = None
```
Kipok, line 111: should we also support audio_position for consistency with image? Can do in another PR I guess, but feels nice to be consistent with parameters across modalities

tests/test_prompts.py
```diff
@@ -982,3 +984,283 @@ def test_llm_as_judge_hle_original_prompt():
         )
         == expected_prompt
     )
+
+
+def test_prompt_config_with_openai_format():
+    """Test that prompt_config works with openai format -- template is applied and audio is merged."""
+    from nemo_skills.inference.generate import GenerationTask
```
Kipok: maybe move to top?

nemo_skills/inference/generate.py
```diff
@@ -91,7 +91,7 @@ class GenerationTaskConfig:
 
     input_file: str  # Path to the input file with data
     output_file: str  # Where to save the generations
-    prompt_config: str | None = None  # How to format the data into prompts
+    prompt_config: Any = None  # How to format the data into prompts (str path, dict, or None)
```
Kipok, line 94:
```suggestion
    prompt_config: str | dict | None = None  # How to format the data into prompts (str path, dict, or None)
```

nemo_skills/prompt/utils.py
Jorjeous: Well, we can, but I do not remember model where audio is not first

nemo_skills/inference/generate.py
Jorjeous: Ok

nemo_skills/inference/generate.py
Jorjeous: Right

nemo_skills/inference/generate.py
Jorjeous: Isn't this too stict?
i mean template can change the len ?

Renaming - ok will be done

nemo_skills/prompt/utils.py
Jorjeous: Yep, if wanted better to put this in a following pr

tests/test_prompts.py
Jorjeous: ok, will do

nemo_skills/inference/generate.py
Jorjeous: no other case, but that's "no silent approach"

nemo_skills/inference/generate.py
Jorjeous: Not passing cicd hydra, reverting

PR #1223, opened 2026-02-09
Allow het servers for nemo-rl jobs

smahdavi4:
Allow heterogenous servers for nemo-rl jobs

## Summary by CodeRabbit

* **New Features**
  * Added LLM-as-a-judge support to GRPO workflow with flexible server configuration.
  * New CLI options for judge server (model, address, type, GPUs, nodes, args) and multi-server control.
  * Support for local or remote judge hosting with automatic port handling.
  * Scheduler now launches and orders judge server tasks to reserve GPU resources before training.
  * Multi-server scheduling and improved sandbox/overlap behavior for GPU-aware runs.


nemo_skills/pipeline/utils/exp.py
```diff
@@ -433,6 +433,7 @@ def add_task(
     keep_mounts_for_sandbox=False,
```
Kipok, line 444: @gwarmstrong do we need to update the declarative code path to reflect these changes?

gwarmstrong: It will need to be updated when we want to use the feature on the declarative path, but at the moment I'm not sure there is value to adding it to the declarative path purely for parity sake.

Is there any way to ensure it is covered by some test case (gpu or slurm probably?) that way when we convert to declarative, we can make sure the functionality isn't dropped?

PR #1224, opened 2026-02-09
Add LibTrace recipe for generating domain-specific reasoning data

jubick1337:
## Summary by CodeRabbit

* **Documentation**
  * Added LibTrace recipe, README, prompts, and nav entry describing the end-to-end pipeline, examples, and usage for physics, chemistry, and biology.

* **New Features**
  * New CLI utilities to harvest docs, prepare inference data, run applicability/relevance filtering, generate/collect problems, and gather/analyze solutions.
  * Added a sandbox container image, environment spec, and startup script for reproducible scientific runtimes.

* **Bug Fixes**
  * Improved sandbox execution robustness by handling transport/timeouts and performing best-effort cleanup.


docs/recipes/libtrace.md
```diff
@@ -0,0 +1,27 @@
...
+
+LibTrace is a recipe for building domain-specific reasoning data from library
+APIs. It harvests docstrings, labels applicability+relevance, generates
+problems, solves them with a boxed-answer prompt, and gathers solutions.
+
+The same workflow applies to chemistry, physics, and biology—swap the domain
+inputs and output paths.
+
+For the full walkthrough, command examples, and configuration details see
+[recipes/libtrace/README.md](/recipes/libtrace/README.md).
```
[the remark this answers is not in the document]
Kipok: this is correct @jubick1337 please run `mkdocs serve` to double check docs are rendered properly and fix any warnings

nemo_skills/code_execution/sandbox.py
```diff
@@ -173,18 +165,25 @@ async def execute_code(
             output = await self._send_request(request, timeout)
         except httpx.TimeoutException:
             output = {"process_status": "timeout", "stdout": "", "stderr": "Client timed out\n"}
+        except httpx.TransportError as e:
```
Kipok: @gwarmstrong to review this change, not fully sure on the impact or whether this should be needed. E.g. is session really lost here?

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -574,13 +575,35 @@ def _create_executor(
         if span_group_nodes and hardware and hardware.num_nodes is not None:
             num_nodes = hardware.num_nodes
 
+        # Only the server component (span_group_nodes=True) should claim group GPUs.
+        # Other components (client, sandbox) run on the same node but don't need GPUs.
+        if span_group_nodes:
```
Kipok: was this causing problems? I think we might have legit cases of sandbox needing gpus, e.g. for cuda code execution. And main client might also need gpus in case of run_cmd jobs

recipes/libtrace/scripts/run_boxed_inference.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
Kipok: do we need scripts like this? Isn't this exactly the same as ns generate with just an appropriate prompt config set?

nemo_skills/code_execution/sandbox.py
jubick1337: I hit this in practice: the custom sandbox image uses PySCF, which can get SIGILL. When a uwsgi worker crashes, the sandbox restarts it, but the HTTP connection to the client is already gone with httpx.TransportError. Without this catch, it bubbles up and kills all 32 concurrent generations.
With this change, only the affected generation gets an error and the rest continue. The "session was lost" message tells the LLM to re-import, which is harmless if the session is actually still alive. @gwarmstrong  happy to hear if there's a better recovery path.

nemo_skills/pipeline/utils/declarative.py
jubick1337: Fair point, I was matching the legacy add_task behavior (sandbox gets 0, client gets 0 when server is present), but you're right it could break run_cmd jobs or CUDA sandbox. Reverted the GPU change. The real fix for my case was the srun flags.

nemo_skills/code_execution/sandbox.py
gwarmstrong: IMO, the right thing to do would be to:
1. include `httpx.TransportError` in the preceding `except` block
2. add a logging command like: `LOG.warning("Sandbox communication error for session %s: %s", request_session_id, e)` to the above, so itnernally we can distinguish between the errors

This would catch both sources. Then, on the subsequent request, if the session was lost, it will message back to the client accordingly. This is consistent with how things are done presently for timeouts: https://github.com/NVIDIA-NeMo/Skills/blob/8e838e188921d669d749fe812dbc03df0b409904/nemo_skills/code_execution/sandbox.py#L252-L261

With the changes proposed presently in this PR, if the session/worker is lost, the the client would actually get two separate "the session died" warnings. Once by the proposed addition and another by the one we already have that catches the session loss retroactively. In the case where session is not lost, the model would get one message about the session being lost.

PR #1228, opened 2026-02-10
switch PythonTool from stdio to persistent HTTP server

gwarmstrong:


Kipok:
@gwarmstrong should we close this?

gwarmstrong:
@Kipok yes--the guidance now is to just use `DirectPythonTool`

PR #1229, opened 2026-02-10
Add nemo-skills-core subpackage for lightweight installs

gwarmstrong:
Adds a lightweight `nemo-skills-core` subpackage (`core/` subdirectory)
with only inference, evaluation, and tool calling deps. Default
`pip install nemo-skills` is unchanged (installs everything).

## Changes

- **`core/pyproject.toml` + `core/requirements.txt`**: New subpackage installable via `pip install ./core` or git URL with `#subdirectory=core`. Single source of truth for core deps, referenced by both core and root `pyproject.toml`.
- **`nemo_skills/pipeline/__init__.py`**: Import guard using `importlib.metadata` -- importing pipeline modules with only core installed raises a clear `ImportError` instead of a cryptic `ModuleNotFoundError`.
- **`nemo_skills/_cli_stub.py`**: Stub `ns` CLI entry point for core-only installs that prints a helpful message.
- **`nemo_skills/evaluation/evaluator/__init__.py`**: Lazy evaluator registry using string paths instead of eager imports, so core-only installs don't fail on benchmark-specific deps (`faiss`, `func_timeout`, etc.).
- **`nemo_skills/dataset/utils.py` + `nemo_skills/pipeline/dataset.py`**: Moved cluster-dependent dataset logic into pipeline module to keep core free of `nemo_run` imports.
- **`requirements/pipeline.txt`**: New requirements file for pipeline-only deps (`nemo_run`, `typer`, etc.).
- **`.github/workflows/tests.yml`**: Install `uv` in CI for use with testing installation.
- **Docs**: Added installation guide, updated CONTRIBUTING.md with dependency placement guidance.

## Summary by CodeRabbit

* **New Features**
  * Lightweight core package with a minimal CLI message and local-first dataset loading (cluster workflows remain available via the pipeline).
  * Evaluators now load lazily on demand to reduce startup cost.

* **Documentation**
  * Expanded installation & dependency guide explaining full vs core installs, extras, and the Core/Pipeline boundary.

* **Chores**
  * Packaging split into core/pipeline/dev dependency groups; added pipeline dependencies and core requirements.
  * CI workflow adds a UV setup step.


nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -13,106 +13,116 @@
...
+    return _resolved_class_map[eval_type]
+
+
+# --- Public API (unchanged signatures) ---
+
+# Keep EVALUATOR_MAP and EVALUATOR_CLASS_MAP as lazy-resolving dicts for
+# any code that iterates them (e.g. listing available types).
+# Direct key access goes through the public functions below.
+EVALUATOR_MAP = _EVALUATOR_MAP_PATHS
+EVALUATOR_CLASS_MAP = _EVALUATOR_CLASS_MAP_PATHS
```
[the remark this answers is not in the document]
gwarmstrong: Yep, leftover from debugging. Removed.

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -122,15 +132,15 @@ def supports_single_eval(eval_type: str, config: Dict[str, Any]) -> bool:
 def evaluate(eval_type, eval_config):
     """Main evaluation function that handles both class-based and function-based evaluators."""
     # Check if it's a class-based evaluator first
-    if eval_type in EVALUATOR_CLASS_MAP:
+    if eval_type in _EVALUATOR_CLASS_MAP_PATHS:
         evaluator = get_evaluator_class(eval_type, eval_config)
         print(f"evaluator: {evaluator}")
```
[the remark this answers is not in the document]
gwarmstrong: Yep, leftover from debugging. Removed.

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -13,106 +13,116 @@
...
+    return eval_type in _EVALUATOR_CLASS_MAP_PATHS or eval_type in _EVALUATOR_MAP_PATHS
 
 
 def register_evaluator(eval_type: str, eval_fn: Callable[[Dict[str, Any]], None]):
     if is_evaluator_registered(eval_type):
         raise ValueError(f"Evaluator for {eval_type} already registered")
 
-    EVALUATOR_MAP[eval_type] = eval_fn
+    _EVALUATOR_MAP_PATHS[eval_type] = None
+    _resolved_evaluator_map[eval_type] = eval_fn
```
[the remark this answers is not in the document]
gwarmstrong: Good catch, switched to a `"<dynamically-registered>"` sentinel to be safe.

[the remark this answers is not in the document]
gwarmstrong: Good catch, switched to a `"<dynamically-registered>"` sentinel to be safe.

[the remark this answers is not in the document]
gwarmstrong: Good catch, switched to a `"<dynamically-registered>"` sentinel to be safe.

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -13,106 +13,116 @@
...
     """Get evaluator instance by type."""
-    if eval_type not in EVALUATOR_CLASS_MAP:
+    if eval_type not in _EVALUATOR_CLASS_MAP_PATHS:
         raise ValueError(
             f"Evaluator class not found for type: {eval_type}.\n"
-            f"Available types with class support: {list(EVALUATOR_CLASS_MAP.keys())}\n"
-            f"All supported types: {list(EVALUATOR_MAP.keys())}"
+            f"Available types with class support: {list(_EVALUATOR_CLASS_MAP_PATHS.keys())}\n"
+            f"All supported types: {list(_EVALUATOR_MAP_PATHS.keys())}"
         )
```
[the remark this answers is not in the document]
gwarmstrong: Fixed, now shows the union of both maps, same as the error in `evaluate()` already did.

nemo_skills/pipeline/dataset.py
```diff
@@ -0,0 +1,114 @@
...
+
+def _get_default_dataset_module(dataset, data_dir=None, cluster_config=None):
+    """Load dataset module with cluster support.
+
+    For local-only loading (no cluster_config), delegates to
+    nemo_skills.dataset.utils.get_dataset_module.
+    """
+    if cluster_config is None or cluster_config.get("executor") in (None, "none"):
+        # Delegate to core for local-only loading
+        return _get_local_dataset_module(dataset, data_dir)
```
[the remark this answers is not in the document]
gwarmstrong: Fixed.

nemo_skills/evaluation/evaluator/__init__.py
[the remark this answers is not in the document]
gwarmstrong: Actually, reverting this back to `None`. The `_resolved_evaluator_map` cache is internal and never cleared, so this scenario cannot happen in practice. Per our project guidelines: "Don't add error handling, fallbacks, or validation for scenarios that can't happen." If the cache were somehow corrupted, crashing is the correct signal.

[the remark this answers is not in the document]
gwarmstrong: Actually, reverting this back to `None`. The `_resolved_evaluator_map` cache is internal and never cleared, so this scenario cannot happen in practice. Per our project guidelines: "Don't add error handling, fallbacks, or validation for scenarios that can't happen." If the cache were somehow corrupted, crashing is the correct signal.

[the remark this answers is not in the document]
gwarmstrong: Actually, reverting this back to `None`. The `_resolved_evaluator_map` cache is internal and never cleared, so this scenario cannot happen in practice. Per our project guidelines: "Don't add error handling, fallbacks, or validation for scenarios that can't happen." If the cache were somehow corrupted, crashing is the correct signal.

nemo_skills/pipeline/dataset.py
```diff
@@ -0,0 +1,114 @@
...
+
+def _get_default_dataset_module(dataset, data_dir=None, cluster_config=None):
+    """Load dataset module with cluster support.
+
+    For local-only loading (no cluster_config), delegates to
+    nemo_skills.dataset.utils.get_dataset_module.
+    """
+    if cluster_config is None or cluster_config["executor"] in (None, "none"):
+        # Delegate to core for local-only loading
+        return _get_local_dataset_module(dataset, data_dir)
```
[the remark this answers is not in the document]
gwarmstrong: Good catch. Forwarded extra_datasets and extra_datasets_type through _get_default_dataset_module to _get_local_dataset_module so the core function handles the fallback correctly.

core/requirements.txt
```diff
@@ -0,0 +1,27 @@
+# Core dependencies for agent runtime (inference, evaluation, tool calling, math/code grading).
+# No cluster orchestration deps (nemo_run, typer, etc.)
+
+
+# --- code evaluation ---
```
Kipok: are you sure this covers all benchmarks? Generally, we should move to keeping this reqs really simple and move most benchmark-specific requirements to install at runtime, but for now probably we might need some more packages here? E.g. datasets is almost certainly needed and then other benchmark specific things, like sacrebleu, etc.

docs/basics/installation.md
```diff
@@ -0,0 +1,74 @@
...
+# Development (everything + dev tools)
+pip install -e ".[dev]"
+```
+
+## Core / Pipeline architecture boundary
+
+The codebase enforces a one-way dependency rule: pipeline modules can import
+from core, but core modules must not import from pipeline.
+
+This boundary is enforced by `tests/test_dependency_isolation.py` which creates
```
Kipok: did you forget to push this test file?

requirements/pipeline.txt
```diff
@@ -0,0 +1,8 @@
+# Pipeline/orchestration dependencies (CLI, cluster management, experiment tracking).
+# These are additional to core.txt.
+
+click < 8.2.0  # https://github.com/ai-dynamo/dynamo/issues/1039
+nemo-evaluator-launcher<0.1.47
+nemo_run @ git+https://github.com/NVIDIA-NeMo/Run
+typer >= 0.13
+wandb
```
Kipok: this is actually a core dependency, it's being used in summarize-results, which is required for core functionality. Currently summarize-results is kind of in a weird half-pipeline state, but we should fix it to cleanly separate it into pipeline and non-pipeline components via https://github.com/NVIDIA-NeMo/Skills/issues/779#issuecomment-3344395623

CONTRIBUTING.md
```diff
@@ -49,6 +49,49 @@ The following things are required when adding new benchmarks
...
+
+**When adding a new dependency**, put it in the right requirements file:
+
+| If the dependency is needed for... | Add it to |
+|---|---|
+| Inference, evaluation, math/code grading, MCP clients, prompt formatting | `core/requirements.txt` |
+| CLI commands, cluster orchestration, experiment tracking | `requirements/pipeline.txt` |
+| Everything else (dataset-specific deps, benchmark-specific packages) | `requirements/main.txt` only |
+
+Dependencies in `core/requirements.txt` should be things that a typical `GenerationTask` run with PythonTool would need. Dataset-specific or benchmark-specific packages (e.g., `faiss-cpu`, `sacrebleu`, `func-timeout`) go only in `requirements/main.txt`.
```
Kipok: this part I don't fully understand - I think benchmark-specific packages should go to core for now as otherwise the code will fail when those benchmarks are used e.g. in evaluator. Eventually we should migrate to jit install, but it's not done yet, so I'd put those into core

CONTRIBUTING.md
```diff
@@ -49,6 +49,49 @@ The following things are required when adding new benchmarks
...
+
+All core and pipeline deps must also appear in `requirements/main.txt` (the monolithic file used for default installs).
```
Kipok: can we not link multiple requirements listed in pyproject.toml? We duplicate?

CONTRIBUTING.md
```diff
@@ -49,6 +49,49 @@ The following things are required when adding new benchmarks
...
+
+**Examples of correct placement:**
+
+- `httpx` -> `core/requirements.txt` (used by model inference clients)
+- `sympy` -> `core/requirements.txt` (used by math graders)
```
Kipok: e.g. math graders are only used for math benchmarks, so it's unclear why they go to core, but bfcl goes to main

CONTRIBUTING.md
```diff
@@ -49,6 +49,49 @@ The following things are required when adding new benchmarks
...
+**Examples of mistakes to avoid:**
+
+- Adding `nemo_run` to `core/requirements.txt` -- it is a pipeline/orchestration dependency, core must not depend on it.
+- Adding `typer` to `core/requirements.txt` -- it is the CLI framework, only used by the pipeline layer.
+- Adding a new dependency only to a subpackage file but forgetting `requirements/main.txt` -- the default install would be missing it.
+
+**When writing new core code:**
+
+- If you need something from `nemo_skills.pipeline`, your code probably belongs in pipeline, not core. Move it.
+- If you have a function that works locally but *also* needs a cluster variant, put the local version in core and a cluster-aware wrapper in `nemo_skills/pipeline/` (see `pipeline/dataset.py` for the pattern).
```
Kipok: I actually think that if we have a case like this, it means we need to redesign something. Ideally separation should be clean, and we shouldn't need to duplicate functionality. E.g. the dataset module part is a bit messy and there is probably a way to do it better, such that there is a pipeline level that only manages pulling from cluster and then there is a local level that always assumes things are present locally and is being called inside pipeline directly

docs/basics/installation.md
gwarmstrong: good catch

CONTRIBUTING.md
gwarmstrong: we should be able to do that. It should be implemented that way now--I updated this at one point so it links against the file rather than duplicating. Will fix.

CONTRIBUTING.md
gwarmstrong: Tbh my main goal was to reduce friction for integration with Gym, and you have a few good comments about the ambiguity of this boundary. I'll revisit and come up with something that is a bit more internally consistent and still satisfies the separation we care about.

CONTRIBUTING.md
gwarmstrong: Fair. My original scope was pretty PythonTool specific, but I think we can come up with something that makes a little more sense in terms of aligning the core code with core dependencies.

core/requirements.txt
gwarmstrong: I revisited the separation. This should contain all the reqs not needed for cluster orchestration now.

docs/basics/installation.md
gwarmstrong: pushed

requirements/pipeline.txt
gwarmstrong: fixed

CONTRIBUTING.md
gwarmstrong: yeah it's now in core and there is a clearer description of what's in `pipeline` vs `core`

CONTRIBUTING.md
gwarmstrong: done

CONTRIBUTING.md
gwarmstrong: moved. I think bfcl was causing some issues for `gym` install, but I think there's probably a better way to handle this (like JIT/dataset-specific installs). But let's handle that as a follow up, this PR just gives the structure. We can solve any residual Gym integration issues issues later.

CONTRIBUTING.md
gwarmstrong: Makes sense, updated the docs here to reflect that and made the implementation more consistent with the guidance here/there.

core/requirements.txt
```diff
@@ -0,0 +1,39 @@
+# Core dependencies for inference, evaluation, tool calling, and all benchmark evaluators.
```
Kipok, line 1: can we keep this inside requirements/core.txt? Would be simpler for people to only look in a single folder for all reqs

requirements/pipeline.txt
```diff
@@ -0,0 +1,7 @@
+# Pipeline/orchestration dependencies (CLI, cluster management, experiment tracking).
+# These are additional to core.txt.
```
Kipok: comment should be updated or we should move core/requirement.txt into core.txt in this folder

tests/test_dependency_isolation.py
```diff
@@ -0,0 +1,79 @@
...
+nemo_run is blocked via a sys.modules override.
+"""
+
+import subprocess
+import sys
+
+import pytest
+
+# Core modules that must be importable without nemo_run / pipeline
+CORE_MODULES = [
```
Kipok: can we dynamically find everything inside nemo_skills except pipeline subfolder?

CONTRIBUTING.md
```diff
@@ -49,6 +49,53 @@ The following things are required when adding new benchmarks
   the dataset into slurm tests. This is the most comprehensive test we can do by running full
   evaluation on cluster with arbitrary model and check that results are as expected.
 
+### Respect the Core / Pipeline dependency boundary
```
Kipok, line 52: can we maybe keep the part in here brief, just summarize the logic in a few sentences / bullet points. And then the full description we move to another .md file? I think the full description is quite helpful, but it's a bit too detailed for the guidelines section, which I hope we can keep relatively short

.github/workflows/tests.yml
```diff
@@ -41,6 +41,8 @@ jobs:
       with:
         python-version: "3.10"
         cache: pip
+    - name: Install uv
```
Kipok: do we use it anywhere?

docs/basics/installation.md
```diff
@@ -0,0 +1,74 @@
...
+fresh virtualenvs and verifies that core modules import successfully without
```
Kipok: is this true? I don't think we create fresh envs in tests?

docs/basics/installation.md
```diff
@@ -0,0 +1,74 @@
...
+pip install -e .
+
+# Core only -- lightweight runtime for downstream integrations
+pip install -e core/
+
+# Development (everything + dev tools)
+pip install -e ".[dev]"
+```
+
+## Core / Pipeline architecture boundary
```
Kipok: I'd maybe move this part somewhere else (e.g. only keep in contributing.md or better in a new .md where extra details from contributing can go as well). It's helpful, but probably a bit too dense for the "basics" part of the docs. More oriented towards people who'd need to modify our code

core/requirements.txt
gwarmstrong: It has to be in the same directory as the `pyproject.toml` for the install of `nemo-skills-core` to work. But I can put a symlink in the `requirements/` directory so it is at least clear that this file exists if someone goes to look in `requirements.txt` for it?

tests/test_dependency_isolation.py
gwarmstrong: done

.github/workflows/tests.yml
gwarmstrong: good catch, it was from an old testing strategy, removed

docs/basics/installation.md
gwarmstrong: yeah this is outdated from old tests, good cattch

CONTRIBUTING.md
gwarmstrong: sure. made it brief and moved the bulk of the guidelines to core/README.md, which is referenced here now

docs/basics/installation.md
gwarmstrong: okay, put it in the `core/README.md`

PR #1234, opened 2026-02-11
feat: migrate sandbox from uwsgi to gunicorn

gwarmstrong:
## Summary
- Replace `tiangolo/uwsgi-nginx-flask` base image with `python:3.10-slim` + nginx + gunicorn
- Remove all uwsgi configuration (UWSGI_PROCESSES, UWSGI_CHEAPER, UWSGI_CPU_AFFINITY, uwsgi_read_timeout)
- Add deprecation warnings for old uwsgi env vars in `start-with-nginx.sh`
- Fix SIGINT handler in `shell_worker` for gunicorn compatibility
- Update `icpc`/`ioi` dataset env vars to use `GUNICORN_WORKERS`
- Test fixes for port conflicts and `sys.executable` usage

Depends on #1233 (lockfiles). Split from #1226.

## Summary by CodeRabbit

* **Chores**
  * Switched to a lighter base image and simplified worker-related environment configuration.
  * Replaced legacy worker orchestration with a unified worker model and updated sandbox environment defaults for scaling.
  * Improved startup reliability with port-availability checks and coordinated worker port assignment.
  * Improved graceful shutdown and lifecycle handling.

* **Bug Fixes**
  * Restored interrupt handling in the local sandbox for proper KeyboardInterrupt behavior and updated test invocation to use the system Python.


Kipok (approved):
lgtm as long as all tests work well. Few AI comments from offline review that seem meaningful to check

  1. Gunicorn with 1024 workers (Medium concern)

  In icpc/__init__.py and ioi/__init__.py:
  "GUNICORN_WORKERS=1024"
  uWSGI's cheaper mode allowed scaling from 1 active process up to 1024 on demand. Gunicorn doesn't have an equivalent — it will fork
  all 1024 workers at startup. This could cause significant memory overhead and slow startup. Consider whether this is the intended
  behavior, or if a smaller worker count should be used. If the 1024 was only needed because uWSGI's cheaper mode made it cheap to
  declare a high max, this number likely needs to be reduced.

  3. worker_had_port_conflict is now dead code (Minor)

  The function at line 164 of start-with-nginx.sh:

  worker_had_port_conflict() {
      grep -q "Address already in use" /var/log/worker${1}.log 2>/dev/null
  }


  This was used by the old Phase 2 retry loop which has been removed. The function is no longer called anywhere and should be deleted.


In shell_worker, resetting SIGINT to signal.SIG_DFL makes timeout interrupts terminate
 the shell process instead of raising KeyboardInterrupt. ShellManager.run_cell() sends SIGINT after timeout expecting the shell to handle interruption and return an
 interrupted response; with SIG_DFL, this path turns into EOFError + forced shell restart, dropping session state on timeouts. This regresses persistent-session behavior and 
should restore Python's default interrupt handler (signal.default_int_handler) rather than the OS default action.

Kipok:
@gwarmstrong what's the status of this one? Some problems with ioi / icpc evals for worker control? Do we want to push for fixes or not worth it given that our current sandbox is hopefully going away in the future?

gwarmstrong:
@Kipok yeah iirc there are some technical constraints with gunicorn that don't give us as much worker concurrency control as we had in uwsgi. It would require further exploration to see if it matters, or figure out how to alternately configure it. Rather than explore more here, I would prefer to not break existing workloads and sunset as much sandbox development as possible.

PR #1240, opened 2026-02-12
Adding an option to store benchmarks in external repo

Kipok:
The main motivation for this pr is to enable registering external benchmarks, so that people can use nemo-skills with custom private eval repositories without having to maintain a whole fork. This required some restructuring which resulted in the following changes
1. Added an option for benchmark to be specified as full path to the dataset folder. Also added a logic to register "benchmark_map.json" that will map short name to full path, which can simplify referencing external benchmarks.
2. Added tests and documentation for the new logic, should be quite extensive
3. Added an option to pass evaluators as file::class, similar to how was already done for metrics
4. Made summarize results command from eval directly specify metrics type, this way it works inside the container without having to reference init from external benchmarks which might be in a different path
5. Added REQUIRES_DATA_DIR option to init directly, was previously hardcoded which made it hard to maintain
6. Made BFCL init.py files explicitly committed, they weren't dynamic, so having them being in the repo makes things simpler and more explicit
7. Removed previous logic to register extra datasets, which only worked with "simple" benchmarks that didn't require custom evaluation / generation / metrics
8. Removed DATASET_GROUP from inits as we didn't really use it much, it makes little sense, was inconsistent and makes things hard with external benchmarks
9. Changed the base generate api to add prompt_format argument. This isn't strictly needed for this PR, but it made an example in the docs much simpler - previously it was hard to switch from our prompt format in first turn to openai format in the second turn, but with this argument, we can override the prompt format and reuse a lot more code.
10. Removed hardcoded logic for datasets with subfolders for packaging and made it dynamically pick all jsonl automatically
11. Made a pythonic interface for launch / stop server. Again, not strictly needed, but made the new tests I added much faster as we reuse the server in them (otherwise would have to re-spin it 12 times). This also will make test_eval.py much better in the future if we add it there since it will allow us to run judge benchmarks, but I didn't yet make this change yet

## Summary by CodeRabbit

* **New Features**
  * External/custom benchmark support: add, prepare, package, and evaluate benchmarks outside the repo; server launch/stop utilities for evaluations.
  * prompt_format passthrough to customize prompt formatting across generation and evaluation tasks.

* **Improvements**
  * Simplified dataset resolution and data-dir/container handling for prepare/package/eval flows.
  * More robust CLI behavior for prepare/eval/summarize and improved packaging of external datasets.

* **Documentation**
  * Added guide for defining and integrating external benchmarks.

* **Tests**
  * New end-to-end integration tests for external benchmark prepare + eval.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,6 +85,7 @@ jobs:
           .
     - name: Run all tests
       env:
+        NV_INFERENCE_API_KEY: ${{ secrets.NV_INFERENCE_API_KEY }}
         NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
       run: |
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ AGENTS.md
 .codex
 
 .idea
+
+#scripts at root level
+/*.sh
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -210,7 +210,7 @@ class GenerationTaskConfig:
     enable_litellm_cache: bool = False
 
     # List of content types to drop from messages (e.g., base64 audio) to keep output files smaller
-    drop_content_types: list[str] = field(default_factory=lambda: ["audio_url"])
+    drop_content_types: list[str] = field(default_factory=lambda: ["audio_url", "input_audio"])
 
     # Audio configuration - set by benchmarks that need audio processing (mmau-pro, audiobench, etc.)
     enable_audio: bool = False  # Enable audio preprocessing (set by benchmark configs)
@@ -633,7 +633,7 @@ class GenerationTask:
 
             # Filter out content types specified in drop_content_types config
             message["content"] = [
-                content for content in message["content"] if content.get("type") not in self.cfg.drop_content_types
+                content for content in message["content"] if content["type"] not in self.cfg.drop_content_types
             ]
 
     async def postprocess_single_output(self, output, original_data_point):
--- a/nemo_skills/inference/model/audio_utils.py
+++ b/nemo_skills/inference/model/audio_utils.py
@@ -145,6 +145,8 @@ def make_audio_content_block(base64_audio: str, audio_format: str = "audio_url")
     if audio_format == "input_audio":
         # OpenAI native format (works with NVIDIA API / Gemini / Azure)
         return {"type": "input_audio", "input_audio": {"data": base64_audio, "format": "wav"}}
-    else:
+    elif audio_format == "audio_url":
         # Data URI format (works with vLLM / Qwen)
         return {"type": "audio_url", "audio_url": {"url": f"data:audio/wav;base64,{base64_audio}"}}
+    else:
+        raise ValueError(f"Unsupported audio_format '{audio_format}'. Use 'audio_url' or 'input_audio'.")
--- a/nemo_skills/inference/model/vllm_multimodal.py
+++ b/nemo_skills/inference/model/vllm_multimodal.py
@@ -17,6 +17,7 @@
 This module provides a multimodal model class that handles:
 - Audio INPUT: encoding audio files to base64, chunking long audio
 - Audio OUTPUT: saving audio responses from the server to disk
+- External API support: NVIDIA Inference API, OpenAI, and other OpenAI-compatible APIs
 """
 
 import base64
@@ -32,6 +33,7 @@ from .audio_utils import (
     audio_file_to_base64,
     chunk_audio,
     load_audio_file,
+    make_audio_content_block,
     save_audio_chunk_to_base64,
 )
 from .vllm import VLLMModel
@@ -43,10 +45,10 @@ DEBUG_INFO_PATTERN = re.compile(r"\n?<debug_info>(.*?)</debug_info>", re.DOTALL)
 
 
 class VLLMMultimodalModel(VLLMModel):
-    """VLLMModel with support for audio input and output.
+    """VLLMModel with support for audio input/output and external APIs.
 
     Audio INPUT capabilities:
-    1. Converts audio file paths to base64-encoded audio_url format
+    1. Converts audio file paths to base64-encoded input_audio format
     2. Chunks long audio files for models with duration limits
     3. Aggregates results from chunked audio processing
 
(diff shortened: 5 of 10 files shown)


docs/evaluation/custom-benchmarks.md
```diff
@@ -0,0 +1,360 @@
...
+    register_external_repo,
+    RepoMetadata,
+)
+from pathlib import Path
+
+# Register repo so it gets packaged inside containers.
+# ignore_if_registered avoids errors when the module is imported more than once.
+register_external_repo(
+    RepoMetadata(name="my_benchmarks", path=Path(__file__).parents[2]),
+    ignore_if_registered=True,
```
gwarmstrong, line 76: when do you anticipate this happening? if multiple custom benchmarks are used? or if you are writing to the same name as an existing dataset?

docs/evaluation/custom-benchmarks.md
```diff
@@ -0,0 +1,360 @@
...
+  Count the number of words in the quoted sentence below.
+  Put your final answer (just the number) inside \boxed{{}}.
+
+  {sentence}
+```
+
+In `GENERATION_ARGS` this is referenced as:
+
+```
+++prompt_config=/nemo_run/code/my_benchmarks/prompt/eval/word_count/default.yaml
```
gwarmstrong: Is there any value to allowing the benchmark_dataset.jsonl to account for prompt configs too? or otherwise use the `::` syntax in some way

nemo_skills/dataset/utils.py
```diff
@@ -87,90 +80,131 @@ def _get_dataset_module_from_cluster(cluster_config, mounted_path):
...
 
-def get_dataset_module(dataset, data_dir=None, cluster_config=None, extra_datasets=None, extra_datasets_type=None):
+def get_extra_benchmark_map():
+    """Load extra benchmark map from NEMO_SKILLS_EXTRA_BENCHMARK_MAP env var if set.
+
+    Returns a dict mapping short names to directory paths, or empty dict.
+    Relative paths in the map are resolved relative to the map file's directory.
     """
-    Get dataset module either in default folder or in extra datasets folder.
+    map_path = os.environ.get("NEMO_SKILLS_EXTRA_BENCHMARK_MAP")
```
gwarmstrong: can we pass this as an argument so there is a little more configurability? e.g., this would enable passing benchmark_map as an arg at the python pipeline level, which I think would be more convenient when writing python than exporting as env var

nemo_skills/evaluation/evaluator/__init__.py
```diff
@@ -85,52 +89,90 @@
...
+    """
+    if "::" in eval_type:
+        module_str, attr_str = eval_type.split("::", 1)
+        if Path(module_str).is_file():
+            module = import_from_path(module_str)
+        else:
+            module = importlib.import_module(module_str)
+        obj = getattr(module, attr_str)
+        is_class = inspect.isclass(obj) and issubclass(obj, BaseEvaluator)
+        return obj, is_class
```
gwarmstrong: is it possible the consolidate the logic between this and `nemo_skills.mcp.utils.locate`?

nemo_skills/inference/eval/arena_judge.py
```diff
@@ -98,10 +98,11 @@ def setup_prompt(self):
...
         return default_prompt
 
-    def fill_prompt(self, data_point, data):
+    def fill_prompt(self, data_point, data, prompt_format=None):
         """Fill prompt with category-specific prompt config."""
-        if self.cfg.prompt_format == "openai":
-            return super().fill_prompt(data_point, data)
+        prompt_format = prompt_format or self.cfg.prompt_format
+        if prompt_format == "openai":
+            return super().fill_prompt(data_point, data, prompt_format)
```
gwarmstrong: can we use keyword arguments for this? I find it good practice for clarity when there's more than one or two arguments to the function

nemo_skills/inference/eval/bfcl.py
```diff
@@ -608,7 +608,7 @@ def _parse_reasoning_from_message_content(self, model_response_text: str | None)
             # If the thinking didn't finish, we can keep it empty
             return ""
 
-    async def process_single_datapoint(self, data_point, all_data):
+    async def process_single_datapoint(self, data_point, all_data, prompt_format=None):
```
gwarmstrong, line 611: while we're in here, can we just convert this to `**kwargs`? that way any update like this in the future, where an argument is added for the sake of one or two evals, doesn't spur a 30 file diff. Any method that uses it directly can specify it explicitly.

nemo_skills/pipeline/eval.py
```diff
@@ -407,17 +406,6 @@ def eval(
     ),
     config_dir: str = typer.Option(None, help="Can customize where we search for cluster configs"),
     log_dir: str = typer.Option(None, help="Can specify a custom location for slurm logs."),
-    extra_datasets: str = typer.Option(
-        None,
-        help="Path to a custom dataset folder that will be searched in addition to the main one. "
-        "Can also specify through NEMO_SKILLS_EXTRA_DATASETS.",
```
gwarmstrong, line 413: any speicif reason this was removed? I think at least being able to specify the benchmark_map.json here would be handy

nemo_skills/pipeline/start_server.py
```diff
@@ -108,6 +108,92 @@ def create_job_tunnel(
...
+        server_config["container"] = server_container
+
+    if sandbox_port is None:
+        sandbox_port = get_free_port(strategy="random") if get_random_port else 6000
+
+    # TODO (igitman): this looks like leak, but also fixing it in a naive way makes everything hang.
+    #     Doesn't seem to cause issues, so keeping like this for now
+    exp = get_exp("server", cluster_config).__enter__()
+
+    add_task(
```
gwarmstrong, line 172: since we're pretty heavily modifying this pipeline anyways, can you take a pass at converting it to the declarative pipeline?

nemo_skills/pipeline/eval.py
Kipok: that was a previous way to specify extra datasets that couldn't pick up the code, but only took jsonl files / read from init. I removed it since that logic is replaced with the new logic in this pr. Let me add a new argument for benchmark map

nemo_skills/pipeline/start_server.py
Kipok: maybe in another PR? This one is already really big

nemo_skills/inference/eval/bfcl.py
Kipok: not sure I understand - the main point is that this prompt format is used in the base class now. The subclasses can always add new arguments and override the method. But if we change the base class, I think it's better to be explicit about what arguments are being used?

nemo_skills/inference/eval/arena_judge.py
Kipok: sure

docs/evaluation/custom-benchmarks.md
Kipok: if multiple benchmarks are used. I think main use-case will be to have a single internal repo with 10s of internal benchmarks. Then each of them has to have this register call and if a few are specified together, it will fail if we don't ignore registered

docs/evaluation/custom-benchmarks.md
Kipok: not sure I fully understand, can you clarify this please? Prompt is just a yaml file, how would we use `::` syntax?

nemo_skills/dataset/utils.py
Kipok: will update

docs/evaluation/custom-benchmarks.md
gwarmstrong: Sorry, I  mean more so is being able to specify it relative to the custom benchmark, like `my_benchmarks/prompt/eval/word_count/default.yaml`. It seems minor, but the `/nemo_run/code` mount causes a lot of friction for people, and I think allowing the relative usage would make the interface a bit more ergonomic for a lot of users.

docs/evaluation/custom-benchmarks.md
gwarmstrong: what do you mean by "specified together"? Is this targeted at clashing across namespaces (e.g., two different benchmarks register a "my_dataset" dataset?)? I'm wary of ignores, because they can make it easy to do something different than you intend.

Kipok: I mean if internal benchmarks repo has "benchmark1" and "Benchmark2" in different folders under dataset. Both have to have this call in their init, but if you do `--benchmarks=benchmark1,benchmark2` it will fail without that ignore argument

docs/evaluation/custom-benchmarks.md
Kipok: got it, will update

nemo_skills/inference/eval/bfcl.py
gwarmstrong: I'm just a bit resistant to refactorings that require a cascading signature change. This signature is pretty clean as it is. Why does this arg need to be introduced to this signature, rather than included in the Generation Config, or the data point?

Kipok: we can't add it to generation config (it's already there), since here we want to do first call with prompt_format=x and second call with prompt_format=y. We could add it to data_point, but is it cleaner? Feels like something that doesn't belong in data_point dictionary as it's not related to any data configuration, rather how we use that data. The only other way to work around this is to do self.cfg.prompt_format=X before that call and then restore it back after, but it also looks worse than adding it to signature? Let me know if you have other suggestions

docs/evaluation/custom-benchmarks.md
gwarmstrong: In what case would you want to not use `ignore_if_registered` then?

Kipok: tbh, I'd remove that parameter and make it default behavior. But that would be a breaking change, so I didn't do it. I guess we can maybe add some check that if name / path pair is the same, then we don't error out, but if name is same, but path is different, then we do? As that's the only dangerous case here

nemo_skills/inference/eval/bfcl.py
gwarmstrong: Yeah agree both those options seem worse. This is probably fine.

nemo_skills/pipeline/start_server.py
gwarmstrong: that is ok

docs/evaluation/custom-benchmarks.md
```diff
@@ -0,0 +1,359 @@
...
+keep proprietary data private, iterate on benchmarks independently of NeMo-Skills
+releases, and share team-owned benchmarks without modifying the main repository.
+
+An external benchmark can customize every part of the evaluation pipeline:
+dataset preparation, prompt template, generation logic, evaluator, and metrics.
+
+## Quick start
+
+1. **Create a repo** with `benchmark_map.json`, a dataset `__init__.py`, and a `prepare.py`.
+2. **Set the env var** `NEMO_SKILLS_EXTRA_BENCHMARK_MAP` to point at your `benchmark_map.json` (`name -> path` structure).
```
gwarmstrong, line 13: maybe we can have this demo prefer the pipeline argument now? Without having a concrete use case that I've iterated on myself a bit, I'm 50/50 on which is generally easier to pass

docs/evaluation/custom-benchmarks.md
gwarmstrong: I'm still a little confused why the double trigger occurs, but I think what you've said in the last comment here seems like a good solution.

docs/evaluation/custom-benchmarks.md
Kipok: let me keep the current one, but we can update it later if that's not how most people use it

docs/evaluation/custom-benchmarks.md
Kipok: https://github.com/NVIDIA-NeMo/Skills/issues/1260

PR #1245, opened 2026-02-18
Fix no_answer metric overcounting in _compute_pass_at_k

sgunasekar:
Issue:
- The no_answer accumulation was overcounting as it inside the for-score_method loop, causing it to be incremented once per score method per sample instead of once per sample. 
- With N score methods (e.g. symbolic_correct, judge_correct, both_correct, any_correct), no_answer was overcounted by N×, producing impossible values > 100%.

Fix: 
- Move no_answer accumulation outside the score_method loop into its own for-k loop\

## Summary by CodeRabbit

* **Refactor**
  * Reordered internal evaluation flow for pass@k and no-answer handling so per-k calculations and updates occur in a single outer loop. This changes internal metric update sequencing while preserving external behavior and public interfaces. Improved predictability of metric computation ordering; no user-facing API changes.

### DIFF
--- a/nemo_skills/dataset/aime26/__init__.py
+++ b/nemo_skills/dataset/aime26/__init__.py
@@ -13,6 +13,5 @@
 # limitations under the License.
 
 # settings that define how evaluation should be done by default (all can be changed from cmdline)
-DATASET_GROUP = "math"
 METRICS_TYPE = "math"
 GENERATION_ARGS = "++prompt_config=generic/math ++eval_type=math"
--- a/nemo_skills/inference/eval/bfcl.py
+++ b/nemo_skills/inference/eval/bfcl.py
@@ -54,7 +54,7 @@ BFCL_REQUIREMENTS = [
     "tqdm",
     "numpy==1.26.4",
     "pandas",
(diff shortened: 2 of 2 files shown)


nemo_skills/evaluation/metrics/base.py
```diff
@@ -410,18 +410,19 @@ def _compute_pass_at_k(
                     predicted_answers=predicted_answers,
                 )
 
-                if predicted_answers is not None and self.compute_no_answer:
-                    no_answer_list = [pred_answer is None for pred_answer in predicted_answers[:k]]
-                    eval_dict[f"pass@{k}"]["no_answer"] += all(no_answer_list)
-                    eval_dict[f"pass@1[avg-of-{k}]"]["no_answer"] += sum(no_answer_list) / k
-
                 self._update_metrics_for_pass(
```
Kipok: not related to this pr, but I guess this needs to be moved out of the loop as well?

sgunasekar: is this referring to _update_metrics_for_pass?

sgunasekar: @Kipok  - added this fix - could you do another quick review before we merge this.

PR #1254, opened 2026-02-18
Add DSBench-DA evaluation

sgunasekar:
Adds evaluation support for the DA (data analysis) portion of DSBench.

- New dataset (`dsbench_da`) with data preparation and prompt configs
- New evaluator with case-insensitive MCQ comparison and recursive list/dict comparison
- Minor change to math evaluator to pass `relaxed` argument through to `extract_answer` — default behavior is unchanged
- Updated `sandbox.lock` with DSBench dependencies (`openpyxl`, `pyxlsb`)

Most files are independent and will not affect existing workflows. Evals have been tested end-to-end.

## Summary by CodeRabbit

* **New Features**
  * DSBench evaluation added with flexible/relaxed answer matching for varied formats.
  * Data preparation pipeline and CLI for DSBench: downloads, validates, converts Excel tasks to prompt-ready entries.
  * New prompt configurations for in-context and tool-mode DSBench usage, enforcing boxed final-answer formatting.

* **Chores**
  * Added Excel/data-processing dependencies and updated ignore entries for local tooling.

### DIFF
--- a/nemo_skills/dataset/dsbench_da/__init__.py
+++ b/nemo_skills/dataset/dsbench_da/__init__.py
@@ -17,7 +17,7 @@ EVAL_SPLIT = "test"
 METRICS_TYPE = "math"
 
 # Use DSBench evaluator (extends MathEvaluator) with relaxed extraction and case-insensitive MCQ and handling of dict and list.
-GENERATION_ARGS = "++prompt_config=generic/dsbench-da ++eval_type=dsbench ++eval_config.relaxed=true"
+GENERATION_ARGS = "++prompt_config=generic/dsbench-da ++eval_type=dsbench ++eval_config.relaxed_extraction=true"
 
 # Recommend running LLM judge to verify dicts and lists correctly
 # JUDGE_PIPELINE_ARGS = {


nemo_skills/evaluation/evaluator/dsbench.py
```diff
@@ -0,0 +1,82 @@
...
+    if is_mcq and verify(parsed_gt, parsed_pred):
+        return verify(parsed_gt, parsed_pred)
+    
+    return math_equal(str(gt_answer), str(predicted_answer))
+
+
+class DSBenchEvaluator(MathEvaluator):
+    def __init__(self, config: dict, num_parallel_requests=10):
+        super().__init__(config, num_parallel_requests)
+        self.eval_config.extract_regex = r"(?:The final answer is |\\boxed=)(.+)$"
```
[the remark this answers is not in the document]
sgunasekar: This is intentional for more relaxed parsing

nemo_skills/dataset/dsbench_da/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+EVAL_SPLIT = "test"
+DATASET_GROUP = "code"
+METRICS_TYPE = "math"
+
+# Use DSBench evaluator (extends MathEvaluator) with relaxed extraction and case-insensitive MCQ and handling of dict and list.
+GENERATION_ARGS = "++prompt_config=generic/dsbench-da ++eval_type=dsbench ++eval_config.relaxed=true"
+
+# # Recommend running LLM judge to verify dicts and lists correctly
```
Kipok: should this not be commented out?

nemo_skills/dataset/dsbench_da/prepare.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # Download and extract if not already cached
+    if not extracted_data_dir.exists() or not metadata_path.exists():
+        print("  Downloading dataset from HuggingFace...")
+
+        # Download data.zip
+        print("    Downloading data.zip...")
+        zip_url = "https://huggingface.co/datasets/liqiang888/DSBench/resolve/main/data_analysis/data.zip"
+        zip_path = data_dir / "data.zip"
+        urllib.request.urlretrieve(zip_url, zip_path)
```
Kipok: can we use an official hf download api? Otherwise this request will very frequently fail because of hf rate limiting

nemo_skills/evaluation/evaluator/dsbench.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def relaxed_equal(gt_answer: Any, predicted_answer: Any) -> bool:
+    """
+    Relaxed equality check with:
+    1. Case-insensitive MCQ matching
+    2. Dict/list comparison using math_equal recursively
+    """
+    if predicted_answer is None:
+        return gt_answer is None
```
Kipok, line 36: can gt_answer be None? If not, probably better to just return False here for clarity

nemo_skills/evaluation/evaluator/dsbench.py
```diff
@@ -0,0 +1,95 @@
...
+import re
+
+from nemo_skills.evaluation.evaluator.math import MathEvaluator
+from nemo_skills.evaluation.math_grader import math_equal
+from nemo_skills.utils import get_logger_name
+
+LOG = logging.getLogger(get_logger_name(__file__))
+
+
+def relaxed_equal(gt_answer: Any, predicted_answer: Any) -> bool:
```
Kipok, line 29: should we update the original math_equal with these changes?

nemo_skills/evaluation/evaluator/dsbench.py
```diff
@@ -0,0 +1,95 @@
...
+    if isinstance(predicted_answer, list):
+        if not isinstance(gt_answer, list):
+            # check if any of the values in predicted_answer are equal to gt_answer
+            return any(relaxed_equal(gt_answer, p) for p in predicted_answer)
+        # check if the lengths are equal and if all the values are equal
+        return len(gt_answer) == len(predicted_answer) and all(relaxed_equal(e, p) for e, p in zip(gt_answer, predicted_answer))
+
+
+    # Try case-insensitive MCQ matching
+    mcq_options = "ABCDEFGHIJKLMN"
```
Kipok: should this go up to Z?

nemo_skills/evaluation/evaluator/dsbench.py
Kipok: if we do that, I guess we'd be able to fully reuse math evaluator here

nemo_skills/evaluation/evaluator/math.py
```diff
@@ -38,6 +38,8 @@ class MathEvaluatorConfig(BaseEvaluatorConfig):
     extract_from_boxed: bool = True
     # only used if extract_from_boxed is False
     extract_regex: str = r"The final answer is (.+)$"
+    # if True: try regex first, then boxed (regardless of extract_from_boxed)
+    relaxed: bool = False
```
Kipok: maybe `relaxed_extraction` is a bit clearer?

nemo_skills/prompt/config/generic/dsbench-da-incontext.yaml
```diff
@@ -0,0 +1,17 @@
+# DSBench Data Analysis - In-context mode (Excel content embedded in prompt)
```
Kipok, line 2: is there a source where this prompt is copied from?

nemo_skills/evaluation/evaluator/dsbench.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Relaxed equality check with:
+    1. Case-insensitive MCQ matching
+    2. Dict/list comparison using math_equal recursively
+    """
+    if predicted_answer is None:
+        return gt_answer is None
+
+    try:
+        parsed_pred = json.loads(predicted_answer)
```
Kipok: should we have a separate try-except for each or is it always the case that both are the same type? Otherwise if one is dict, but the other is string, the string will be left as is even though it could've been converted

Kipok (commented):
thanks, mostly good, just a few minor comments! Could you please also add this to docs?

nemo_skills/dataset/dsbench_da/__init__.py
sgunasekar: Was going back and forth on it: The llm as judge catches about 1-2% - so think it is ok to skip for every run and only use for final reporting. But wanted to leave the config in the doc in case folks so took the middle ground of leaving as comment. Any recommendation on this?

nemo_skills/evaluation/evaluator/dsbench.py
sgunasekar: Yes updating math evaluator would be most cleanest: two issues 
- directly but there was concern that it would make existing workflows inconsistent 
- if mathevaluator is pegged to match some external 3rd party evaluations, it would break those too

One option is to use the "relaxed" argument that is already there for extract_answer and use it to branch into relaxed-mcq. 
All of these are pretty straightforward to implement, so let me know what you prefer and I can make the changes.

nemo_skills/evaluation/evaluator/dsbench.py
sgunasekar: Not in this eval per se - but scenario I am thinking is a benchmark where some questions have short answers and others require some action (e.g., saving files). Then the gt answer can be None.

nemo_skills/dataset/dsbench_da/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+EVAL_SPLIT = "test"
+DATASET_GROUP = "code"
```
Kipok: this is no longer used after a recent pr
```suggestion
```

nemo_skills/evaluation/evaluator/dsbench.py
Kipok: I'd probably make a change directly. It feels like this is the right way to compare things. E.g. if options are A, B, C, D and llm says \\boxed{a}, where A is correct, that should be counted as correct I guess. And the same for the other change.

So my suggestion would be to make a change directly but please run e.g. nano-v3 math eval on maybe comp-math-24-25 and if we get score within normal random variance, we should be good

Kipok: And maybe also some mcq benchmark, eg gpqa

sgunasekar: I would still feel more comfortable doing this as a new PR so that if it breaks anyones workflow they can revert to it. Plus would unblock dsbench for now.

Kipok (approved):
Let's create issues for unfinished items and we can merge as is as long as gpu tests are passing

PR #1265, opened 2026-02-20
feat: add custom judge type support for external repo integration

peri044:
Add generic 'custom' judge type that allows external repos to register their own judge task creators using the locate() pattern. This enables external benchmarks to define judge tasks without modifying nemo-skills source code.

Changes:
- Add _create_custom_judge_tasks() function to dynamically import and call external judge creators
- Add 'custom' judge type routing in prepare_eval_commands()
- Uses existing locate() pattern (same as METRICS_TYPE, eval_type)

## Summary by CodeRabbit

* **New Features**
  * Support for dynamic judge loading by path, including user-provided custom judge creators.
  * Added built-in Comet and NVEmbed judge integrations for translation and embedding-based evaluations.

* **Refactor**
  * Consolidated judge creation to the dynamic loader; removed legacy specialized judge creation paths.
  * Preserved existing LLM-judge fallback and ensured judge model and pipeline args propagate to dynamic judges.

### DIFF
--- a/docs/evaluation/speech-audio.md
+++ b/docs/evaluation/speech-audio.md
@@ -408,3 +408,78 @@ or
 ns prepare_data librispeech-pc --split test-other --data_dir=/path/to/data
 ```
 
+## Numb3rs
+
+Numb3rs is a speech benchmark for evaluating text normalization (TN) and inverse text normalization (ITN) capabilities of audio-language models. It contains paired written/spoken forms with corresponding synthetic audio, allowing evaluation of whether a model transcribes numbers in written form (e.g., `$100`, `3.14`) or spoken form (e.g., `one hundred dollars`, `three point one four`).
+
+**Dataset:** [nvidia/Numb3rs on HuggingFace](https://huggingface.co/datasets/nvidia/Numb3rs)
+
+**Categories:** `ADDRESS`, `CARDINAL`, `DATE`, `DECIMAL`, `DIGIT`, `FRACTION`, `MEASURE`, `MONEY`, `ORDINAL`, `PLAIN`, `TELEPHONE`, `TIME`
+
+**Size:** ~10K samples, ~4.89h total audio duration
+
+### Dataset Location
+
+- Benchmark is defined in [`nemo_skills/dataset/numb3rs/__init__.py`](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/dataset/numb3rs/__init__.py)
+- Original dataset is hosted on [HuggingFace](https://huggingface.co/datasets/nvidia/Numb3rs)
+
+### Key Features
+
+- **Dual reference evaluation**: Each sample has both a written form (`text_tn`) and a spoken form (`text_itn`). WER is computed against both references.
+- **Three prompt variants** generated as separate split files:
+    - `test_neutral`: Neutral transcription prompt ("Transcribe the audio file into English text.")
+    - `test_tn`: Text normalization prompt — expects written form (e.g., `$100`)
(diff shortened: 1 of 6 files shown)


nemo_skills/pipeline/eval.py
```diff
@@ -212,6 +212,66 @@ def _create_nvembed_judge_tasks(
...
+    partition,
+    judge_server_gpus,
+    judge_server_nodes,
+    run_after,
+    reuse_code_exp,
+    reuse_code,
+    installation_command,
+    skip_hf_home_check,
+    sbatch_kwargs,
+):
```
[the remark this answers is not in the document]
Kipok: @peri044 probably a good idea to do this

nemo_skills/pipeline/eval.py
```diff
@@ -694,6 +754,28 @@ def eval(
                     skip_hf_home_check=skip_hf_home_check,
                     sbatch_kwargs=sbatch_kwargs,
                 )
+            elif benchmark_judge_type == "custom":
```
Kipok: instead of maintaining this if-elif list, could you please put those current judges into separate files and use the new locate interface for them direclty? So this block in here will always call create custom judge tasks, but corresponding benchmarks that specify judge_type would instead specify judge_step parameter or something like this that will be a :: path to the corresponding implementation. This way internal and external benchmarks will follow the same logic and it will make eval clearer and more modular

peri044: updated it now. Please take a look

nemo_skills/pipeline/eval.py
```diff
@@ -643,43 +471,29 @@ def eval(
             benchmark_args.eval_subfolder = benchmark_args.eval_subfolder[4:]
             judge_pipeline_args["output_dir"] = str(Path(output_dir) / benchmark_args.eval_subfolder)
 
-            # Check for per-benchmark judge_type, fall back to global judge_type
-            benchmark_judge_type = judge_pipeline_args.pop("judge_type", judge_type)
+            # judge_step is a :: path to the judge creator function (locate() convention).
+            # Benchmarks set this directly in JUDGE_PIPELINE_ARGS; falls back to None for LLM judge.
+            judge_creator_path = judge_pipeline_args.pop("judge_step", None)
```
Kipok, line 476: can we have this as an explicit pipeline parameter? Similar to how judge_type was? That way it will be simpler to update from command line

nemo_skills/pipeline/eval.py
```diff
@@ -643,43 +471,29 @@ def eval(
...
+                from nemo_skills.dataset.utils import locate
+
+                judge_creator_fn = locate(judge_creator_path)
+
+                # Call with standardized parameters
+                judge_tasks = judge_creator_fn(
                     exp=exp,
                     expname=expname,
-                    benchmark=benchmark,
+                    benchmarks=[benchmark],
```
Kipok, line 492: what's the reason to have multiple benchmarks here?

Kipok (commented):
Could you please update a reference to judge_type in the docs in here /home/igitman/workspace/NeMo-Skills/docs/evaluation/multilingual.md?

Also can you please re-create from a branch, so that our tests can run?

nemo_skills/pipeline/eval.py
peri044: Renamed the judge_step to judge_path and added it as a pipeline parameter.

nemo_skills/pipeline/eval.py
peri044: No specific reason. Reverted this.

PR #1272, opened 2026-02-23
Add MMLU-Pro 10% optimized subset for checkpoint selection

ka00ri:
PR to support subsampling of MMLU-Pro benchmark.

- Add hash-based subset filtering for MMLU-Pro.
- Add support for `10pct_opt_v1` subset (1,203 questions) optimized for checkpoint selection while preserving topic balance and full score tracking 
- Add utility funcs for question/options normalization, subset ID loading, and filtering
- Validate e2e with:
  - `ns prepare_data mmlu-pro --split=10pct_opt_v1`
  - `ns eval --benchmarks=mmlu-pro --split=10pct_opt_v1 ...`
- Raise `ValueError` when subset IDs do not fully match source split entries

## Summary by CodeRabbit

* **New Features**
  * Added support for a new dataset split option "10pct_opt_v1" when preparing MMLU-Pro.
  * Preparation now supports subset-based filtering: it can load and validate a subset list and produce a filtered dataset for non-standard splits.
  * Validation and test splits retain existing behavior; non-standard splits now write filtered results to the corresponding split output.


nemo_skills/dataset/mmlu-pro/subsets/10pct_opt_v1.txt
```diff
@@ -0,0 +1,1203 @@
+{"id": "004d49ebfd29e6a0ac84ddd9f7386287ecb5704dc428b09ada25ae7cc309905a", "question": "According to Scott (1990), a document is 'authentic' if it:"}
```
Kipok: can we just keep the hashes / ids? Do we have to also keep the "question" here?

PR #1295, opened 2026-03-07
eval kit support v2

Jorjeous:
Added ability to run eval kit in namo skills nativly, runnning benchmarks from EK in nemo skills with it's path and NS benchmarks throug EK path 

## Summary by CodeRabbit

* **New Features**
  * VLMEvalKit integration: run benchmarks in two modes (in-process Megatron or vLLM client), including self-contained in-process workflows and multi-GPU merging.
  * Ability to consume pre-computed VLMEvalKit aggregates so external benchmark metrics can be imported and reported.

* **Documentation**
  * New user guide and index entry covering setup, modes, available benchmarks, sample commands, config snippets, results interpretation, and troubleshooting.

### DIFF
--- a/docs/evaluation/eval-kit.md
+++ b/docs/evaluation/eval-kit.md
@@ -1,0 +1,282 @@
+# VLMEvalKit Integration (eval_kit)
+
+This page explains how to run VLMEvalKit benchmarks through NeMo Skills using the `eval_kit` generation module. This enables evaluating Megatron multimodal models on VLMEvalKit's benchmark collection (MMBench, LibriSpeech, TedLium, etc.) without leaving the NeMo Skills pipeline.
+
+## Overview
+
+Two inference modes are available:
+
+| Mode | How it works | When to use |
+|------|-------------|-------------|
+| **mcore** | Megatron model loaded in-process via `torchrun` (no HTTP server) | Megatron checkpoints |
+| **vllm** | NeMo Skills starts a vLLM server, VLMEvalKit connects as client | HF models served by vLLM |
+
+Both modes use the same pipeline command — the only difference is the `++model_type` flag.
+
+## Prerequisites
+
+Before running eval_kit benchmarks, you need four things set up:
+
+### 1. VLMEvalKit source code (local)
+
+The `vlmeval/` directory from VLMEvalKit gets packaged and shipped to the cluster automatically. You need a local clone:
+
+```bash
+# Clone VLMEvalKit (NVIDIA internal fork with MultiModalMCore support)
+git clone VLMEvalKitMcore /path/to/VLMEvalKitMcore
+```
+
+Then set the environment variable **before running any `ns eval` command**:
+
+```bash
+export NEMO_SKILLS_VLMEVALKIT_PATH=/path/to/VLMEvalKitMcore
+```
+
+!!! important
+    This path is read **locally at submission time**. The pipeline packages the `vlmeval/` subdirectory and rsyncs it to the cluster. It does NOT need to exist on the cluster.
+
+### 2. eval_kit container on the cluster
+
+The eval_kit container must have PyTorch, Megatron, and VLMEvalKit dependencies pre-installed. Add it to your cluster config:
+This container can be found in container storage
+
+```yaml
+# cluster_configs/my_cluster.yaml
+containers:
+  eval_kit: /path/to/eval-kit-nemo-skills.sqsh
+  # ... other containers
+```
+
+### 3. Megatron path (for mcore mode)
+
+The container needs access to a Megatron-LM installation. Set it in your cluster config:
+
+```yaml
+env_vars:
+  - MEGATRON_PATH=/path/to/megatron-lm
+  - PYTHONPATH=/path/to/megatron-lm
+```
+
+And ensure the path is mounted:
+
+```yaml
+mounts:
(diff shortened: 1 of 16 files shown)


docs/evaluation/eval-kit.md
```diff
@@ -0,0 +1,282 @@
...
+## Prerequisites
+
+Before running eval_kit benchmarks, you need four things set up:
+
+### 1. VLMEvalKit source code (local)
+
+The `vlmeval/` directory from VLMEvalKit gets packaged and shipped to the cluster automatically. You need a local clone:
+
+```bash
+# Clone VLMEvalKit (NVIDIA internal fork with MultiModalMCore support)
```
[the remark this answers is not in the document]
Kipok: there we can also potentially just add it as a submodule to ensure it's always packaged without the need for special env vars

nemo_skills/inference/mcore_skills.py
```diff
@@ -0,0 +1,547 @@
...
+
+            # Content: str or list of content items (text, image_url)
+            if isinstance(content, str):
+                if content.strip():
+                    text_parts.append(content.strip())
+            elif isinstance(content, list):
+                for item in content:
+                    if isinstance(item, dict):
+                        if item.get("type") == "text" and "text" in item:
+                            text_parts.append(item["text"].strip())
```
melllinia, line 316: Text parts are accumulated and appended at the end, after all media items. This can destroy text/media ordering ("Describe [image1] vs [image2]" becomes [image1, image2, "Describe vs"]). 

Suggestion: emit entries into the `mcore` list in-order during iteration, rather than buffering all text for a final append.

nemo_skills/inference/mcore_skills.py
```diff
@@ -0,0 +1,547 @@
...
+                        }
+                    )
+
+            # Re-write output.jsonl with cleaned generations
+            with open(output_file, "w", encoding="utf-8") as fout:
+                for entry in entries:
+                    fout.write(json.dumps(entry) + "\n")
+
+            # Compute WER using VLMEvalKit (same function as eval_kit path)
+            wer_score = asr_wer(results)
```
melllinia, line 511: This always calls `asr_wer()` regardless of the benchmark type. For non-ASR datasets, this may produce meaningless metrics. Can we filter based on `dataset_name` or a `metric_type` config field?

nemo_skills/evaluation/evaluator/audio.py
```diff
@@ -505,13 +505,35 @@ def evaluate_sample(sample: dict[str, Any], config: AudioEvaluatorConfig) -> dic
...
+    generation_raw = sample.get("generation")
+    generation = generation_raw.strip() if isinstance(generation_raw, str) else ""
     expected_answer = sample.get("expected_answer", "").strip()
 
     # Strip helpful prefixes for ASR tasks (e.g., "The audio says: ...")
     if config.strip_helpful_prefixes:
         generation = strip_helpful_prefixes(generation)
 
+    # Normalise AudioBench speech-translation task types (ST-EN-ZH -> Translation)
+    _ASR_TYPES = {"ASR", "ASR-ZH", "ASR-PC", "ASR_LEADERBOARD"}
```
melllinia, line 517: `ASR_LEADERBOARD` is referenced in `_ASR_TYPES` but `asr-leaderboard` dataset preparation is updated and uses `task_type="ASR"` because the evaluation for two types were identical. Can we consider safe removal of `ASR_LEADERBOARD` references to avoid confusion, please?

nemo_skills/pipeline/utils/eval.py
```diff
@@ -461,10 +521,31 @@ def prepare_eval_commands(
...
+                # Self-contained tasks don't use NeMo Skills server, so skip
+                # server-related args that configure_client adds to job_extra_arguments.
+                # Tasks with configure_client_overrides translate server params into
+                # their own config format (e.g. eval_kit uses flat ++server_url instead
+                # of nested ++server.* overrides).
+                if benchmark_args.self_contained_task:
+                    effective_extra_args = extra_arguments
+                elif hasattr(generation_task, "configure_client_overrides"):
+                    # rsplit to handle URLs like http://host:port (takes last colon)
+                    host, port = (job_server_address or "localhost:5000").rsplit(":", 1)
```
melllinia, line 533: For a URL like `http://host:5000`, `rsplit(":", 1)` produces `host = "http://host"`, which could cause malformed URLs downstream. Can we use `urllib.parse.urlparse` for proper parsing here?

nemo_skills/dataset/eval_kit/__init__.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
melllinia, line 1: Can you please update the copyright from 2025 to 2026 for newly created files?

melllinia (changes requested):
Overall looks good, can you please address the comments and since there are core changes let's wait for @Kipok or @gwarmstrong review as well, thank you!

Jorjeous:
@melllinia on the way + lets quick chat offline

Kipok:
I don't think we will merge this, I think there is some work to get an equivalent functionality merged into gym. Please ping me in slack if you think this should be merged in skills or have questions

PR #1364, opened 2026-04-15
fix: align HLE answer extraction with prompt format

tamohannes:
## Summary
- align HLE answer extraction with the `Explanation:` / `Answer:` / `Confidence:` format used by `generic/hle`
- apply the same extraction override to `hle_verified`
- add focused regression tests for plain answer lines, markdown-styled answers, boxed fallback, and exact-match outputs

## Why
HLE prompts do not require `\\boxed{...}`, but the default math extraction path is box-first. In practice this can leave `predicted_answer` unset even when the answer is clearly present on the `Answer:` line. This patch keeps HLE extraction aligned with the benchmark prompt while preserving boxed fallback for models that still emit boxed answers.

This does not change the main HLE judge path, which already reads the full raw generation. It fixes extracted-answer artifacts and symbolic bookkeeping for HLE-style outputs.

## Test plan
- `python -m pytest tests/test_hle_answer_extraction.py -q`
- local non-live pytest collection in my environment was blocked by a missing optional dependency (`soundfile`) in `tests/test_magpie_tts_backend.py`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved answer extraction: case-insensitive, multi-line matching, prefers non-empty groups, trims whitespace, and returns cleaner results

* **Refactor**
  * HLE generation defaults updated so enhanced extraction settings are applied consistently across HLE datasets

* **Tests**
  * Added tests for plain, markdown-styled, boxed, and free-form Answer: extraction cases

* **Chores**
  * CI now conditionally enables registry cache and registry login when credentials are available


.github/workflows/tests.yml
```diff
@@ -53,15 +56,30 @@ jobs:
         # Python generates: locally-built-{sanitized_path}:{sha256_hash[:12]}
         REPO_LOWER=$(echo "${{ github.repository }}" | tr '[:upper:]' '[:lower:]')
 
+        # Fork PRs do not receive repository secrets, so skip remote cache usage
+        # when GHCR_PAT is unavailable.
+        if [ "$HAS_GHCR_PAT" = "true" ]; then
```
gwarmstrong, line 61: we  still need secrets for the tests dependent on API keys, so I'm not sure this accomplishes much

gwarmstrong (changes requested):
I think we can remove the build changes since we still need secrets for test functionality. The rest lgtm. Please push to a branch on this fork and resubmit.

tamohannes:
Superseded by #1373 — pushed to an upstream branch and removed CI changes per reviewer feedback.

PR #1205, opened 2026-01-31
Add arena-hard v2

Kipok:
Add the following fixes on top of https://github.com/NVIDIA-NeMo/Skills/pull/1152
1. Add an appropriate handling for different baselines for hard prompts and writing prompts
2. Add an appropriate judge prompt for writing which doesn't ask to give a response first
3. Add category in metrics reporting and tests

## Release Notes

* **New Features**
  * Added evaluation support for the arena-hard-v2 benchmark with per-category metrics tracking and aggregation
  * Implemented category-specific judging for improved evaluation accuracy across different task types including hard prompts and creative writing

* **Documentation**
  * Added comprehensive setup and execution guides for arena-hard and arena-hard-v2 evaluations, including sample commands, environment configuration, and results examples

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1368, opened 2026-04-16
Fix arena-hard (v1) judge KeyError by dropping category tag at prep time

### LINKED ISSUE #1367: arena-hard (v1) judge crashes with KeyError: 'arena-hard-v0.1'
## Summary
Running the judge stage of the **arena-hard (v1)** benchmark crashes with `KeyError: 'arena-hard-v0.1'` the moment `log_example_prompt` reaches the first data point, so no judgement is ever produced and `eval-results/arena-hard/` never gets `output.jsonl` / `metrics.json`.

## Reproduction
```bash
ns eval --benchmarks arena-hard --model <any model> ...
```

## Root cause
`arena-hard` (v1) dataset's `test.jsonl` has `category: "arena-hard-v0.1"` on every row — it's a dataset-version tag, not one of the real v2 categories. But `ArenaJudgeTask.setup_prompt` only registers the v2 categories in `self.category_prompts`:

https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/inference/eval/arena_judge.py#L96-L108

```python
self.category_prompts = {}
if self.cfg.prompt_config_creative:
    self.category_prompts["creative_writing"] = get_prompt(...)
# registering default prompt explicitly for hard_prompt
self.category_prompts["hard_prompt"] = default_prompt
```

And `fill_prompt` then raises immediately on unknown categories:

https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/inference/eval/arena_judge.py#L120-L125

```python
category = data_point.get("category")
if not category:
    prompt = self.prompt
else:
    # will fail if category not in category_prompts as this is unexpected
    prompt = self.category_prompts[category]
```

The comment two lines earlier (L87) explicitly states the default prompt is "used for most categories including hard_prompt, arena-hard-v0.1, etc." — so the stated intent is that `arena-hard-v0.1` uses the default, but the dict never contains that key.

## Stack trace
```
File ".../nemo_skills/inference/eval/arena_judge.py", line 140, in log_example_prompt
    "Example prompt:\nData dictionary: %s\nPrompt: %s", data_point, self.fill_prompt(data_point, all_data)
File ".../nemo_skills/inference/eval/arena_judge.py", line 113, in fill_prompt
    prompt = self.category_prompts[category]
KeyError: 'arena-hard-v0.1'
```

## Proposed fix
Register `arena-hard-v0.1` as pointing to the default prompt in `setup_prompt` — matches the existing L87 comment, keeps the fail-fast behavior for truly unknown categories. PR to follow.

## Regression range
Introduced in commit d820200e ("Add arena-hard v2", #1205, 2026-02-03) when the multi-category prompt dispatch was added. arena-hard-v2 works (its categories match) but v1 has been broken since.

## Environment
- Skills main, arena-hard (v1) benchmark
- Any prompt_format other than `openai` (the openai early-return path bypasses the bug)

bzantium:
Closes #1367

## Summary
Drop the `category` field from arena-hard (v1) `test.jsonl` at prep time. This removes the `"arena-hard-v0.1"` dataset-version string that `lmarena/arena-hard-auto` ships in the category field, letting `ArenaJudgeTask.fill_prompt` reach its already-intended default-prompt fallback (`if not category: prompt = self.prompt`).

## Why a data-side fix?
Downstream consumers were already written assuming v1 has no category:

- `arena_metrics.py` L57: `# Track category for per-category scoring (defaults to None for v1 compatibility)`
- `arena_metrics.py` L109: per-category breakdown only fires when `len(unique_categories) > 1`, so v1's `{None}` is already skipped
- `arena_judge.py` L121: `if not category: prompt = self.prompt` is already in place

The mismatch is that `lmarena/arena-hard-auto` writes `"arena-hard-v0.1"` (a version tag, not a real category) into the category field, which `fill_prompt`'s else branch fails on via strict dict lookup. Dropping the noisy field at prep time aligns the data with the existing "no category" assumption rather than patching each consumer.

Regression introduced in #1205 when multi-category prompt dispatch was added. v2 behavior unchanged (its `hard_prompt` / `creative_writing` values are real categories and remain populated).

## Changes
- `nemo_skills/dataset/arena-hard/prepare.py`: pop `category` when writing `test.jsonl`.
- `nemo_skills/inference/eval/arena_judge.py`: drop the inaccurate `including arena-hard-v0.1` from the default-prompt comment.

## Test plan
- [x] `ns eval --benchmarks arena-hard --model <small model> ...` runs the judge to completion and writes `eval-results/arena-hard/output.jsonl` / `metrics.json`.
- [x] arena-hard-v2 unaffected — v2 `test.jsonl` still ships real category values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure prompt selection no longer relies on a preserved `category` field in processed dataset records, restoring correct fallback behavior for unmapped prompt categories.
* **Documentation**
  * Clarified default prompt description so fallback behavior for unmapped categories is accurately documented.

### DIFF
--- a/nemo_skills/dataset/arena-hard/prepare.py
+++ b/nemo_skills/dataset/arena-hard/prepare.py
@@ -50,4 +50,9 @@ if __name__ == "__main__":
             data = json.loads(line)
             data["question"] = data.pop("prompt")
             data["baseline_answer"] = baseline_answers[data["uid"]]
+            # The upstream lmarena/arena-hard-auto source stores a dataset-version
+            # string ("arena-hard-v0.1") in the category field. v1 has no real
+            # sub-categories, so drop it to let ArenaJudgeTask.fill_prompt fall
+            # through to the default prompt via its `if not category` branch.
+            data.pop("category", None)
             fout.write(json.dumps(data) + "\n")
--- a/nemo_skills/inference/eval/arena_judge.py
+++ b/nemo_skills/inference/eval/arena_judge.py
@@ -84,7 +84,7 @@ class ArenaJudgeTask(GenerationTask):
         if self.cfg.prompt_format == "openai":
             return None
 
-        # Load the default prompt (used for most categories including hard_prompt, arena-hard-v0.1, etc.)
+        # Load the default prompt (used for any category not explicitly mapped below, including hard_prompt)
         default_prompt = get_prompt(
             prompt_config=self.cfg.prompt_config,
             tokenizer=self.tokenizer,


bzantium:
@coderabbitai review again now.

bzantium:
@Kipok, I'm fixing a bug in arena-hard caused by the recent arena-hard-v2 logic changes. It missed handling the 'arena-hard-v0.1' category in the data.

Kipok (approved):

PR #1373, opened 2026-04-16
fix: align HLE answer extraction with prompt format

tamohannes:
## Summary

* align HLE answer extraction with the `Explanation:` / `Answer:` / `Confidence:` format used by `generic/hle`
* apply the same extraction override to `hle_verified`
* add focused regression tests for plain answer lines, markdown-styled answers, boxed fallback, and exact-match outputs

## Why

HLE prompts do not require `\boxed{...}`, but the default math extraction path is box-first. In practice this can leave `predicted_answer` unset even when the answer is clearly present on the `Answer:` line. This patch keeps HLE extraction aligned with the benchmark prompt while preserving boxed fallback for models that still emit boxed answers.

This does not change the main HLE judge path, which already reads the full raw generation. It fixes extracted-answer artifacts and symbolic bookkeeping for HLE-style outputs.

## Changes from previous PR (#1364)

Removed all `.github/workflows/tests.yml` changes per reviewer feedback — secrets are still needed for test functionality.

## Test plan

* `python -m pytest tests/test_hle_answer_extraction.py -q`

## Summary by CodeRabbit

* **New Features**
  * Added a configurable extraction override for answer extraction in math evaluations.

* **Improvements**
  * Extraction now prefers non-empty captures, trims whitespace, and avoids consuming following metadata. Extraction configuration survives shell/pipeline round-trips.

* **Tests**
  * Added unit and integration tests covering plain, markdown, boxed, exact-match, empty-answer, and shell round-trip scenarios.


Kipok (commented):
@tamohannes what do we need this for? Not fully sure why this feature is needed given that predicted answer is only really relevant for majority and it's not computed with metrics type "hle"

Kipok:
Let me close this for now but feel free to reopen if there is some reason we need this that I missed

PR #1375, opened 2026-04-17
Add Korean benchmarks: KMMLU, KMMLU-Pro, KMMLU-Redux, KoIFEval, KorMedMCQA, KoSimpleQA

### LINKED ISSUE #1374: Add Korean benchmarks: KMMLU, KMMLU-Pro, KMMLU-Redux, KoIFEval, KorMedMCQA, KoSimpleQA
## Summary

Add six Korean-language evaluation benchmarks covering multiple-choice knowledge, instruction following, and factual QA.

| Benchmark | Type | Source | Samples | Notes |
|---|---|---|---|---|
| **kmmlu** | MCQ (4-choice) | [HAERAE-HUB/KMMLU](https://huggingface.co/datasets/HAERAE-HUB/KMMLU) | ~35k | 45 categories across HUMSS/STEM/Applied/Other |
| **kmmlu-pro** | MCQ (5-choice) | [LGAI-EXAONE/KMMLU-Pro](https://huggingface.co/datasets/LGAI-EXAONE/KMMLU-Pro) | ~2.8k | Professional-level Korean MCQ |
| **kmmlu-redux** | MCQ (4-choice) | [LGAI-EXAONE/KMMLU-Redux](https://huggingface.co/datasets/LGAI-EXAONE/KMMLU-Redux) | ~2.6k | Curated/cleaned subset of KMMLU |
| **kormedmcqa** | MCQ (5-choice) | [sean0042/KorMedMCQA](https://huggingface.co/datasets/sean0042/KorMedMCQA) | ~12k | Korean medical exam (dentist/doctor/nurse/pharm) |
| **koifeval** | Instruction Following | [allganize/IFEval-Ko](https://huggingface.co/datasets/allganize/IFEval-Ko) | 342 | Korean port of IFEval with 25+ instruction types |
| **kosimpleqa** | Factual QA (judge) | [bzantium/KoSimpleQA](https://huggingface.co/datasets/bzantium/KoSimpleQA) | 1,000 | Korean SimpleQA; reuses existing SimpleQA judge infra |

## Implementation approach

- **MCQ (4 benchmarks)**: Reuse existing `eval_type=multichoice` (mcq.py) + `METRICS_TYPE="multichoice"` (MathMetrics). Add a Korean `정답:` regex fallback in `mcq.py` alongside the existing English/Arabic/Bengali/Japanese patterns. Two new Korean prompt configs (`eval/korean/mcq-4choices.yaml`, `eval/korean/mcq-5choices.yaml`).
- **KoIFEval**: Reuse existing `IFMetrics` for aggregation. Add a Korean-specific evaluator (`koifeval.py`) that handles Korean sentence splitting, word counting, and 25+ instruction compliance checks.
- **KoSimpleQA**: Fully reuses existing SimpleQA infrastructure — same metrics class, same judge prompt (`judge/simpleqa`), same o3-mini judge pipeline. Only a new `prepare.py` and `__init__.py`.

### New files
- `nemo_skills/dataset/{kmmlu,kmmlu-pro,kmmlu-redux,koifeval,kormedmcqa,kosimpleqa}/` — `__init__.py` + `prepare.py` each
- `nemo_skills/prompt/config/eval/korean/mcq-4choices.yaml`, `mcq-5choices.yaml`
- `nemo_skills/evaluation/evaluator/koifeval.py`

### Modified files
- `nemo_skills/evaluation/evaluator/__init__.py` — register `"koifeval"` eval_type
- `nemo_skills/evaluation/evaluator/mcq.py` — add Korean `정답:` regex fallback (+5 lines)

bzantium:
Closes #1374

Adds six Korean evaluation benchmarks. Reuses existing nemo-skills infrastructure where it works; the only shared-code touchpoint outside the new benchmark directories is a pip dependency for the IFEval-Ko evaluator package.

## MCQ (4 benchmarks)

| Benchmark | Source | Choices | Prompt |
|---|---|---|---|
| kmmlu | [HAERAE-HUB/KMMLU](https://huggingface.co/datasets/HAERAE-HUB/KMMLU) (45 subjects) | 4 | `eval/korean/mcq-4choices` |
| kmmlu-pro | [LGAI-EXAONE/KMMLU-Pro](https://huggingface.co/datasets/LGAI-EXAONE/KMMLU-Pro) | 5 | `eval/korean/mcq-5choices` |
| kmmlu-redux | [LGAI-EXAONE/KMMLU-Redux](https://huggingface.co/datasets/LGAI-EXAONE/KMMLU-Redux) | 4 | `eval/korean/mcq-4choices` |
| kormedmcqa | [sean0042/KorMedMCQA](https://huggingface.co/datasets/sean0042/KorMedMCQA) (dentist / doctor / nurse / pharm) | 5 | `eval/korean/mcq-5choices` |

All four use the existing `eval_type=multichoice` (mcq.py) and `METRICS_TYPE="multichoice"`. The Korean prompt instructs the model to end its response with `정답: A/B/C/D[/E]`. Each MCQ entry sets a per-sample `extract_regex` that matches that exact line (along with `extract_from_boxed=False` and `relaxed=False`), so non-conforming outputs surface as wrong rather than being rescued by a greedy fallback. `mcq.py` stays English-centric.

`kmmlu` and `kormedmcqa` previously hardcoded their per-subject category lists; both now load all subjects in a single `load_dataset` call by globbing the per-config files.

## Factual QA

**kosimpleqa** uses [bzantium/KoSimpleQA](https://huggingface.co/datasets/bzantium/KoSimpleQA) (1,000 samples, 10 topics). Reuses the existing SimpleQA infrastructure: same `SimpleQAMetrics`, same `judge/simpleqa` prompt, same o3-mini judge pipeline. The only divergence from the English setup is `subset_for_metrics: topic`, which gives a per-topic breakdown without changing the aggregate metric.

## Instruction Following

**koifeval** uses [allganize/IFEval-Ko](https://huggingface.co/datasets/allganize/IFEval-Ko) (342 samples). The actual instruction checkers (Korean sentence splitting, word counting, and the `Instruction` subclasses) live in a separate pip-installable package: https://github.com/bzantium/koifeval (Apache-2.0, derived from `google-research/instruction_following_eval` and `allganize/IFEval-Ko`). The evaluator in this repo is a 60-line subprocess wrapper that mirrors `ifeval.py`. The package is pinned to a specific commit in `core/requirements.txt`.

## Changes

| File | Change |
|---|---|
| `nemo_skills/dataset/{kmmlu,kmmlu-pro,kmmlu-redux,koifeval,kormedmcqa,kosimpleqa}/` | New benchmark directories |
| `nemo_skills/prompt/config/eval/korean/mcq-{4,5}choices.yaml` | Korean MCQ prompts |
| `nemo_skills/evaluation/evaluator/koifeval.py` | 60-line subprocess wrapper for the koifeval package |
| `nemo_skills/evaluation/evaluator/__init__.py` | Register `koifeval` eval_type (+1 line) |
| `core/requirements.txt` | Pin `koifeval` package to a specific commit (+1 line) |

## Release Notes

* **New Features**
  * Added support for six new Korean benchmark datasets: KMMLU, KMMLU-Pro, KMMLU-Redux, KorMedMCQA, KoSimpleQA, and KoIFEval
  * Introduced Korean instruction-following evaluation with strict and loose scoring modes
  * Added Korean multiple-choice prompt templates for 4-choice and 5-choice formats


bzantium:
## Evaluation Results

Tested with two models on the six Korean benchmarks added by this PR:

### MCQ Benchmarks

| Benchmark | kanana-2-30b-a3b-instruct-2601 | Qwen3-30B-A3B-Instruct-2507 |
|---|:---:|:---:|
| **kmmlu** (symbolic_correct) | 54.22 | 67.82 |
| &nbsp;&nbsp;HUMSS | 59.73 | 65.13 |
| &nbsp;&nbsp;STEM | 56.34 | 71.27 |
| &nbsp;&nbsp;Applied Science | 51.03 | 68.74 |
| &nbsp;&nbsp;Other | 52.74 | 64.14 |
| **kmmlu-pro** | 50.43 | 56.98 |
| **kmmlu-redux** | 51.22 | 61.85 |
| **kormedmcqa** | 68.66 | 77.67 |
| &nbsp;&nbsp;dentist | 56.35 | 61.65 |
| &nbsp;&nbsp;doctor | 62.53 | 75.17 |
| &nbsp;&nbsp;nurse | 78.25 | 86.10 |
| &nbsp;&nbsp;pharm | 74.92 | 84.85 |

### Instruction Following

| Metric | kanana-2-30b-a3b-instruct-2601 | Qwen3-30B-A3B-Instruct-2507 |
|---|:---:|:---:|
| **koifeval** prompt_strict_accuracy | 78.07 | 80.41 |
| instruction_strict_accuracy | 83.58 | 85.45 |
| prompt_loose_accuracy | 82.16 | 83.33 |
| instruction_loose_accuracy | 86.49 | 87.73 |

### Factual QA (Judge-Based)

| Metric | kanana-2-30b-a3b-instruct-2601 | Qwen3-30B-A3B-Instruct-2507 |
|---|:---:|:---:|
| **kosimpleqa** F1 | 49.90 | 13.81 |
| correct | 49.80 | 13.80 |
| incorrect | 49.80 | 86.10 |
| not_attempted | 0.40 | 0.10 |

All six benchmarks run end-to-end: prepare → generation → evaluation/judge → metrics aggregation.

shuoyangd:
Hi @bzantium! Thanks for contributing to Nemo-Skills. I'll be reviewing both PRs (#1375 #1378).

To start, may I ask if you are using these benchmarks for evaluation in Nemo-Skills? I don't have a lot of exposure on Korean LLM benchmarking, but from my perspective, we wouldn't want to inflate our list of datasets too much by including some benchmarks that are sub-par in quality and end up with limited adoptions. We should also consider create some hierarchy in `datasets` folder and group them into one directory.

My colleague @eagle705 will help assess the quality of those benchmarks. I'll follow-up once I've taken a closer look at the code.

bzantium:
Hi @shuoyangd, thanks for the thoughtful review!

To answer your question:

**1) Are we using these for evaluation?**
Yes, these are part of our active evaluation workflow.

**2) On benchmark quality and adoption:**
- KMMLU and the related Korean benchmarks in this PR are widely used as standard evaluation suites in the Korean LLM community.
- KoIFEval and KoMedCPQA are actively used within our team for Korean instruction-following and medical QA evaluation, respectively.
- Several Korean technical reports rely on these datasets for evaluation. A few examples:
  - https://arxiv.org/abs/2601.09200
  - https://arxiv.org/abs/2604.08644
  - https://arxiv.org/abs/2502.18934

**3) On the directory hierarchy:**
I fully agree that grouping would be cleaner. Happy to reorganize the additions under something like `datasets/korean/` (or whatever structure you and the maintainers prefer) so we don't bloat the top-level dataset list. Just let me know your preference and I'll restructure both PRs accordingly.

Looking forward to @eagle705's input on the quality side as well. Thanks again!

nemo_skills/evaluation/evaluator/mcq.py
```diff
@@ -101,6 +101,12 @@ def extract_letter(
             if match:
                 parsed_letter = match[-1].strip().upper()
 
+        # Korean MCQ format: "정답: A" or "정답은: A"
+        if parsed_letter is None:
+            match = re.findall(r"(?i)[\*\_]{0,2}정답[^A-Z]*?[\*\_]{0,2}\s*:?[\s\*\_]{0,2}\s*([A-Z])(?![a-zA-Z0-9])", text)
```
naymaraq: Adding Korean symbols here is not desirable, as this regex is not truly multilingual.
Instead, you can provide your own regexes in prepare.py. See how we did it for the mmmlu benchmark.

nemo_skills/evaluation/evaluator/koifeval.py
```diff
@@ -0,0 +1,1448 @@
...
+    # removing temporary metric directory to avoid reusing it
+    shutil.rmtree(output_dir)
+
+
+# ============================================================================
+# Korean Instructions Utility Functions
+# ============================================================================
+
+# ISO 639-1 codes to language names.
+LANGUAGE_CODES = immutabledict.immutabledict(
```
naymaraq: Why do we need all these languages if the benchmark is for Korean?

naymaraq (changes requested):

nemo_skills/evaluation/evaluator/mcq.py
bzantium: Good point. Followed the mmmlu pattern. Reverted the language specific fallback in `mcq.py`. Each Korean MCQ `prepare.py` (kmmlu, kmmlu-pro, kmmlu-redux, kormedmcqa) now defines its own `_EXTRACT_REGEX = [...]` inline and injects it per sample, keeping the shared evaluator language agnostic.

nemo_skills/evaluation/evaluator/koifeval.py
bzantium: The dict is needed by `ResponseLanguageChecker`, and IFEval-Ko itself contains 25 samples whose `language:response_language` kwargs target non-Korean languages (e.g. "respond in German/Russian/…"). Verified against the dataset: 17 distinct language codes appear (`ar, bn, de, gu, hi, kn, ko, mr, ne, pa, pt, ru, sw, te, th, ur, vi`), all explicitly specified (no `random.choice` fallback path is exercised). Trimmed `LANGUAGE_CODES` from 30 → exactly those 17 used by the benchmark.

shuoyangd: @bzantium Can you use [this package](https://pypi.org/project/iso639-lang) instead? It's already a dependency

bzantium: Good suggestion. Replaced the hardcoded `LANGUAGE_CODES` dict with `from iso639 import Lang` and use `Lang(code).name` inside `ResponseLanguageChecker.build_description`. Also removed the now-unused `import immutabledict`. The `random.choice` fallback for `language=None` is replaced by a default of `"ko"` since this benchmark is Korean only and IFEval-Ko's data always specifies `language` explicitly.

nemo_skills/dataset/kmmlu-pro/prepare.py
```diff
@@ -0,0 +1,74 @@
...
+}
+
+# Korean MCQ answer-line patterns (e.g. "정답: A", "답변: A", "답: A") plus a
+# greedy single-letter fallback. Injected per-sample via `extract_regex` so
+# the shared MCQ evaluator stays language-agnostic.
+_EXTRACT_REGEX = [
+    r"(?i)정답\s*은?\s*[:：]?[ \t]*[\*\_]{0,2}\(?([A-E])(?![a-zA-Z0-9])",
+    r"(?i)답변\s*[:：]?[ \t]*[\*\_]{0,2}\(?([A-E])(?![a-zA-Z0-9])",
+    r"(?i)답\s*[:：][ \t]*[\*\_]{0,2}\(?([A-E])(?![a-zA-Z0-9])",
+    r"[\s\S]*\b\(?\s*([A-E])\s*\)?\.?\b",
```
shuoyangd: I'm wondering if you really want so many layers of fall-back? You may get random things like "Answer: A" or "Antwort: A" which may not be entirely desirable as a Korean benchmark. Even something that's Korean but different from what you instructed the model to generate in the prompt wrapper is a sign of non-instruction following behavior. In general I think it's better to expose them rather than covering them up with a high number, but it's your call.

(Same with other MCQ datasets, since you use the same set of fallback regexes)

nemo_skills/dataset/kmmlu-pro/prepare.py
```diff
@@ -0,0 +1,74 @@
...
+]
+
+
+def format_entry(entry):
+    return {
+        "expected_answer": answer_map[entry["solution"]],
+        "extract_from_boxed": False,
+        "extract_regex": _EXTRACT_REGEX,
+        "relaxed": False,
+        "subset_for_metrics": entry["license_name"],
```
shuoyangd: Do you usually use `license_name` or `subject`? Both seem good to me. Just something worth thinking twice.

nemo_skills/dataset/kmmlu-redux/prepare.py
```diff
@@ -0,0 +1,73 @@
...
+]
+
+
+def format_entry(entry):
+    return {
+        "expected_answer": answer_map[entry["solution"]],
+        "extract_from_boxed": False,
+        "extract_regex": _EXTRACT_REGEX,
+        "relaxed": False,
+        "subset_for_metrics": entry["category"],
```
shuoyangd: Same question here. `category` or `license_name`?

nemo_skills/dataset/kmmlu/prepare.py
```diff
@@ -0,0 +1,178 @@
...
+import json
+from pathlib import Path
+
+from datasets import concatenate_datasets, load_dataset
+from tqdm import tqdm
+
+from nemo_skills.dataset.utils import get_mcq_fields
+
+
+categories = [
```
shuoyangd: Looks like the only place where the big `categories` dictionary is used is L165. I wonder if we can load the full hf dataset without having to iterate through the categories, and if we can get the list of categories.

Claude gave me the following answer to the second question:

```
from datasets import get_dataset_config_names
cats = get_dataset_config_names("HAERAE-HUB/KMMLU")  # 45 subjects
```

So I think we can let this go at least.

nemo_skills/dataset/kmmlu/prepare.py
```diff
@@ -0,0 +1,178 @@
...
+def write_data_to_file(output_file, data):
+    with open(output_file, "wt", encoding="utf-8") as fout:
+        for entry in tqdm(data, desc=f"Writing {output_file.name}"):
+            json.dump(format_entry(entry), fout, ensure_ascii=False)
+            fout.write("\n")
+
+
+def main(args):
+    _datasets = []
+    for category in categories:
```
shuoyangd: Not a deal breaker, but if we can load the full dataset without multiple API calls it might make the code cleaner.

nemo_skills/dataset/kormedmcqa/__init__.py
shuoyangd, line 1: FYI there is an existing implementation of this one in lm-evaluation-harness: https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/kormedmcqa

I know they are more designed towards evaluating pre-trained models and not instruction fine-tuned models but let's borrow whatever looks reasonable. Upon quick glance I don't really see any but it'd be great for a Korean speaker to double-check prompt wrapper, etc.

nemo_skills/dataset/kosimpleqa/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# settings that define how evaluation should be done by default (all can be changed from cmdline)
+METRICS_TYPE = "simpleqa"
+GENERATION_ARGS = "++prompt_config=generic/default ++eval_type=math"
+
+JUDGE_PIPELINE_ARGS = {
+    "model": "o3-mini-2025-01-31",
```
shuoyangd, line 22: Why are we being very specific about this one? It'd be great to point to a source if you have one.

nemo_skills/dataset/kormedmcqa/prepare.py
```diff
@@ -0,0 +1,88 @@
...
+import json
+from pathlib import Path
+
+from datasets import concatenate_datasets, load_dataset
+from tqdm import tqdm
+
+from nemo_skills.dataset.utils import get_mcq_fields
+
+
+categories = [
```
shuoyangd: Similarly, I think you might be able to retrieve this dynamically.

nemo_skills/dataset/kosimpleqa/prepare.py
```diff
@@ -0,0 +1,68 @@
...
+
+from datasets import load_dataset
+from tqdm import tqdm
+
+# SimpleQA dataset from OpenAI's simple-evals repository
+
+
+def format_entry(entry: dict, idx: int) -> dict:
+    """Format an entry to match Nemo-Skills format."""
+    return {
```
shuoyangd, line 28: Is it not advisable to have some breakdown for this one (`subset_for_metrics`)? I think `topic` could be a candidate. But again up to you.

nemo_skills/evaluation/evaluator/koifeval.py
```diff
@@ -0,0 +1,1412 @@
...
+_NUM_WORDS_UPPER_LIMIT = 500
+
+
+class Instruction:
+    """An instruction template."""
+
+    def __init__(self, instruction_id):
+        self.id = instruction_id
+
+    def build_description(self, **kwargs):
```
shuoyangd: Are we going to need those?

nemo_skills/evaluation/evaluator/koifeval.py
shuoyangd, line 1: I see this is quite a big script and can potentially be re-used in other codebases for evaluating koifeval. A couple of things I'm wondering:

(1) Is there another script/toolkit in the research community that does this? If so, we should re-use.
(2) If not, is this the right repo to keep this module? We don't really have the expertise to check or maintain this, so in the long run I think the community would be better off if you keep this in a separate repo, and we add you as a dependency.

Note, for example, for English [ifeval](https://github.com/NVIDIA-NeMo/Skills/blob/main/nemo_skills/evaluation/evaluator/ifeval.py) we simply use google's evaluation tool.

What do you think?

shuoyangd (commented):
Thanks again for working on this. I made an initial pass and added a few points for modification or discussion. Please take a look and let me know what you think.

I also need to manually run the CPU test for you. Will report back with another review if something is breaking.

nemo_skills/evaluation/evaluator/koifeval.py
shuoyangd: Also, this python file is missing a copyright header

shuoyangd:
Additionally, can you read [CONTRIBUTING.md](https://github.com/NVIDIA-NeMo/Skills/blob/main/CONTRIBUTING.md#pre-commit-hooks) and make sure you (1) run lint and format (2) sign your commits? Thanks

nemo_skills/dataset/kmmlu-pro/prepare.py
bzantium: Agreed. Removed the per-sample `_EXTRACT_REGEX` (and the accompanying `extract_from_boxed`/`relaxed` overrides) from all four MCQ files (`kmmlu`, `kmmlu-pro`, `kmmlu-redux`, `kormedmcqa`) and added a Korean `정답:` fallback in `mcq.py` mirroring the existing English `Answer:` fallback right above it. Now the Korean MCQs use the same global extraction chain as MMLU/GPQA: `\boxed{}`, "The final answer is", "Answer:", "정답:". Generations that don't follow the prompt's `정답: X` format are correctly counted as wrong rather than rescued by a greedy single-letter fallback.

nemo_skills/dataset/kmmlu-pro/prepare.py
bzantium: Switched to `subject`. The HF schema has both: `subject` is the academic field (e.g. "민법" / Civil Law) and `license_name` is the licensing certification name (e.g. "감정평가사" / appraiser). `subject` is the more semantic axis and matches how MMLU/MMLU-Pro break down by academic subject, which makes it easier to compare cross-benchmark.

nemo_skills/dataset/kmmlu-redux/prepare.py
bzantium: kmmlu-redux already uses `entry["category"]`. That's the only field the KMMLU-Redux HF schema exposes; there is no `license_name` here. The comment may have been intended for kmmlu-pro, where I switched from `license_name` to `subject` per your other thread.

nemo_skills/dataset/kmmlu/prepare.py
bzantium: Done. Dropped the hardcoded 45-item list. Combined with the next thread, replaced both the `categories` constant and the per-config loop with a single `data_files={args.split: f"data/*-{args.split}.csv"}` glob load. Verified end to end: still 35,030 rows with the `Category` column intact on every row. Kept the `supercategories` dict (HUMSS / STEM / Applied Science / Other) since that grouping comes from the KMMLU paper (Son et al. 2024) and isn't in the HF metadata.

nemo_skills/dataset/kmmlu/prepare.py
bzantium: Done. `data_files={args.split: f"data/*-{args.split}.csv"}` collapses the 45 per-config calls into one. Verified: 35,030 rows, same schema, prepare.py runs roughly 3x faster.

nemo_skills/dataset/kormedmcqa/__init__.py
bzantium: Took a look at [`EleutherAI/lm-evaluation-harness/lm_eval/tasks/kormedmcqa`](https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/kormedmcqa). Same dataset (`sean0042/KorMedMCQA`), same 5-choice schema, same 1→A / 2→B / ... / 5→E answer mapping.

The remaining differences are intentional and align with how nemo-skills evaluates all other MCQ benchmarks (MMLU, GPQA, KMMLU, etc.):

* **Eval paradigm.** lm-eval uses `generate_until` with `정답：` as a fixed continuation prompt, designed for base/pre-trained models. We use free-form generation followed by answer extraction in the shared MCQ evaluator, which is designed for instruction-tuned models and matches the existing English MCQ pipeline.
* **Subject organization.** They expose 4 separate tasks (`kormedmcqa_doctor` / `_nurse` / `_pharm` / `_dentist`) aggregated as a `kormedmcqa` group. We load all 4 subjects into a single dataset and use `subset_for_metrics: subject` for the breakdown. Same surface, fewer task definitions.
* **Question template.** Their `A. {A}` vs our `A) {A}` differs only because `get_mcq_fields` (`nemo_skills/dataset/utils.py`) is shared across every MCQ benchmark in this repo. Diverging the format for Korean only would break that consistency.
* **Metric.** They use `exact_match` (case-insensitive) on the predicted letter; we use `symbolic_correct` after regex-extracting the letter. Functionally equivalent for this benchmark.

Asked a Korean speaker on the team to spot-check the prompt wrapper. The `'정답: A/B/C/D/E'` instruction in `prompt/config/eval/korean/mcq-5choices.yaml` reads naturally and matches what an instruction-tuned model is expected to follow. Nothing else from lm-eval seemed worth borrowing without breaking nemo-skills' MCQ conventions, but happy to revisit if you spot something specific.

nemo_skills/dataset/kosimpleqa/__init__.py
bzantium: This config is byte-identical to the English `nemo_skills/dataset/simpleqa/__init__.py`: same model, same server, same `JUDGE_ARGS`. I added a comment in `kosimpleqa/__init__.py` pointing at the English config and noting the rationale (consistency with SimpleQA reports). All values remain CLI-overridable for users who want to swap in a locally hosted judge.

nemo_skills/dataset/kormedmcqa/prepare.py
bzantium: Done. Dropped the hardcoded `["dentist", "doctor", "nurse", "pharm"]` list and switched to `data_files={args.split: f"*/{args.split}-*.parquet"}`, which globs all per-subject parquet files in one call. Verified: 3,009 rows, same `subject` distribution (nurse 878, dentist 811, pharm 614, doctor 435, pharmacy 271). The `pharmacy` rows live inside the `pharm` HF config.

nemo_skills/dataset/kosimpleqa/prepare.py
bzantium: Added `subset_for_metrics: entry["topic"]`. The dataset has 10 well-distributed topics: science & technology (196), politics (163), art (129), other (106), geography (98), sports (87), music (80), tv show (68), history (41), games (32). This diverges slightly from English SimpleQA (which doesn't set `subset_for_metrics` and stores topic only in `metadata`), but the breakdown is useful and the aggregate metric is unchanged.

nemo_skills/evaluation/evaluator/koifeval.py
bzantium: Resolved together with the "separate repo" thread. The entire 1,412-line file is replaced by a 60-line subprocess wrapper. The constants and English-only checkers (`CapitalLettersEnglishChecker`, `LowercaseLettersEnglishChecker`, `CapitalWordFrequencyChecker`) now live in [bzantium/koifeval](https://github.com/bzantium/koifeval), where I audited which `Instruction` subclasses IFEval-Ko's 342 samples actually exercise and trimmed the rest.

nemo_skills/evaluation/evaluator/koifeval.py
bzantium: Agreed. Extracted to https://github.com/bzantium/koifeval (Apache-2.0, derived from `google-research/instruction_following_eval` and `allganize/IFEval-Ko`, with attribution in `LICENSE` and `NOTICE`). The package is pip-installable. This PR now ships:

* `nemo_skills/evaluation/evaluator/koifeval.py` reduced from 1,412 lines to a 60-line subprocess wrapper that mirrors `ifeval.py` byte-for-byte except for the cmd line.
* A `pip install` line in `dockerfiles/Dockerfile.nemo-skills` pinned to commit `7a27889c0cf285e60b1a7b4795211846dc328aaf`.

Validated parity with the original implementation on the existing Qwen3-4B IFEval-Ko output (342 samples): all 342 `strict_eval` and `loose_eval` fields match byte for byte; prompt_strict=70.18, prompt_loose=73.39 unchanged. Also re-ran a fresh end to end eval with Qwen3-1.7B (50 samples, `enable_thinking=false`): strict=50.0%, loose=56.0% with the wrapper invoking the external package successfully.

bzantium: Added. The new 60-line wrapper carries the standard NVIDIA Apache-2.0 header. The external `bzantium/koifeval` package is Apache-2.0 with attribution to Google Research's original IFEval and `allganize/IFEval-Ko` in `LICENSE` and `NOTICE`.

bzantium:
Done. Ran `pre-commit run` over all changed files (ruff check, ruff format, end-of-file fixer, trailing whitespace, etc.) and rebased so all commits on this branch carry a DCO sign-off.

nemo_skills/dataset/kmmlu-redux/prepare.py
shuoyangd: I see `license_name` here: https://huggingface.co/datasets/LGAI-EXAONE/KMMLU-Redux

nemo_skills/dataset/kmmlu-pro/prepare.py
shuoyangd: Good improvement, but please keep it in the per-sample extract_regex :) We would like to keep `mcq.py` English centric and avoid bloating it with multilingual patterns.

nemo_skills/dataset/kmmlu-redux/prepare.py
bzantium: You're right, both fields are present. Looking at the values: `license_name` is the specific licensing certification (e.g. 조경기사) while `category` is the broader academic / industrial field that the certification covers (e.g. 건설, Statistics). For a per-subject breakdown across models the academic field is the more meaningful axis — a model strong in Statistics should look strong in Statistics regardless of which specific certification exam the question came from. So I went with `category` here.

nemo_skills/dataset/kmmlu-pro/prepare.py
bzantium: Reverted the mcq.py change and moved the Korean pattern back to a per-sample `extract_regex` in the four MCQ prepare.py files. The pattern is a single tight match for the prompt's `정답: A/B/C/D[/E]` format with `extract_from_boxed=False, relaxed=False`, so there's no greedy single-letter fallback and no boxed fallback. The English `Answer:` global fallback in mcq.py stays untouched.

dockerfiles/Dockerfile.nemo-skills
```diff
@@ -63,6 +53,9 @@ RUN pip install langdetect absl-py immutabledict nltk ipython && \
     python -c "import nltk; from spacy.cli import download; nltk.download('punkt'); nltk.download('punkt_tab'); \
     nltk.download('stopwords'); nltk.download('averaged_perceptron_tagger_eng'); download('en_core_web_sm')"
 
+# koifeval (Korean adaptation of google-research/instruction_following_eval)
+RUN pip install --no-cache-dir "koifeval @ git+https://github.com/bzantium/koifeval.git@7a27889c0cf285e60b1a7b4795211846dc328aaf"
```
shuoyangd: Can you take a look at the README [here](https://github.com/NVIDIA-NeMo/Skills/tree/main/core) and add that dependency accordingly? This is not the right place to add it

shuoyangd:
CPU test pass locally.
[full-cpu-ci-20260513-151439.log](https://github.com/user-attachments/files/27734748/full-cpu-ci-20260513-151439.log)

dockerfiles/Dockerfile.nemo-skills
bzantium: Moved the install to `core/requirements.txt` and removed the line from `Dockerfile.nemo-skills`!

Kipok (commented):
Can we expose this in the documentation @bzantium ?

Kipok:
@shuoyangd could you also run gpu tests locally? Or feel free to recreate from a branch on your side this PR and then we can run a ci properly

shuoyangd:
@Kipok Trying to do GPU test too. Having some troubles. If it doesn't work I may create a mirror branch & PR solely for running CI. In the longer term we should probably make this more friendly to external contributors.

@bzantium FYI [this](https://github.com/NVIDIA-NeMo/Skills/blob/main/docs/evaluation/multilingual.md) is a good place to add some docs on this. It will show up [here](https://nvidia-nemo.github.io/Skills/evaluation/multilingual/).

shuoyangd:
GPU tests pass from a mirror PR #1449 

One small thing I noticed in the log -- can we use more intelligible names for things like `kmmlu-pro-민법` and `kmmlu-redux-건설`? Having read the code, my guess is that those are fall-back patterns for answer extraction, but for a non-Korean-speaking person it's hard to figure out what those are.

Once the lingering comments are addressed I think we can override the test requirement and merge this. I'm not sure how to override so we may merge through #1449 instead. You can keep pushing to your branch and I'll make sure everything is mirrored before we merge.

@naymaraq do you want to review again? You requested some changes and they are still blocking.

bzantium:
@Kipok @shuoyangd  Added per-benchmark docs to `docs/evaluation/multilingual.md`. Each of the six benchmarks now has a short description, a reference-numbers table (Nemotron-Cascade-2-30B-A3B, kanana-2-30b-a3b-thinking-2601, gemma-4-26B-A4B-it), and a reproduction command. Let me know if anything needs reshaping.

bzantium:
@shuoyangd  Those are not regex patterns. They are the per-subject (`subject` in kmmlu-pro) and per-category (`category` in kmmlu-redux) field from the upstream HF dataset, passed through as `subset_for_metrics` so the metrics output breaks down by topic. The actual answer extraction is a single per-sample regex matching `정답: X`, set in `prepare.py`.

I can translate them to English in `prepare.py` via a Korean to English mapping. For kmmlu-redux that is 14 categories (e.g., 안전관리 → Safety Management, 건설 → Construction); for kmmlu-pro it is 61 subjects (e.g., 민법 → Civil Law, 회계학 → Accounting). Want me to push that?

eagle705:
> My colleague @eagle705 will help assess the quality of those benchmarks. I'll follow-up once I've taken a closer look at the code.

@shuoyangd 
These Korean benchmarks are commonly used for evaluating Korean LLMs. (Typically, KMMLU-*, KoBALT, and CLIcK are used.)
I also added KMMLU/KMMLU-Pro and CLIcK for internally evaluating checkpoints.

>I can translate them to English in prepare.py via a Korean to English mapping. For kmmlu-redux that is 14 categories (e.g., 안전관리 → Safety Management, 건설 → Construction); for kmmlu-pro it is 61 subjects (e.g., 민법 → Civil Law, 회계학 → Accounting). Want me to push that?

@bzantium I think that would improve readability.

bzantium:
Quick update on this thread.

Earlier in this PR I argued for `subject` over `license_name` for KMMLU-Pro on cross-benchmark grounds (MMLU-style academic-field breakdown). After pulling up the KMMLU-Pro and KMMLU-Redux paper (arXiv:2507.08924) I switched to license-level grouping instead. The reasoning:

- The paper itself reports KMMLU-Pro at the license level (14 KNPLs), and the headline metric in the paper is the number of licenses a model can pass. The 61 individual subjects are sub-exams within a license. A test taker who wants to become a Certified Public Accountant has to clear *all* subjects in that license together, so the license is the natural atomic unit here, more so than for general academic MMLU.
- The paper publishes exact English labels for the 14 licenses ("Names of KNPLs (EN)" table), so the metrics output now uses paper-blessed names instead of my own translations of 61 subjects.
- 14 buckets in the per-subset table reads more cleanly than 61, which addresses the intelligibility concern from earlier in this thread.

Trade-off: the per-subject (61) view is gone. If anyone wants it back for MMLU-style cross-benchmark comparison, it's a one-line change in `prepare.py` (swap `entry["license_name"]` for `entry["subject"]`).

For **KMMLU-Redux**: kept the 14-category split (same as before), updated the English names to match the paper's "Category (EN)" table exactly — `Construction`, `Mechanical Engineering`, `Environment & Energy`, etc. (paper-exact, including the `&` separators).

Pushed in f95d756f. Doc strings in `multilingual.md` updated as well.

naymaraq (approved):

shuoyangd:
@bzantium When you are done with you changes, can you merge in new changes from main and give me a ping? I'm going to mirror your branch to #1449  and re-run tests after those changes. If they all checkout I'll go ahead and merge

bzantium:
@shuoyangd Merged latest `main` into the branch (no conflicts), pushed in a5590c60. Ready for the mirror + re-run on #1449 whenever you are.

shuoyangd:
Thank you @bzantium for your contribution! Looking forward to continuing to work with you on the other PRs :)

PR #1376, opened 2026-04-17
Add universal lock files for sandbox

activatedgeek:
## Summary by CodeRabbit

* **Chores**
  * Simplified sandbox build to use a single, consistent dependency lock and unconditional install during image creation.
  * Restricted certain optional Python packages to x86_64 systems via platform-specific install markers.
  * Updated guidance for regenerating dependency locks to ensure consistent, reproducible environment builds.


activatedgeek:
Multi-platform build with the same lockfile succeeds.

PR #1380, opened 2026-04-17
Multimodel profiling instead of difficulty estimation

tatevik-t:
Replaces the single-model `difficulty_estimation` stage in the
OpenScienceReasoning SDG pipeline with a multi-model `profiling` stage that
runs per-model `generate → judge → aggregate` chains in parallel and merges
the results into a single `profiling` array per problem.

Each problem now carries per-model pass-rate metrics instead of a single
scalar, making it straightforward to filter on agreement/disagreement
across a panel of reference models.

## Output schema

Per-row field produced by the new stage:

```json
"profiling": [
  {"model": "qwen3-30b-a3b",   "pass_rate": 0.5, "pass_at_n": "2/4"},
  {"model": "nemotron-super",  "pass_rate": 0.8, "pass_at_n": "4/5"},
  {"model": "gemma-4-31b-it",  "pass_rate": 0.3, "pass_at_n": "1/4"}
]
```
filter_solutions accepts a corresponding dict bound:

```
profiling_pass_rate_range:
  qwen3-30b-a3b:  [0.0, 0.9]
  nemotron-super: [0.1, 1.0]
```
(bounds are min-exclusive, max-inclusive — consistent with the existing
generation_model_pass_rate_range.)


Kipok:
@tatevik-t @rimashahbazyan what's the status of this pr? Do we expect to be able to merge it soon?

PR #1381, opened 2026-04-17
Use upstream NeMo-RL entrypoints and simplify sandbox integration

wedu-nvidia:
## Summary by CodeRabbit

* **New Features**
  * Added config override and container image options for GRPO and SFT pipelines
  * Added Ray sandbox startup templates for multi-node cluster bring-up and optional sandboxed execution

* **Refactor**
  * Switched from local build to a pinned upstream runtime image
  * Training pipelines now invoke upstream training entrypoints and accept configurable training configs
  * Removed legacy on-host training entrypoints and local SFT config files
  * Improved sandbox behavior for Ray-enabled heterogeneous jobs

* **Bug Fixes**
  * SLURM executor now fails the job if any task exits non-zero (stricter failure handling)

* **Documentation**
  * Cleared references to deprecated script paths


wedu-nvidia:
I updated the code to support local ray template for sandbox

dockerfiles/Dockerfile.nemo-rl
```diff
@@ -1,149 +1,21 @@
...
-# Store outside /opt/NeMo-RL to avoid being overwritten by user mounts
-RUN python tools/generate_fingerprint.py > /opt/nemo_rl_container_fingerprint
+# Keep a compatibility symlink for paths that may still reference the older
+# mixed-case location used by previous local images.
+RUN if [ ! -e /opt/NeMo-RL ] && [ -e /opt/nemo-rl ]; then ln -s /opt/nemo-rl /opt/NeMo-RL; fi
 
-# NOTICES.txt file points to where the OSS source code is archived
-RUN echo "This distribution includes open source which is archived at the following URL: https://opensource.nvidia.com/oss/teams/nvidia/nemo-rl/${RC_DATE}:linux-${TARGETARCH}/index.html" > NOTICES.txt && \
-    echo "For further inquiries or assistance, contact us at oss-requests@nvidia.com" >> NOTICES.txt
+WORKDIR /opt/NeMo-Skills
```
Kipok: do we need nemo skills to be installed? Ideally we just use the corresponding nightly in cluster configs and not even have this dockerfile

nemo_skills/pipeline/nemo_rl/grpo.py
```diff
@@ -319,7 +333,14 @@ def grpo_nemo_rl(
         help="If specified, will reuse the code from this experiment. "
         "Can provide an experiment name or an experiment object if running from code.",
     ),
+    config: str = typer.Option(None, help="Override training config YAML; defaults to the upstream container config"),
+    container: str = typer.Option(None, help="Override container image for NeMo-RL training/conversion jobs"),
     with_sandbox: bool = typer.Option(False, help="If True, will start a sandbox container alongside this job"),
+    sandbox_container: str = typer.Option(None, help="Override container image for the sandbox sidecar"),
+    ray_template: str = typer.Option(
+        None,
+        help="Override the Ray template for training tasks. Can be a template name or a local .j2 path. Only applies to Ray-backed jobs; local executor ignores it.",
```
Kipok:
```suggestion
        help="Override the Ray template for training tasks. Can be a template name or a local .j2 path. Only applies to slurm-based jobs; local executor ignores it.",
```

nemo_skills/pipeline/nemo_rl/grpo.py
```diff
@@ -392,15 +412,23 @@ def grpo_nemo_rl(
...
+            raise ValueError("training_data is required for NeMo-Gym GRPO training")
+        if validation_data is None:
+            raise ValueError("validation_data is required for NeMo-Gym GRPO training")
+        if training_data.startswith("/"):  # could ask to download from HF
+            training_data = get_mounted_path(cluster_config, training_data)
+        if validation_data.startswith("/"):
+            validation_data = get_mounted_path(cluster_config, validation_data)
     if run_conversion_only:
         dependent_jobs = -1
     if dependent_jobs > 0:
```
Kipok:
```suggestion
    if dependent_jobs >= 0:
```

nemo_skills/training/nemo_rl/configs/sft.yaml
```diff
@@ -2,118 +2,123 @@
 sft:
```
Kipok: is this not available in nemo-rl directly?

nemo_skills/pipeline/utils/exp.py
```diff
@@ -531,23 +530,42 @@ def add_task(
...
+    server_needs_gpus = server_config is not None and int(server_config.get("num_gpus", 0)) > 0
+    client_num_gpus = num_gpus or 0
+    # For ray heterogenous jobs, nemo-run assumes the first het group is the main task.
+    # So we send the server last if the job needs gpus.
+    server_goes_first = server_needs_gpus and not client_num_gpus
+
     het_group = 0
     het_group_indices = []
-    total_het_groups = (n_servers if server_config is not None else 0) + bool(cmd) + with_sandbox
+    het_group_num_nodes = {}
```
Kipok: not fully sure i understand this change, @gwarmstrong could you also have a look?

nemo_skills/training/nemo_rl/configs/sft.yaml
wedu-nvidia: The nano-v3 yaml has not merged, and I can put it for reference, I can remove it now or after nemo-rl pr merged.

dockerfiles/Dockerfile.nemo-rl
wedu-nvidia: I think we do not need it, I put it in case later we might need nemo-skills for this. 
I think we can remove the dockerfile completely now

nemo_skills/pipeline/nemo_rl/grpo.py
wedu-nvidia: will do

wedu-nvidia:
@Kipok 
I removed below
- the local `Dockerfile.nemo-rl`
- the local `nemo_rl/configs/sft.yaml`
- SFT-specific template override exposure

nemo_skills/pipeline/nemo_rl/grpo.py
```diff
@@ -319,7 +333,14 @@ def grpo_nemo_rl(
         help="If specified, will reuse the code from this experiment. "
         "Can provide an experiment name or an experiment object if running from code.",
     ),
+    config: str = typer.Option(None, help="Override training config YAML; defaults to the upstream container config"),
```
Kipok:
```suggestion
    config: str = typer.Option(None, help="Override training config YAML; defaults to the upstream container config for nano-v3"),
```

Kipok (dismissed):
lgtm, as long as tests pass. Did we launch slurm test (the omr one) here?

cluster_configs/example-local.yaml
```diff
@@ -23,7 +23,7 @@ containers:
   sandbox: dockerfile:dockerfiles/Dockerfile.sandbox
   nemo-skills: dockerfile:dockerfiles/Dockerfile.nemo-skills
   verl: dockerfile:dockerfiles/Dockerfile.verl
-  nemo-rl: dockerfile:dockerfiles/Dockerfile.nemo-rl
+  nemo-rl: nvcr.io/nvidian/nemo-rl:733c7d6-48669320
```
Kipok: is this internal container? We ideally want something publicly available here

nemo_skills/pipeline/nemo_rl/ray_templates/nemo_skills_sandbox_ray.sub.j2
```diff
@@ -0,0 +1,502 @@
+#!/bin/bash
```
Kipok, line 1: why do we need two templates? If one was for lax cluster, we can probably not port it as it's not needed anymore

wedu-nvidia: It is for lax, I keep it here in case one day we need for lax types. But I can delete it now

cluster_configs/example-local.yaml
wedu-nvidia: It is for nightly container published internally, but not sure if public ones are up to date, I will have a check

wedu-nvidia:
> lgtm, as long as tests pass. Did we launch slurm test (the omr one) here?

not yet, I can test for this.

cluster_configs/example-local.yaml
wedu-nvidia: @Kipok I contacted the team and was told public container is released for every 2 months roughly, it seems too old then?

Kipok: I guess it's ok to keep it as private as it has a commit in the name. Can you update the readme in dockerfiles folder to mention that it defaults to private container accessible to nvidia employees, but for anyone else they can build it from corresponding dockerfile / commit in nemo-rl repo? But one thing we should make sure is that our ci still works, so can you check what key / setting we need to set there and sync with me offline so that I can update our repo settings and we can test that gpu / gitlab ci passes

wedu-nvidia:
> lgtm, as long as tests pass. Did we launch slurm test (the omr one) here?

sorry for late reply here, it seems omr cannot pass. I checked the sft loss curve which looks good to me, but eval results is bad, I will investigate this.

wedu-nvidia:
@Kipok I upgraded NeMo-RL to v0.6.0 and verified both SFT and GRPO locally and on the cluster. I also re-enabled the GPU test workflow for this PR. If all checks pass, I think this should be ready to merge.

nemo_skills/pipeline/nemo_rl/ray_templates/nemo_skills_sandbox_ray.sub.j2
Kipok: should we remove this?

Kipok: or this one is needed?

nemo_skills/pipeline/nemo_rl/sft.py
```diff
@@ -86,9 +91,24 @@ def format_train_args(self):
         return cmd
 
     def format_data_args(self):
-        cmd = f"+data.train_data_path={self.prompt_data} "
+        cmd = (
```
Kipok, line 159: why does this part need changes? Some api changed in nemo-rl for sft?

Kipok: or this is just to maintain our previous defaults after removing starting scripts?

tests/gpu-tests/test_train.py
```diff
@@ -139,52 +115,85 @@ def test_grpo_nemo_rl(backend):
...
     # need to clean up current cluster configuration as we mount /tmp and it causes problems
     # need to clean up cache folder as otherwise megatron backend might fail when checkpoint format changes
     docker_rm(["/tmp/ray/ray_current_cluster", "/mnt/datadrive/nemo-skills-test-data/hf-cache/nemo_rl/", output_dir])
 
-    grpo_nemo_rl(
-        ctx=wrap_arguments(
-            "++data.prompt.prompt_config=qwen/math-cot "
-            "++grpo.num_prompts_per_step=2 "
-            "++policy.max_total_sequence_length=256 "
+    common_overrides = (
```
Kipok, line 124: I guess we'd need updates to all sft commands referenced in the docs as well?

nemo_skills/pipeline/nemo_rl/ray_templates/nemo_skills_sandbox_ray.sub.j2
wedu-nvidia: The lax/enroot-specific one can be removed and already removed. This general `nemo_skills_sandbox_ray.sub.j2` is still needed as the opt-in NeMo-Skills Ray sandbox template.

It is not used by default; users only get it when passing `ray_template`. Without it, Ray jobs with `with_sandbox=True` would fall back to the default Run template, which does not launch the sandbox sidecars across the Ray nodes.

tests/gpu-tests/test_train.py
wedu-nvidia: ok, yes, we should and revised the sft command

nemo_skills/pipeline/nemo_rl/sft.py
wedu-nvidia: Yes, this is because we now call upstream `examples/run_sft.py` directly. The old `start_sft.py` wrapper used `data.train_data_path`/`data.val_data_path`, while upstream NeMo-RL SFT uses `data.train.*` and `data.validation.*`. This preserves the old NeMo-Skills defaults for input/output JSONL data after removing the local wrapper.

docs/releases/openmathreasoning/training.md
```diff
@@ -102,10 +102,10 @@ ns nemo_rl sft \
     ++policy.max_total_sequence_length=32768 \
     ++policy.train_micro_batch_size=1 \
     ++policy.train_global_batch_size=1024 \
-    ++policy.tensor_model_parallel_size=1 \
-    ++policy.context_parallel_size=2 \
-    ++policy.lr=3e-4 \
-    ++policy.min_lr=3e-7 \
+    ++policy.megatron_cfg.tensor_model_parallel_size=1 \
```
Kipok, line 112: are you sure this is enough changes? I think there were quite a few parameters in our config that didn't default same way as in official nemo-rl one? So we'd need to explicitly list all of them I guess?

wedu-nvidia: yeah, I need to double check again

wedu-nvidia: I audited the old NeMo-Skills SFT config against the upstream NeMo-RL config. I first listed all differing defaults, but trimmed it down to only the overrides that are required for correctness with the upstream entrypoint: disabling validation when no validation dataset is passed, preserving NeMo-Skills SFT data formatting/chat-template behavior, and explicitly setting sequence packing / divisibility for the long-context CP examples. The optimizer/DDP/checkpoint/logger values are either upstream-owned defaults or recipe hyperparameters, so I avoided duplicating the whole YAML in docs.

docs/releases/openmathreasoning/training.md
```diff
@@ -102,13 +104,22 @@ ns nemo_rl sft \
...
+    ++policy.megatron_cfg.optimizer.lr=3e-4 \
+    ++policy.megatron_cfg.optimizer.min_lr=3e-7 \
     ++policy.megatron_cfg.scheduler.lr_warmup_iters=3000 \
     ++policy.megatron_cfg.scheduler.lr_warmup_init=0 \
+    ++data.add_bos=false \
+    ++data.add_eos=false \
+    ++data.add_generation_prompt=false \
+    ++data.num_workers=10 \
     ++checkpointing.save_period=7500 \
+    ++sft.val_period=0 \
```
Kipok, line 130: I think still not everything is ported. E.g. checkpointing.keep_top_k should be 50 while default in nemo-rl is 3? Can we do some kind of complete check here on all parameters programmatically or something like that?

wedu-nvidia: I just added a programmatic audit comparing the old NeMo-Skills SFT defaults against upstream NeMo-RL v0.6 defaults; whenever a training default differed, I preserved the old config’s value in the wrapper and updated the release reproduction docs to explicitly set those mismatched Megatron parameters when they were not already overridden.

Kipok (approved):

PR #1383, opened 2026-04-17
Expose frequency and presence penalty in inference configs for use with ns eval

shuoyangd:
Those parameters are necessary for Qwen3.5 inference. #1329 added them for underlying functions that build vllm/megatron requests, but they are not exposed in the inference config for commands like `ns eval`. This PR addresses that.

## Summary by CodeRabbit

* **New Features**
  - Added `frequency_penalty` and `presence_penalty` parameters to inference configuration, enabling fine-tuned control over model token generation behavior. Users can now reduce repetitive outputs and improve response diversity when supported by the underlying model.


Kipok:
@shuoyangd why is this a draft? Do you want to merge this?

shuoyangd:
I found a solution to do this without exposing those arguments so let's close it.

PR #1389, opened 2026-04-21
update slurm test for nano v3

wedu-nvidia:
## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end SLURM evaluation workflow for Nano-30B covering no-tools, with-tools, formal-math, and agentic modes.
  * Added a results validation CLI that aggregates metrics, normalizes percent/fraction values, and enforces metric bounds.
  * Added checks for tool usage patterns and detection of timeout regressions across evaluation outputs.

### DIFF
--- a/tests/slurm-tests/nano_30b_eval/check_results.py
+++ b/tests/slurm-tests/nano_30b_eval/check_results.py
@@ -33,7 +33,7 @@ NO_TOOLS_METRICS = {
 }
 
 WITH_TOOLS_METRICS = {
-    "aime25": ("pass@1[avg-of-4]", "symbolic_correct", (96.0, 100.0)),
+    "aime25": ("pass@1[avg-of-4]", "symbolic_correct", (88.0, 100.0)),
     "gpqa": ("pass@1[avg-of-4]", "symbolic_correct", (72.0, 78.0)),
     "hle": ("pass@1", "judge_correct", (13.0, 19.0)),
 }
@@ -64,13 +64,17 @@ def load_metrics_block(metrics_path: Path, benchmark: str):
     return data[benchmark]
 
 
+def normalize_percent(value: float) -> float:
+    return value * 100.0 if 0.0 <= value <= 1.0 else value
+
+
 def check_metric_group(eval_dir: Path, metric_config: dict[str, tuple[str, str, tuple[float, float]]]):
     for benchmark, (agg_key, field, (lo, hi)) in metric_config.items():
         metrics_path = eval_dir / "eval-results" / benchmark / "metrics.json"
         metrics = load_metrics_block(metrics_path, benchmark)
         soft_assert(agg_key in metrics, f"Missing aggregation key {agg_key} in {metrics_path}")
         soft_assert(field in metrics[agg_key], f"Missing field {field} in {metrics_path}")
-        value = float(metrics[agg_key][field])
+        value = normalize_percent(float(metrics[agg_key][field]))
         print(f"{eval_dir.name}/{benchmark}/{agg_key}/{field}: {value}")
         soft_assert(lo <= value <= hi, f"{benchmark}: {field}={value} out of range [{lo}, {hi}]")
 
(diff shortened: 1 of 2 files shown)


wedu-nvidia:
@Kipok @gwarmstrong 
I created a SLURM test suite for Nano v3 and included most of the benchmarks listed on the Hugging Face model page: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16

For a few tasks, I could not find corresponding support in NeMo-Skills, such as Terminal Bench and TauBench V2. I also skipped RULER for now because it requires downloading and preparing the data on the cluster first, and it seems that AA-LCR is currently the higher-priority long-context benchmark.

Regarding the range, I do not know it is reasonable for all of benchmarks listed above. 
Feel free to provide any comments.

tests/slurm-tests/nano_30b_eval/check_results.py
```diff
@@ -0,0 +1,256 @@
...
+import sys
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).resolve().parent.parent))  # for utils.py
+from utils import assert_all, get_nested_value, load_json, soft_assert  # noqa: E402
+
+NO_TOOLS_METRICS = {
+    "aime25": ("pass@1[avg-of-4]", "symbolic_correct", (88.0, 94.0)),
+    "gpqa": ("pass@1[avg-of-4]", "symbolic_correct", (69.0, 76.0)),
+    "mmlu-pro": ("pass@1", "symbolic_correct", (74.0, 82.0)),
```
Kipok, line 29: for large datasets like this, I think we can use max_samples=256 or something like this to make it faster

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+
+from nemo_skills.pipeline.cli import eval as run_eval
+from nemo_skills.pipeline.cli import prepare_data, run_cmd, wrap_arguments
+
+MODEL = "nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16"
+MODEL_DIRNAME = MODEL.split("/")[-1]
+JUDGE_MODEL = "openai/gpt-oss-120b"
+JUDGE_MODEL_DIRNAME = JUDGE_MODEL.split("/")[-1]
+REASONING_PARSER_FILENAME = "nano_v3_reasoning_parser.py"
+DEFAULT_SERVER_CONTAINER = (
```
Kipok, line 27: let's not have references to internal infra, this can be specified in your cluster config and by default we just use what's in there

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+    "/lustre/fsw/portfolios/llmservice/users/igitman/super-models/"
+    "vllm-openai-nightly-097eb544e9a22810c9b7a59e586b61627b308362.sqsh"
+)
+
+NO_TOOLS_PARAMS = (
+    "++inference.tokens_to_generate=120000 "
```
Kipok, line 33: instead of setting tokens to generate, let's set max-model-len in server args, going to be more efficient

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+    "++inference.temperature=1.0 "
+    "++inference.top_p=0.95 "
+    "++chat_template_kwargs.enable_thinking=true "
+    "++tool_modules=[nemo_skills.mcp.servers.python_tool::DirectPythonTool] "
+    "++max_tool_calls=100 "
+)
+
+FORMAL_MATH_PARAMS = (
+    "++inference.tokens_to_generate=38912 "
```
Kipok, line 49: this one is an exception if it's using lower tokens to generate, so here we can set it explicitly

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+
+def get_local_judge_model_path(workspace: str) -> str:
+    return f"{workspace}/{JUDGE_MODEL_DIRNAME}"
+
+
+def get_local_model_path(workspace: str) -> str:
+    return f"{workspace}/{MODEL_DIRNAME}"
+
+
+def setup(workspace, cluster, expname_prefix):
```
Kipok, line 82: we shouldn't need this, you can just use model names and it will put them in hf cache

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+    wandb_project,
+    partition,
+    num_jobs,
+    server_gpus,
+    server_container,
+    run_after,
+):
+    output_dir = f"{workspace}/no_tools"
+    server_args = build_server_args(
+        parser_path=f"{workspace}/nano_v3_parser/{REASONING_PARSER_FILENAME}",
```
Kipok, line 113: latest vllm accepts just nemotron3 for parser or something like this, maybe we use that for simplicitly

tests/slurm-tests/nano_30b_eval/run_test.py
```diff
@@ -0,0 +1,608 @@
...
+        enable_tools=False,
+    )
+    expnames = []
+
+    expname = f"{expname_prefix}-no-tools-aime25"
+    run_eval(
```
Kipok, line 119: we should group more benchmarks together for efficiency. Single 8 node server can likely run more in parallel

Kipok (changes requested):

PR #1391, opened 2026-04-21
fix(parse_reasoning): preserve generation when end tag is missing

### LINKED ISSUE #1390: parse_reasoning wipes `generation` when `</think>` tag is missing
## What happens

`parse_reasoning()` in `nemo_skills/utils.py` sets `generation = ""` whenever the end-reasoning tag (default `</think>`) is absent from the sample's generation text. It emits a warning, but the payload is destroyed.

## Why it's a problem

The "end tag absent" condition does not actually imply "no usable answer." There are two legitimate cases where this wipes real data:

1. **Server-side reasoning parser already stripped the tags upstream.** vLLM's `--reasoning-parser` flag (and SGLang's equivalent) returns OpenAI-compatible responses with `choice.message.content` as the clean answer and `choice.message.reasoning_content` as the trace. `nemo_skills/inference/model/base.py:383-396` already picks up `reasoning_content` into a separate key. The clean `content` has no `</think>` in it — so turning on `++parse_reasoning=True` on top of a server-side parser wipes every generation.
2. **`parse_reasoning=True` set by mistake on a non-reasoning model.** Every row becomes `""`. The existing warning is easy to miss in long runs, and downstream metrics silently score zeros.

A third, less common case is truncated reasoning (the model ran out of budget mid-`<think>`). Wiping is no better than keeping the partial text — downstream evaluation will still score it as wrong, but the raw text is at least inspectable for debugging.

## Proposed fix

When the end tag is missing, leave `generation` untouched and rely on `_{generation_key}_finished_thinking = False` as the explicit diagnostic signal downstream code can gate on. That signal is already set today; `nemo_skills/evaluation/metrics/mcq_multilingual_metrics.py` already guards on it, so the fallback logic there is unchanged.

PR: #1391

## Repro

```python
from nemo_skills.utils import parse_reasoning

# Server-side reasoning parser already stripped tags:
sample = {"generation": "The answer is 42."}
parse_reasoning(sample)
print(sample["generation"])  # current: "" — every row wiped
```

bzantium:
Fixes #1390.

## Summary

`parse_reasoning()` used to set `generation = ""` whenever the end-reasoning tag (default `</think>`) was absent from the sample. This silently wipes real data in two legitimate cases:

1. **Server-side reasoning parser already stripped the tags** (vLLM `--reasoning-parser`, SGLang equivalents). `content` is the clean answer, `reasoning_content` holds the trace — but with no `</think>` in `content`, `parse_reasoning` wiped it.
2. **`parse_reasoning=True` set by mistake on a non-reasoning model.** Every row got wiped instead of passing through.

The `_{generation_key}_finished_thinking` flag is already the explicit diagnostic signal downstream code gates on. This PR keeps the original text in the missing-tag branch and lets the flag do its job.

## Changes

- `nemo_skills/utils.py`: in the missing-tag branch, no longer set `generation=""`; update warning text to mention the server-side-parser case. Deduplicates `_full_{generation_key}` assignment by hoisting it above the conditional.
- `nemo_skills/evaluation/metrics/mcq_multilingual_metrics.py`: existing fallback already gates on `_generation_finished_thinking`, so behavior is unchanged — just refresh the now-stale comment.
- `tests/test_utils.py` (new): 7 unit tests covering the end-tag-present branch, end-tag-missing branch, multiple tags, custom `end_reasoning_string`, custom `generation_key`, and the non-string no-op path.

## Release Notes

* **Bug Fixes**
  * Improved text preservation logic to prevent unintended data loss when processing content with specific markers or tags.

* **Tests**
  * Added comprehensive test suite covering multiple scenarios and edge cases for text processing functionality.

* **Documentation**
  * Enhanced inline documentation with clarified comments about text handling behavior.

### DIFF
--- a/nemo_skills/evaluation/metrics/mcq_multilingual_metrics.py
+++ b/nemo_skills/evaluation/metrics/mcq_multilingual_metrics.py
@@ -72,8 +72,9 @@ class MCQMultilingualMetrics(MathMetrics):
                 # so we don't mutate the original prediction dict
                 temp = {"generation": prediction["generation"]}
                 parse_reasoning(temp)
-                # _generation_finished_thinking=False means no </think> tag was found,
-                # in which case parse_reasoning sets generation="" — fall back to original
+                # _generation_finished_thinking=False means no </think> tag was found;
+                # parse_reasoning leaves generation unchanged in that case but we fall back
+                # to the original for clarity.
                 text_for_lang_detection = (
                     temp["generation"] if temp.get("_generation_finished_thinking") else prediction["generation"]
                 )
--- a/nemo_skills/utils.py
+++ b/nemo_skills/utils.py
@@ -50,16 +50,20 @@ def parse_reasoning(sample: dict, generation_key: str = "generation", end_reason
     if not isinstance(sample[generation_key], str):
         return
     sample[f"_{generation_key}_finished_thinking"] = end_reasoning_string in sample[generation_key]
+    sample[f"_full_{generation_key}"] = sample[generation_key]
     if end_reasoning_string in sample[generation_key]:
-        sample[f"_full_{generation_key}"] = sample[generation_key]
         sample[generation_key] = sample[generation_key].split(end_reasoning_string)[-1].strip()
     else:
-        sample[f"_full_{generation_key}"] = sample[generation_key]
-        sample[generation_key] = ""  # no end tag, so setting the generation to empty
+        # End tag missing — either the model never closed its reasoning (truncation) or the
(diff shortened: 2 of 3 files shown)


Kipok:
@bzantium the current behavior is actually by design like this. We want to match what reasoning parser is doing on the server side and if model didn't emit </think> the generation would be empty in that case. So we want to do the same here and indeed score of 0 on benchmarks is expected.

bzantium:
@Kipok understood, thanks for the clarification. Closing this PR and the linked issue.

PR #1392, opened 2026-04-21
update vllm dockerfile for ray

wedu-nvidia:
## Summary by CodeRabbit

* **New Features**
  * Multi-node deployments can enable a distributed execution backend for scaled inference.

* **Chores**
  * Updated container base image tags and added additional runtime packages (including Ray) to image builds.

* **Tests**
  * Added Slurm end-to-end benchmark workflows and a results-check script validating AIME25 metrics across backends.

* **Documentation**
  * Updated referenced Docker image tags in build documentation.


nemo_skills/inference/server/serve_vllm.py
```diff
@@ -27,6 +27,9 @@ def main():
     args, unknown = parser.parse_known_args()
 
     extra_arguments = join(unknown)
+    has_distributed_backend = any(
```
Kipok: we shouldn't need this change I think - the user args will always override the default anyway. So we cna just check for multinode and if that's set we add the argument always

Kipok (commented):
can you run gpu and slurm tests on this?

Kipok:
we could combine with upgrade to sglang and trtllm as well so that we do them all in one batch of tests

nemo_skills/inference/server/serve_vllm.py
wedu-nvidia: sure, we can get rid of this

wedu-nvidia:
> we could combine with upgrade to sglang and trtllm as well so that we do them all in one batch of tests

Why I did not see sglang and trtllm docker files?

you suggest to revise?

_containers = {
    "trtllm": "nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc1",
    "sglang": "lmsysorg/sglang:v0.5.8",
}

Kipok:
yes, as well as all configs

wedu-nvidia:
@Kipok I updated all three image versions, added a Slurm test for the Super model, and verified that all three are working correctly.

tests/slurm-tests/super_120b_aime25/run_test.py
```diff
@@ -0,0 +1,149 @@
...
+    "++chat_template_kwargs.enable_thinking=true "
+    "++inference.tokens_to_generate=131072 "
+    "++inference.temperature=1.0 "
+    "++inference.top_p=0.95 "
+)
+VLLM_SERVER_ARGS = (
+    "--async-scheduling "
+    "--dtype auto "
+    "--kv-cache-dtype fp8 "
+    "--max-model-len 262144 "
```
Kipok:
```suggestion
    "--max-model-len 131072 "
```

tests/slurm-tests/super_120b_aime25/run_test.py
```diff
@@ -0,0 +1,149 @@
...
+# limitations under the License.
+
+import argparse
+
+from nemo_skills.pipeline.cli import eval, prepare_data, run_cmd, wrap_arguments
+
+MODEL = "nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16"
+COMMON_CTX = (
+    "++chat_template_kwargs.enable_thinking=true "
+    "++inference.tokens_to_generate=131072 "
```
Kipok:
```suggestion
```

tests/slurm-tests/super_120b_aime25/run_test.py
```diff
@@ -0,0 +1,149 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
Kipok, line 1: let's add a test for TIR with --enable-prefix-caching (for vllm) and similar options for other backends

Kipok (approved):
if gpu tests pass, should be good to merge

wedu-nvidia:
I added TIR backend support, and we can wait for the GPU tests. By the way, I already verified TIR performance on the cluster.

PR #1393, opened 2026-04-22
vllm: add server_type=vllm_dp_ray for DP + extra Ray nodes

gwarmstrong:
## Summary
- Adds a new `server_type=vllm_dp_ray` that runs vLLM in-process with two monkey-patches to `CoreEngineActorManager` so vLLM DP works when total Ray nodes > `data_parallel_size` (e.g. an extra Ray worker for a neural-eval metric like xCOMET).
- `serve_vllm_dp_ray.py` connects to the existing Ray cluster, pre-reserves DP rank 0's PG pinned to the head node, applies the patches, then runs the OpenAI `api_server` in the same process so the patches stick.
- `get_ray_server_cmd` gains an optional `dp_size` arg: workers with `SLURM_PROCID >= dp_size` start with `--num-gpus=0` plus a custom `extra_gpu` resource and `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1`, hiding their GPUs from vLLM's Ray cluster view while keeping them usable by opt-in tasks. Existing callers (no `dp_size`) are unchanged.
- Wires `vllm_dp_ray` into `SupportedServers{,SelfHosted}` and the HF-model-name whitelist in `get_server_command`.

Temporary until vLLM lands equivalent upstream support.

## Usage: how vLLM nodes vs. extra (Gym) nodes are specified

Partitioning is purely by SLURM rank (`SLURM_PROCID`):

```
SLURM_PROCID = 0                       -> head node: starts Ray head + runs serve_vllm_dp_ray.py in-process.
                                          Reserves DP rank 0's PG pinned to this node (via node:<ip> auto-resource).
SLURM_PROCID in 1..dp_size-1           -> DP worker: `ray start --block --address=$SLURM_MASTER_NODE:6379`
                                          (full GPUs visible to Ray; vLLM schedules additional DP ranks here).
SLURM_PROCID in dp_size..num_nodes-1   -> extra worker: `ray start --num-gpus=0
                                            --resources='{"extra_gpu": 8}'` + RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1
                                          (GPUs hidden from Ray accounting so vLLM ignores the node;
                                          Gym/neural-eval tasks opt in via @ray.remote(num_gpus=0, resources={"extra_gpu": N})).
```

The caller controls the split with two knobs:
- `num_nodes` — total SLURM nodes in the allocation.
- `server_args="... --data-parallel-size=M ..."` — `get_server_command` regex-parses `M` out and passes it to `get_ray_server_cmd` as `dp_size`.

So the contract is: allocate `N` nodes, say `--data-parallel-size=M`, and the first `M` SLURM ranks become vLLM DP nodes, the remaining `N - M` become extra/Gym nodes.

Example — 5-node allocation, 4 DP replicas for vLLM, 1 extra node for a neural-eval Ray task:

```python
nemo_gym_rollouts(
    ...,
    server_type="vllm_dp_ray",
    server_gpus=8,
    server_nodes=5,                                # 5 total nodes
    server_args="--data-parallel-size=4 ...",      # 4 -> vLLM DP, 1 -> extra
    ...,
)
```

The opt-in Ray task would then need to request `resources={"extra_gpu": 8}` (and set its own `CUDA_VISIBLE_DEVICES` if it wants the physical GPUs, since `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1` is set on extra nodes).

Caveats:
- `extra_gpu: 8` is currently hardcoded per node (assumes 8-GPU hosts).
- If `--data-parallel-size` is omitted from `server_args`, `dp_size` defaults to `num_nodes` (i.e. no extra nodes). Pass `--data-parallel-size` explicitly when you want an extra worker.

## Test plan
Shouldn't affect existing behavior (opt-in via new `server_type`), but will run the Nano tool-calling tests (~10 min) to verify no regressions.

## Summary by CodeRabbit

* **New Features**
  * Added a vllm_dp_ray server option that launches vLLM servers coordinated via Ray with an in-process entrypoint and model registry support.

* **Behavior Changes**
  * Server startup now computes serving node counts from topology flags, orchestrates mixed serving/worker roles, and maps vllm_dp_ray/vllm_multimodal to the vLLM container image.

* **Documentation**
  * Added README describing five topology tests and validation rules.

* **Tests**
  * Added wmt24pp_gym_topology test suite, orchestrator, and result checker; updated test run invocation.


nemo_skills/inference/server/serve_vllm_dp_ray.py
```diff
@@ -0,0 +1,378 @@
...
+cluster has more nodes than DP replicas — for example, a 5-node allocation
+running DP=4 with the 5th node reserved for a neural-eval Ray task.
+
+NeMo Gym's ``LocalVLLMModel`` already carries the fix as two in-process
+monkey-patches (``_patch_create_dp_placement_groups`` and
+``_patch_init_data_parallel``). This wrapper ports them so pipelines that
+serve vLLM via ``nemo_skills.inference.server`` (``nemo_gym_rollouts``,
+``generate``, ``eval``, ...) can use the extra-node topology too.
+
+The patches are intended to be temporary — see the upstream vLLM PR
```
Kipok: can you add pr link?

nemo_skills/inference/server/serve_vllm_dp_ray.py
```diff
@@ -0,0 +1,378 @@
...
+    vLLM's ``launcher.serve_http`` installs SIGINT/SIGTERM handlers via
+    ``loop.add_signal_handler`` + ``signal.signal``. Neither is safe when
+    a nested event loop runs off the main thread. We run on main thread
+    here, so this is defensive — only needed if a future caller wraps
+    this entrypoint in a background thread.
+    """
+    signal.signal = lambda *a, **kw: None
+
+
+def _build_vllm_argv(args: argparse.Namespace, extra: Sequence[str]) -> list[str]:
```
Kipok, line 353: would be good to share things like this with the other server_vllm.py, e.g. there is another pr from Wei that changes some arguments in that script, so good to make changes in a single place

Kipok (commented):
I don't fully understand the logic, but one thing that's not clear to me is what happens if we have mutlinode parallelization + dp ? E.g. serving big model with tp=16 (2 nodes) and then adding dp=4 on top to have 8 nodes in total. Will that work? Or just tp=16,pp=2,dp=1 for 4 nodes, does that need extra handling?

nemo_skills/inference/server/serve_vllm_dp_ray.py
gwarmstrong: Done. Added in 9ef608b

gwarmstrong:
> I don't fully understand the logic, but one thing that's not clear to me is what happens if we have mutlinode parallelization + dp ? E.g. serving big model with tp=16 (2 nodes) and then adding dp=4 on top to have 8 nodes in total. Will that work? Or just tp=16,pp=2,dp=1 for 4 nodes, does that need extra handling?

@Kipok good catch. The original scope was fairly narrow--it pretty just worked for a model that runs on 1 node. I have expanded it now so it should generalize across parallel settings--I've tested the following configurations, and added slurm tests for them as well:


| # | TP | DP | replica spans | model nodes | extra     (COMET) nodes | total | model |
|---|---:|---:|---:|---:|---:|---:|---|
| C1 | 8 | 1 | 1 node | 1 | 1 | 2 | Nemotron-3-Nano | 
| C2 | 8 | 4 | 1 node | 4 | 1 | 5 | Nemotr    on-3-Nano | 
| C3 | 16 | 2 | 2 nodes | 4 | 1 | 5 | DeepSeek-V2-Lite | 
| C4 | 16 | 1 | 2 nodes | 2 | 1 | 3 | DeepSeek-V2-Lite | 
| C    5 | 8 | 1 | 1 node | 1 | 2 | 3 | Nemotron-3-Nano |

nemo_skills/inference/server/serve_vllm_dp_ray.py
gwarmstrong: Agree, but I would prefer to push to a follow-up. This PR is purely additive (new `server_type=vllm_dp_ray`, zero change to existing `vllm` usage).

Happy to build some shared utilities as a separate PR.

PR #1394, opened 2026-04-22
fix(pipeline): honor keep_mounts=False to prevent sandbox mount leak

gwarmstrong:
## Summary

`SandboxScript(keep_mounts=False)` is supposed to give the sandbox container zero access to cluster filesystem mounts. In practice, `Pipeline._create_executor` was silently reintroducing them: `exec_config["mounts"] or None` collapsed the empty-list isolation signal, and the subsequent additive merge unconditionally appended cluster mounts even when the sandbox had asked to be isolated.

This PR propagates `keep_mounts` through `execution_config` and branches on it in Stage B. The three bug rows in the matrix below are now fixed; all other rows are preserved.

### Case matrix

| `Command.mounts` | `script.keep_mounts` | Before              | After               |
| ---------------- | -------------------- | ------------------- | ------------------- |
| `None`           | absent (non-sandbox) | cluster mounts      | cluster mounts      |
| `None`           | `True`               | cluster mounts      | cluster mounts      |
| `None`           | `False`              | cluster mounts ❌   | `[]` ✅             |
| `[]`             | absent               | cluster mounts      | cluster mounts      |
| `[]`             | `True`               | cluster mounts      | cluster mounts      |
| `[]`             | `False`              | cluster mounts ❌   | `[]` ✅             |
| `[/a:/b]`        | absent               | cluster + `/a:/b`   | cluster + `/a:/b`   |
| `[/a:/b]`        | `True`               | cluster + `/a:/b`   | cluster + `/a:/b`   |
| `[/a:/b]`        | `False`              | cluster + `/a:/b` ❌ | `[/a:/b]` ✅        |

### User-facing impact

Sandboxes running untrusted generated code no longer expose cluster filesystem mounts (e.g. user homes, dataset caches on `/lustre`). Workflows that were implicitly relying on the leak to reach cluster paths from inside the sandbox will need to either:
- set `keep_mounts_for_sandbox=True` at the CLI layer, or
- pass explicit mounts via `Command.mounts`.

### Changes

- `nemo_skills/pipeline/utils/declarative.py`
  - Stage A (`Command.prepare_for_execution`): compute `keep_mounts` unconditionally and add it to `execution_config`.
  - Stage B (`Pipeline._create_executor`): replace the `[] or None` collapse + unconditional additive merge with a 4-way switch on `(extra_mounts, keep_mounts)`.
- `tests/test_declarative_pipeline.py`
  - New `TestMountsResolution` class: 9 parametrized Stage A cases + 9 end-to-end cases covering the full matrix, with the 3 bug rows explicitly asserted.

## Test plan

- [x] `pytest tests/test_declarative_pipeline.py` — 51 pass (18 new)
- [x] Pre-commit hooks (ruff check + format) pass
- [ ] Smoke-test a sandbox generate job and confirm the container no longer mounts cluster paths when `keep_mounts_for_sandbox=False`

## Release Notes

* **Bug Fixes**
  * Fixed mounts resolution logic in pipeline execution to correctly handle cluster mount inheritance, preventing unintended mounts from being included when explicitly specified or in sandbox scenarios.

* **Tests**
  * Added regression test suite for pipeline mounts resolution, verifying correct mount inheritance and behavior across different execution scenarios.


sgunasekar:
LGTM for fix

sgunasekar (approved):
LGTM

PR #1396, opened 2026-04-23
update-ray/remove cgraph

wedu-nvidia:
## Summary by CodeRabbit

* **Chores**
  * Updated Ray package installation in Docker image configuration for simplified dependencies.


Kipok (approved):
let's merge for now, but if perf issues are found, we should change this. Also we should probably update to 0.20 anyway to support dsv4 @wedu-nvidia

wedu-nvidia:
> let's merge for now, but if perf issues are found, we should change this. Also we should probably update to 0.20 anyway to support dsv4 @wedu-nvidia

Sure, will get it tested tomorrow

PR #1406, opened 2026-04-30
fix: bug where empty strings overwrite user/system messages

naymaraq:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of system and user message overrides in prompt generation.
  * Fixed behavior so explicit empty-string system messages remove any existing system prompt, and explicit user messages apply only when explicitly provided (not just truthy).


Jorjeous (changes requested):
Lets add guard on empty messages list when system_message is non-empty

Other LGTM, will approve when above done

Jorjeous (approved):
fixed

PR #1411, opened 2026-05-01
fix: aggregate arena metrics correctly under task:N

### LINKED ISSUE #1410: ArenaMetrics collapses task:N samples to best-of-N, biasing Elo upward
## Summary

`ArenaMetrics` in `nemo_skills/evaluation/metrics/arena_metrics.py` does not handle `task:N>1` correctly. When N samples are generated per prompt, the metric currently:

1. Selects the best-of-N judgment per (row, direction), favoring the most candidate-friendly outcome.
2. Reports the result under `pass@N`, even though the underlying score is a single Elo / win-rate, not a probability over N attempts.

The class itself acknowledges this in `arena_metrics.py:124-126`:

```python
# TODO: the class should support pass@k, but this forces it to report as pass@1.
#       There is some error here for k>1
self.agg_mode = "pass@1"
```

## Why this matters

Best-of-N picking inflates the candidate's Elo because every row's score is the maximum across N independent draws of the judge. Users running `task:N` to estimate sampling variance instead get a strictly upward-biased point estimate, with no `pass@1[avg-of-N]` reporting and no per-sample variance.

## Proposed fix

Flatten every `(row × sample)` pair into its own battle entry before calling `get_aggregate_score`. This is mathematically equivalent to `pass@1[avg-of-N]` for binary win/loss/tie outcomes, and the bootstrap CI in `get_aggregate_score` naturally absorbs sampling variance. The result should be reported under `pass@1` (with a `pass@1[avg-of-N]` alias when N>1).

## Behavior change

Users currently running `task:N>1` on arena benchmarks will see lower (unbiased) Elo / win-rates than before. `task:1` behavior is unchanged.

## Affected benchmarks

- `arena-hard`
- `arena-hard-v2`
- `m-arena-hard`
- `m-arena-hard-v2`

bzantium:
## Summary

- Fix `ArenaMetrics` so `task:N>1` no longer collapses to best-of-N (biased upward Elo).
- Each `(row, sample)` battle pair now feeds `get_aggregate_score` directly, so the bootstrap CI absorbs sampling variance — equivalent to `pass@1[avg-of-N]` for binary outcomes.
- Result is reported under `pass@1`, with a `pass@1[avg-of-N]` alias when `N>1`. `task:1` behavior is unchanged.

Closes #1410

## Behavior change

Users currently running `task:N>1` on `arena-hard`, `arena-hard-v2`, `m-arena-hard`, or `m-arena-hard-v2` will see lower (unbiased) win-rates than before. The previous best-of-N picking was buggy — the class itself flagged this in the TODO at `arena_metrics.py:124-126`.

## Summary by CodeRabbit

* **New Features**
  * Metrics output unified under a single "pass@1" key while still reporting per-category breakdowns.

* **Bug Fixes**
  * Judgment records now flatten so every prediction sample is counted individually and ordering is preserved.
  * Corrected counting so max_k reflects samples per row and total reflects rows, ensuring accurate aggregation across batches.

### DIFF
--- a/nemo_skills/evaluation/metrics/arena_metrics.py
+++ b/nemo_skills/evaluation/metrics/arena_metrics.py
@@ -43,84 +43,51 @@ class ArenaMetrics(BaseMetrics):
     def update(self, predictions):
         """Updating the evaluation results with the current element.
 
+        Each (row, sample) pair contributes its own judgment pair to ``self.scores``,
+        so the bootstrap Elo in ``get_aggregate_score`` naturally averages over
+        sampling variance — equivalent to ``pass@1[avg-of-N]`` for binary outcomes.
+
         Args:
             predictions (list[dict]): aggregated predictions across all generations.
                 The content of the file is benchmark specific.
         """
-        # this shouldn't do any heavy calculation, but just read the metric from existing json entry
-        # all the heavy lifting should be done in the evaluation script
         super().update(predictions)
-        self.scores.append([])
-        self.agg_mode = f"pass@{len(predictions)}"
-
-        # Track category for per-category scoring (defaults to None for v1 compatibility)
         category = predictions[0].get("category")
-        self.categories.append(category)
-
-        if len(predictions) > 1:
-            judge_scores = [self._get_judge_score(elem["judgement-gen-base"]) for elem in predictions]
-            # adding the best score out of all the generations
-            possible_scores = ["A>>B", "A>B", "A=B", "B>A", "B>>A"]
-            for possible_score in possible_scores:
-                # picking the best available score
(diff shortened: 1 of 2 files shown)


Kipok (commented):
@bzantium would https://github.com/NVIDIA-NeMo/Skills/pull/1429 achieve a simialr purpose? I think that one is a bit cleaner and matches conventions used for other benchmarks, so maybe we merge that one instead?

bzantium:
@Kipok sure! Thanks for taking it into account :)

Kipok:
closing in favor of https://github.com/NVIDIA-NeMo/Skills/pull/1429

PR #1424, opened 2026-05-02
make exclude_domains_config optional in TavilySearchTool

jiacheng-xu:
Tavily won't fail if exclude_domains_config is not provided. 

## Bug Fixes
* Improved error handling for unset configuration options—the tool now gracefully handles missing settings with warnings instead of failing during initialization and execution.


gwarmstrong (commented):
Will not merge. We intentionally want to make it difficult to use the tool with no config.

jiacheng-xu:
Hi @gwarmstrong may I ask what's the reason for this?

PR #1426, opened 2026-05-03
fix bleu score calculation

naymaraq:
Addressing the [issue](https://github.com/NVIDIA-NeMo/Skills/issues/1408) reported by @shuoyangd 

> Sentence-level BLEU was aggregated [here](https://github.com/NVIDIA-NeMo/Skills/blame/6e20e4e4b924288c4d7ec9bd21238b1df37d0bfd/nemo_skills/evaluation/metrics/audio_metrics.py#L311) and used as corpus-level BLEU. This should NEVER be done, because sentence-level BLEU is calculated with aggressive smoothing which is not necessary at corpus-level. Numbers reproduced in this way will be significantly different compared to regular corpus-level BLEU.

## Summary by CodeRabbit

* **Improvements**
  * Audio evaluation now computes corpus BLEU at evaluation time and applies language-aware tokenization.

* **Changes**
  * Standardized dataset export fields to "source" and "reference".
  * Metric aggregation now uses the "reference" field as ground-truth for translation scoring.

* **Removals**
  * Removed a legacy evaluation pipeline configuration constant from several dataset modules.

### DIFF
--- a/nemo_skills/evaluation/metrics/audio_metrics.py
+++ b/nemo_skills/evaluation/metrics/audio_metrics.py
@@ -40,6 +40,42 @@ from nemo_skills.utils import get_logger_name
 LOG = logging.getLogger(get_logger_name(__file__))
 
 
+def compute_corpus_bleu(
+    hyps: list[str],
+    refs: list[str],
+    tgt_langs: list[str | None],
+) -> float:
+    """Compute corpus BLEU using the same tokenizer that sentence-level BLEU used.
+
+    Groups (hyp, ref) pairs by the tokenize resolved from each sample's tgt_lang
+    so ja/zh/ko aren't silently scored under the default 13a tokenizer. With a
+    single language, this is one corpus_bleu call; with mixed languages we
+    weighted-average per-language corpus_bleu scores by sample count.
+    """
+    from sacrebleu import corpus_bleu
+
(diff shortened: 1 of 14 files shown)


nemo_skills/evaluation/metrics/audio_metrics.py
```diff
@@ -221,7 +223,10 @@ def update(self, predictions):
             if "per" in pred and pred["per"] is not None:
                 self.per_scores.append(pred["per"])
             if "bleu" in pred and pred["bleu"] is not None:
-                self.bleu_scores.append(pred["bleu"])
+                self.bleu_hyps.append(pred.get("pred_text", "") or "")
```
shuoyangd: I think if a line is missing those fields (this and the next line) it's better to throw an Error rather than falling back to a null string default. Will this break any existing patterns?

Also, why do you need `or ""` when you already have a `get` with default?

naymaraq: Fixed.

shuoyangd (approved):
Per off-line discussion with @naymaraq , we additionally rolled back the `source_key`/`reference_key` mechanism with COMET judge and use `source` and `reference` keys across the board.

This unifies the keys used by BLEU and COMET and simplifies the code logic.

PR #1428, opened 2026-05-05
deps: bump litellm to 1.83.14 (CVE), pin nemo_run to avoid leptonai/httpx conflict

ushnish-de:
## Summary

Bumps `litellm[caching]` from `1.83.0` → `1.83.14` to address compliance-flagged CVEs (any version below `1.83.7` is flagged). Pins `nemo_run` to a specific upstream commit so the dependency graph is resolvable.

## Why bump to 1.83.14 specifically (not 1.83.7)

`1.83.14` is the only viable version that satisfies all of:

- **Compliance** — must be `>= 1.83.7` to clear the CVE scan.
- **`compute-eval` compatibility** — `compute-eval` requires `python-dotenv >= 1.2.1`. `litellm 1.83.7` pins `python-dotenv == 1.0.1`, which conflicts. `litellm 1.83.14` pins `python-dotenv == 1.2.2`, which satisfies both.

This is the same conflict that caused the original `1.83.7` bump to be reverted ~2 months ago.

## Why pin `nemo_run`

After bumping `litellm` to `1.83.14`, a *second* dependency conflict surfaces:

- `litellm >= 1.83.7` pins `httpx == 0.28.1`.
- `nemo_run @ main` hard-depends on `leptonai >= 0.26.6`, which pins `httpx[http2] == 0.27.2`.
- These are mutually incompatible — even the latest `leptonai 0.27.1` (April 2026) still pins `httpx[http2] == 0.27.2`.

To resolve, `nemo_run` is pinned to commit [`fb86cd02`](https://github.com/NVIDIA-NeMo/Run/commit/fb86cd02366644b5046b47b112e8e9aa7da8f356) — the last commit before [PR #224 (LeptonExecutor support)](https://github.com/NVIDIA-NeMo/Run/pull/224) added `leptonai` as a hard dependency.

This is safe because:

- `nemo-skills` does not import `LeptonExecutor` or `leptonai` (verified via `grep`).
- Our cluster configs do not reference `LeptonExecutor`.
- A future PR can move `nemo_run` forward once upstream either drops `leptonai`, makes it optional, or relaxes its `httpx` pin.

## Changes

| File | Change |
|---|---|
| `core/requirements.txt` | `litellm[caching]==1.83.0` → `litellm[caching]==1.83.14` |
| `requirements/pipeline.txt` | `nemo_run @ git+...` → pinned to `@fb86cd02366644b5046b47b112e8e9aa7da8f356` with explanatory comment |

## Local validation

Local `pip install -e .[dev]` hit `resolution-too-deep` — pip's resolver couldn't navigate the new graph efficiently. CI may hit the same issue.

If CI fails on `resolution-too-deep` rather than on tests, the fix is one of:

1. **Tighten pins in `core/requirements.txt`** (e.g., explicit `httpx==0.28.1`, `python-dotenv==1.2.2`, `jiter>=0.10.0`) — smallest change, no workflow modification needed.
2. **Switch CI's install to `uv pip install`** in `.github/workflows/tests.yml` — uv's resolver handles this graph easily, but requires a workflow change.
3. **Stage the install in phases** in CI.

Happy to follow up with whichever approach is preferred.

## Risk assessment

- ✅ `litellm 1.83.14` API surface used by `nemo-skills` patches (`patch_litellm_logging.py`, `litellm_hybrid_cache.py`, custom subclasses in `inference/model/`) was verified compatible — all internal symbols and method signatures used are unchanged from `1.83.0`.
- ✅ Pinned `nemo_run` commit predates the most recent `nemo_run` features (KubeflowExecutor, RayCluster API improvements) — those are not used by `nemo-skills` today.
- ⚠️ Pin will need to be bumped forward eventually as `nemo_run` evolves; tracked as a future-PR item.

## Summary by CodeRabbit

* **Chores**
  * Updated core runtime dependencies to improved HTTP handling and caching behavior.
  * Pinned an infrastructure component to a specific commit and added explanatory comments documenting the rationale and pinning details.

### DIFF
--- a/requirements/pipeline.txt
+++ b/requirements/pipeline.txt
@@ -3,5 +3,8 @@
 
 click < 8.2.0  # https://github.com/ai-dynamo/dynamo/issues/1039
 nemo-evaluator-launcher<0.1.47
-nemo_run @ git+https://github.com/NVIDIA-NeMo/Run
+# Pinned to the last commit before LeptonExecutor was added (PR #224, 2025-05-14).
+# leptonai (a hard dep of newer nemo_run) pins httpx[http2]==0.27.2, which conflicts
+# with litellm>=1.83.7's httpx==0.28.1 pin. nemo-skills does not use LeptonExecutor.
+nemo_run @ git+https://github.com/NVIDIA-NeMo/Run@fb86cd02366644b5046b47b112e8e9aa7da8f356
 typer >= 0.13
--- a/core/requirements.txt
+++ b/core/requirements.txt
@@ -22,7 +22,7 @@ iso639-lang
 langcodes
 langdetect
 language-data
-litellm[caching]==1.83.0
+litellm[caching]==1.83.14
 math-verify[antlr4_9_3]
 mcp
 numpy
--- a/core/requirements.txt
+++ b/core/requirements.txt
@@ -14,7 +14,7 @@ fire
 flask
 func-timeout
 gradio
-httpx
+httpx==0.28.1                  # pin to match litellm 1.83.14's requirement and avoid resolver thrashing
 huggingface_hub
 hydra-core
 ipython
@@ -25,10 +25,12 @@ language-data
 litellm[caching]==1.83.14
 math-verify[antlr4_9_3]
 mcp
+multidict>=6.0                 # explicit lower bound to stop pip backtracking through ancient 4.x sdists
 numpy
 openai
 openpyxl>=3.1.0
 pandas>=2.0.0
+python-dotenv>=1.2.1           # match compute-eval's requirement explicitly so pip pins it early
 pyxlsb>=1.0.10
 pyyaml
 rank_bm25


requirements/pipeline.txt
```diff
@@ -3,5 +3,8 @@
 
 click < 8.2.0  # https://github.com/ai-dynamo/dynamo/issues/1039
 nemo-evaluator-launcher<0.1.47
-nemo_run @ git+https://github.com/NVIDIA-NeMo/Run
+# Pinned to the last commit before LeptonExecutor was added (PR #224, 2025-05-14).
+# leptonai (a hard dep of newer nemo_run) pins httpx[http2]==0.27.2, which conflicts
+# with litellm>=1.83.7's httpx==0.28.1 pin. nemo-skills does not use LeptonExecutor.
+nemo_run @ git+https://github.com/NVIDIA-NeMo/Run@fb86cd02366644b5046b47b112e8e9aa7da8f356
```
Kipok, line 9: that's really old, right? I'm pretty sure this will break a bunch of code as we won't have newer nemo-run features

PR #1429, opened 2026-05-06
Add pass@1[avg-of-N] support to ArenaMetrics

kslowik:
## Summary

`ArenaMetrics` currently emits only `pass@N` (best-of-N win rate) when `num_repeats > 1`. This PR adds `pass@1[avg-of-N]` (average single-shot win rate) so Arena-Hard runs surface both metrics, matching the convention of sister metric classes (`OmniMetrics`, `MathMetrics`) that emit both via `BaseMetrics._compute_pass_at_k`.

Resolves the TODO in `arena_metrics.py:124` ("the class should support pass@k").

## Why

Arena-Hard scoring is a global Bradley-Terry / Elo win rate via `get_aggregate_score` rather than a per-prediction binary correctness score, so `ArenaMetrics` cannot reuse `BaseMetrics._compute_pass_at_k` directly. The existing `update()` picked the best-of-N pair per prompt eagerly and discarded the other N-1 generations, which made `pass@1[avg-of-N]` impossible to compute downstream.

## How

- `update()` now stores **all** per-prediction `(judgement-gen-base, judgement-base-gen)` score pairs in `self.per_prompt_scores` (no eager selection, no `N==1` vs `N>1` branching).
- `get_metrics()` derives both metrics from the same stored data:
  - **pass@N** (best-of-N): pick the most candidate-favorable label per direction across all N predictions for each prompt, then run Elo bootstrap on the resulting one-pair-per-prompt list. Identical numerical behavior to before.
  - **pass@1[avg-of-N]**: run Elo bootstrap separately on each repeat's predictions (N independent bootstraps over the same 500-prompt set), then average the resulting win rates. Skipped for `N==1` (degenerate with pass@1).
- Per-category breakdown (arena-hard-v2) computed for both metrics via the same `_aggregate` helper.
- Numpy types cast to native Python before return so `yaml.safe_dump` works downstream.
- `evaluations_to_print` overridden to drop the inherited `majority@N` request that Arena never emitted (matches `OmniMetrics` convention).
- Drops dead state: `self.lengths` (never read by anything), `self.agg_mode` (no longer needed).

## Behavior preserved

- `pass@N` numerical output unchanged for any input.
- `N==1` case (single prediction per prompt) still emits exactly `pass@1` and nothing else.
- Per-category breakdown semantics unchanged.
- `get_incorrect_sample`, `_get_judge_score`, `__init__` all unchanged.

## Test plan

- [x] `pytest tests/test_arena_metrics.py` — 6/6 passing (5 existing tests + 1 new for `num_repeats > 1`).
- [x] `yaml.safe_dump` round-trip on the returned metrics dict (verifies the numpy-to-native casts).
- [x] Bit-for-bit match against real Arena-Hard-v2 production data: ran the patched class on `output-rs*.jsonl` from a 500-prompt × 5-repeat eval; produced `pass@5 = 97.82` (matching the unchanged best-of-N path) and `pass@1[avg-of-5] = 92.258` (matching an offline reference computation that uses the same `np.random.seed(42)` Elo bootstrap; per-repeat scores `[91.99, 91.95, 93.82, 91.48, 92.05]`).

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate computation of pass@N and pass@1[avg-of-N] with correct per-repeat averaging; metric outputs now serialize to native types.

* **Refactor**
  * Metrics now store per-prompt score pairs and perform centralized aggregation with best-direction selection for Arena-style metrics.

* **Tests**
  * Expanded tests for pass@N, pass@1[avg-of-N], repeat behavior, per-category breakdowns, and invalid-score handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Skills/pull/1429)


kslowik:
@coderabbitai review

shuoyangd:
Same thing as #1411 ?

Kipok (approved):

PR #1433, opened 2026-05-06
Pin litellm to 1.83.14 and switch installs to uv

gwarmstrong:
# Pin litellm to 1.83.14 and switch installs to `uv` to satisfy the resolver

## Summary

Bumps `litellm[caching]` from `1.83.0` to `1.83.14` (CVE remediation). That
version pins `httpx==0.28.1`, which conflicts with the `httpx[http2]==0.27.2`
hard pin transitively required by `nemo_run` → `leptonai`. The pip resolver
gives up; uv has the same problem unless we tell it to override the pin.

Fix:

- Add `[tool.uv].override-dependencies` in `pyproject.toml` to relax
  `httpx[http2]` to `>=0.28.1` and `urllib3` to `>=2.6.3` (also a CVE; pinned
  via `torchx 0.7.0`'s `<1.27` constraint).
- Switch the install commands in `dockerfiles/Dockerfile.nemo-skills` and the
  three CI workflows (`lint.yml`, `tests.yml`, `gpu_tests.yml`) from `pip` to
  `uv pip install --system`. `[tool.uv]` is uv-only — plain pip ignores it and
  the resolver fails.

`uv` is already installed in `Dockerfile.nemo-skills`; the CI workflows now
add it alongside the existing `pip install --upgrade pip` step.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflows and Docker setup to use a new dependency management tool for installing project and test dependencies.
  * Upgraded litellm[caching] from 1.83.0 to 1.83.14.
  * Added dependency override settings in project configuration to enforce specific package versions and improve resolution consistency.

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -42,10 +42,12 @@ jobs:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
       run: |
         cd ${{ github.run_id }}
-        python -m pip install --upgrade pip
-        pip uninstall -y nemo-skills nemo_run
-        pip install -e .
-        pip install -r requirements/common-tests.txt
+        python -m pip install --upgrade pip uv
+        uv pip uninstall --system nemo-skills nemo_run || true
+        # Use `uv pip` so [tool.uv].override-dependencies in pyproject.toml is honored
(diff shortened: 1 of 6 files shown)


Kipok (approved):
maybe also update installation instructions in docs to recommend going through uv, but otherwise looks good

PR #1439, opened 2026-05-11
Require tokenizer only when necessary

Kipok:
## Summary by CodeRabbit

* **Bug Fixes**
  * Made tokenizer initialization more selective: tokenizers are now only loaded when needed for retry-based prompt reduction strategies, avoiding unnecessary initialization.

* **Tests**
  * Added tests validating tokenizer initialization behavior across different inference and retry configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Skills/pull/1439)


gwarmstrong (approved):
FYI I don't think Gym has an equivalent concepts for `enable_soft_fail` or `context_limit_retry_strategy`, but otherwise LGTM

PR #1440, opened 2026-05-11
Make python tool timeout in description customizable

Kipok:
Also changes default to align description with actual default code behavior

## Release Notes

* **New Features**
  * Python code execution tools now support fully configurable execution timeouts. Users can specify custom maximum execution durations when creating tools, with improved consistency through a unified default timeout approach working across all tool implementations.

* **Tests**
  * Added comprehensive test coverage validating timeout configuration, validation, and correct propagation through tool listing and execution workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Skills/pull/1440)

### DIFF
--- a/nemo_skills/mcp/servers/python_tool.py
+++ b/nemo_skills/mcp/servers/python_tool.py
@@ -43,18 +43,24 @@ mcp = FastMCP(name="python_tool")
 # Initialized from config in main()
 sandbox = None
 
-# TODO: how should we control timeout in description?
-description = (
-    "Call this function to execute Python code in a stateful Jupyter notebook environment. "
-    "Python will respond with the output of the execution or time out after 120.0 seconds."
-)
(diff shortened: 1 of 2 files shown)


gwarmstrong (commented):
Can you do a test to confirm that if you set the timeout here: https://github.com/NVIDIA-NeMo/Gym/blob/main/resources_servers/ns_tools/configs/ns_tools.yaml#L34

That it actually propagates to the timeout for the sandbox? My intuition is that it does, but good to double check. 

We should also put some thought into how we pass a different timeout to Gym's dataset preparation: https://github.com/NVIDIA-NeMo/Gym/blob/main/resources_servers/ns_tools/prepare_dataset.py

Kipok:
verified that it works on the gym side. There is no way to customize when preparing the dataset, but we can address that separately

PR #1441, opened 2026-05-11
Allow separate sbatch_kwargs for judge / summarize_results

Kipok:
This is useful when setting segment, e.g. if main generation uses 2 nodes but judge 1, currently code fails if {segment: 2} is used in sbatch_kwargs. New parameters allow to override that.

Also fix a bug with summarize_results not propagating account parameter. 

## Summary by CodeRabbit

* **New Features**
  * Added `--judge-sbatch-kwargs` and `--summarize-sbatch-kwargs` CLI options for configuring separate Slurm batch job parameters at the judge and summarize-results pipeline stages, enabling independent resource allocation per stage while maintaining default inheritance behavior.

* **Tests**
  * New tests validate Slurm configuration inheritance and overrides across evaluation stages, ensuring proper configuration propagation and account preservation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Skills/pull/1441)

### DIFF
--- a/nemo_skills/pipeline/eval.py
+++ b/nemo_skills/pipeline/eval.py
@@ -46,6 +46,12 @@ class SingleNodeMode(str, enum.Enum):
     parallel = "parallel"
 
 
+def _resolve_child_sbatch_kwargs(sbatch_kwargs, child_sbatch_kwargs):
+    if child_sbatch_kwargs is None:
+        return sbatch_kwargs
+    return parse_kwargs(child_sbatch_kwargs)
+
+
 def _create_llm_judge_tasks(
     ctx,
     expname,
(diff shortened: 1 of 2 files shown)


nemo_skills/pipeline/eval.py
```diff
@@ -753,7 +773,7 @@ def convert_server_type_to_string(st):
                     _task_dependencies=_task_dependencies,
                     installation_command=installation_command,
                     skip_hf_home_check=skip_hf_home_check,
-                    sbatch_kwargs=sbatch_kwargs,
+                    sbatch_kwargs=judge_sbatch_kwargs,
```
gwarmstrong, line 776: do we want these to always be separate, or just if they are specified?

nemo_skills/pipeline/eval.py
```diff
@@ -850,12 +870,13 @@ def convert_server_type_to_string(st):
...
                     reuse_code_exp=reuse_code_exp,
                     reuse_code=reuse_code,
+                    account=account,
                     task_dependencies=(
                         dependent_tasks if cluster_config["executor"] == "slurm" else all_tasks + _task_dependencies
                     ),
                     installation_command=installation_command,
                     skip_hf_home_check=skip_hf_home_check,
-                    sbatch_kwargs=sbatch_kwargs,
+                    sbatch_kwargs=summarize_sbatch_kwargs,
```
gwarmstrong, line 879: same here

nemo_skills/pipeline/eval.py
Kipok: they default to be same as sbatch_kwargs by `_resolve_child_sbatch_kwargs`

gwarmstrong: Ah I see that now--thanks!

gwarmstrong (approved):

PR #1461, opened 2026-05-24
pipeline: skip cross-experiment string deps in declarative._reuse_exp branch

nicgupta-nvidia:
## Summary

When an orchestrator chains multiple nemo-skills CLI calls across separate `run.Experiment` contexts on a **non-Slurm executor** (Ray, local), declarative pipelines crash with:

```
✗ Evaluation failed: Dependency sft-training not found.
```

This blocks any production flow where one tool submits an upstream stage, exits, and a downstream tool then submits an eval / downstream stage that references the upstream stage by name via `run_after=["sft-training"]`. The Slurm branch handles this case by resolving cross-experiment refs against the persisted experiment directory via `get_exp_handles()`; the non-Slurm `_reuse_exp` branch assumes the dep is in the current experiment's `jobs` list and blindly appends it to `internal_deps`, which then triggers the assertion in nemo-run's `Experiment.add` (`run/experiment.py:582`).

## Reproduction

Two-experiment Ray flow that mirrors what an orchestrator (e.g., a workflow runner that chains stages) does:

```python
from nemo_skills.pipeline.utils.exp import add_task, get_exp
from nemo_skills.pipeline.cli import eval as nemo_eval

# Stage 1 — training in its own experiment context
with get_exp("my-train-expname", cluster_config) as exp_a:
    add_task(
        exp_a,
        cmd=train_cmd,
        task_name="sft-training",        # bare expname, not suffixed
        cluster_config=cluster_config,   # executor: ray
        num_gpus=8,
    )
# exp_a closes; nemo-run persists experiment metadata

# Stage 2 — eval in a fresh experiment context, references stage-1 by name
nemo_eval(
    cluster=cluster,
    expname="my-eval-expname",
    benchmarks="secque:1",
    run_after=["sft-training"],          # cross-experiment string dep
)
```

On Slurm this works because `get_exp_handles("sft-training")` loads the prior experiment from disk and pushes its handles into `external_deps` (which become `--dependency=afterok:<jobid>` on the sbatch). On Ray:

```
File "/.../nemo_skills/pipeline/utils/declarative.py", line 484, in run
    LOG.info(f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)")
File "/.../nemo_run/run/experiment.py", line 582, in add
    assert dep in job_ids, f"Dependency {dep} not found."
AssertionError: Dependency sft-training not found.
```

## Root cause

`declarative.py`'s `_reuse_exp` branch:

```python
elif _reuse_exp:
    # For non-SLURM executors with _reuse_exp, string deps are internal task handles
    internal_deps.append(dep)
    LOG.info(f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)")
```

`internal_deps` is then passed as `dependencies=` to `exp.add(...)`, which calls into `nemo_run/run/experiment.py:582`:

```python
job_ids = set([job.id for job in self.jobs])
for dep in dependencies or []:
    assert dep in job_ids, f"Dependency {dep} not found."
```

The assertion is correct for *intra-experiment* deps. The bug is that the `_reuse_exp` branch blindly appends *any* string dep without verifying it actually exists in the reused experiment's `jobs` list — so a cross-experiment dep (legitimately referenced by an orchestrator chaining nemo-skills CLI calls) trips the assertion.

## Fix

Check `_reuse_exp.jobs` before appending. In-experiment string deps go to `internal_deps` (unchanged behavior). Cross-experiment string deps are logged as a warning and skipped; the caller is expected to have ensured upstream completion before reaching this code path.

The dep-type semantics now break down cleanly:

| Dep type | Resolution |
|---|---|
| `dict` (`{"name": ...}`) | In-experiment reference; resolved against `job_name_to_handle` populated during the same `Pipeline.run()` call. |
| `str` on Slurm | Cross-experiment reference; resolved via `get_exp_handles()` reading nemo-run's persisted experiment dir. |
| `str` on non-Slurm with `_reuse_exp` | **Before:** assumed in-experiment, blindly appended. **After:** checked against `_reuse_exp.jobs`; in-experiment refs preserved, cross-experiment refs warned + skipped. |

For non-Slurm executors there's no equivalent of `get_exp_handles()` — Ray's job state lives in the Ray cluster, not nemo-run's on-disk experiment directory, and there's no general cross-experiment handle to push to `external_deps`. The orchestrator is responsible for sequencing across experiment contexts (most commonly via a synchronous wait on the upstream job's terminal state before the downstream call returns). nemo-skills shouldn't crash when receiving a dep it can't itself resolve — log + skip is the well-defined fallback.

## What this PR does NOT change

- **Slurm behavior**: unchanged. The Slurm branch (~line 464) still resolves cross-experiment deps via `get_exp_handles()` first; it only falls into the `_reuse_exp` branch when `get_exp_handles()` returns empty (defensive fallback). Happy to apply the same `_reuse_exp.jobs` check there in a follow-up commit if reviewers prefer.
- **Dict deps**: unchanged. They still resolve against `job_name_to_handle` (the intra-experiment map populated as the loop iterates).
- **In-experiment string deps**: unchanged. They still flow into `internal_deps` and through `exp.add(dependencies=...)`.
- **No new dependency**, no new public API.

## Empirical validation

Validated against an internal Ray-on-Slurm cluster, applied as a runtime patch downstream while this PR is in review. 3 cross-experiment string deps (`sft-training`, `sft-eval-prepare-data`, `sft-eval-step10-convert`) correctly logged + skipped on a representative SFT+eval pipeline. All upstream Ray jobs reached SUCCEEDED before the eval-generation call. No crash at `experiment.py:582`'s assertion.

## Test plan

- [ ] Manual: run an orchestrator that chains two nemo-skills CLI calls across separate Experiment contexts on `executor: ray` with `run_after=[<upstream-stage-name>]`. Before fix: crash. After fix: WARNING log + eval proceeds.
- [ ] Suggested unit test home: `tests/pipeline/test_declarative.py` (doesn't yet exist as of this writing). Happy to add one if reviewers point at the right test scaffold.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened dependency handling when reusing experiments: string-form dependencies are now validated against the set of reused job IDs and only accepted when matched.
  * SLURM executors still resolve string dependencies into external handles when available.
  * Non-reused runs no longer silently drop string dependencies—these now emit warnings and are skipped.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/Skills/pull/1461?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nemo_skills/pipeline/utils/declarative.py
+++ b/nemo_skills/pipeline/utils/declarative.py
@@ -463,9 +463,7 @@ class Pipeline:
                 # otherwise nemo-run's Experiment.add asserts on missing membership.
                 # Compute the set once per pipeline.run() call to avoid the cost
                 # of rebuilding it per dep.
-                reuse_exp_job_ids = (
-                    {job.id for job in _reuse_exp.jobs} if _reuse_exp else set()
-                )
+                reuse_exp_job_ids = {job.id for job in _reuse_exp.jobs} if _reuse_exp else set()
                 for dep in job_dependencies:
                     if isinstance(dep, str):
                         # String dependency = external experiment name
--- a/nemo_skills/pipeline/utils/declarative.py
+++ b/nemo_skills/pipeline/utils/declarative.py
@@ -458,6 +458,14 @@ class Pipeline:
                     run_after_list = self.run_after if isinstance(self.run_after, list) else [self.run_after]
                     job_dependencies = run_after_list
 
+                # String deps with _reuse_exp must be checked against the reused
+                # experiment's jobs list before being appended to internal_deps,
+                # otherwise nemo-run's Experiment.add asserts on missing membership.
+                # Compute the set once per pipeline.run() call to avoid the cost
+                # of rebuilding it per dep.
+                reuse_exp_job_ids = (
+                    {job.id for job in _reuse_exp.jobs} if _reuse_exp else set()
+                )
                 for dep in job_dependencies:
                     if isinstance(dep, str):
                         # String dependency = external experiment name
@@ -468,11 +476,21 @@ class Pipeline:
                                     f"No pending or running tasks found for experiment {dep}, cannot set dependencies."
                                 )
                                 # If no experiment found, treat as direct task handle (for _reuse_exp case)
+                                # — but only if it's actually a job in the reused experiment.
+                                # Cross-experiment refs that didn't resolve via get_exp_handles
+                                # cannot be enforced via exp.add() either; warn and skip.
                                 if _reuse_exp:
-                                    internal_deps.append(dep)
-                                    LOG.info(
-                                        f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)"
-                                    )
+                                    if dep in reuse_exp_job_ids:
+                                        internal_deps.append(dep)
+                                        LOG.info(
+                                            f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)"
+                                        )
+                                    else:
+                                        LOG.warning(
+                                            f"Job '{job_name}' references external task handle '{dep}' "
+                                            f"not present in reused experiment; skipping dependency "
+                                            f"declaration. Caller must ensure '{dep}' has completed."
+                                        )
                             else:
                                 external_deps.extend(exp_handles)
                                 LOG.info(
@@ -487,7 +505,6 @@ class Pipeline:
                             # exp.add() — nemo-run asserts dep in self.jobs. Skip with
                             # a warning so the caller (who has already ensured upstream
                             # completion) can proceed.
-                            reuse_exp_job_ids = {job.id for job in _reuse_exp.jobs}
                             if dep in reuse_exp_job_ids:
                                 internal_deps.append(dep)
                                 LOG.info(f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)")
@@ -497,6 +514,16 @@ class Pipeline:
                                     f"not present in reused experiment; skipping dependency "
                                     f"declaration. Caller must ensure '{dep}' has completed."
                                 )
+                        else:
+                            # Non-SLURM, no _reuse_exp: no way to resolve a string dep.
+                            # Don't silently drop it — surface it so callers can debug.
+                            LOG.warning(
+                                f"Job '{job_name}' references string dependency '{dep}', "
+                                f"but executor '{self.cluster_config['executor']}' has no "
+                                f"reused experiment context to resolve it against; "
+                                f"skipping dependency declaration. Caller must ensure "
+                                f"'{dep}' has completed before this job runs."
+                            )
                     elif isinstance(dep, dict):
                         # Dict dependency = internal job reference (by job spec object)
                         dep_name = dep.get("name")
--- a/nemo_skills/pipeline/utils/declarative.py
+++ b/nemo_skills/pipeline/utils/declarative.py
@@ -479,9 +479,24 @@ class Pipeline:
                                     f"Job '{job_name}' depends on external experiment '{dep}' ({len(exp_handles)} tasks)"
                                 )
                         elif _reuse_exp:
-                            # For non-SLURM executors with _reuse_exp, string deps are internal task handles
-                            internal_deps.append(dep)
-                            LOG.info(f"Job '{job_name}' depends on task handle '{dep}' (from reused experiment)")
+                            # For non-SLURM executors with _reuse_exp, string deps are
+                            # internal task handles ONLY if they reference a job already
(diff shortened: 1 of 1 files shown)


gwarmstrong (commented):
We don't really support this kind of cross `add_task`/`Pipeline` dependency. Why is it not sufficient to just drop the `run_after` value from your `eval()` call if you're already orchestrating the execution ordering of your jobs?

nicgupta-nvidia:
> We don't really support this kind of cross `add_task`/`Pipeline` dependency. Why is it not sufficient to just drop the `run_after` value from your `eval()` call if you're already orchestrating the execution ordering of your jobs?

@gwarmstrong thanks for the look. Worth clarifying what the warning actually means here because I think it's reading as "drop dep, hope for the best" because that's not quite what's happening.

The dep IS still enforced on the Ray path; it gets enforced via polling `in _wait_for_dependencies` (`ray_executor.py`) instead of via `exp.add()`'s `dependencies=` list. The PR's `declarative.py` change fires the warning at a layer where the dep was already going to be handed off to the polling path anyway. So I think you're right that the warning is misleading, but the underlying behavior isn't silent ignore, just split between two enforcement mechanisms (Slurm afterok vs Ray polling).


Two ways to clean this up:

1. Drop the warning (or downgrade to INFO): the dep DOES get enforced; the WARNING level is what makes it read as an error coerced to a swallow.
2. Push the fix earlier: detect "cross-experiment string dep and Ray executor" at the `nemo_eval` boundary (or one layer up) and either (a) just dispatch to polling silently with one INFO log, or (b) require the caller to pass `_external_run_after=` so the contract is explicit at the public API.

For my call site (`nvflow/recipes/finance/stages/evaluation/evaluate.py:642`) I can't drop `run_after` outright because it's load-bearing on Slurm (`--dependency=afterok:<jobid>`). I'd need a way to signal "Slurm: use afterok, Ray: use polling" at one place. Option 2(b) would give me that cleanly.

 Which direction do you want? Happy to refactor.

nicgupta-nvidia:
> > We don't really support this kind of cross `add_task`/`Pipeline` dependency. Why is it not sufficient to just drop the `run_after` value from your `eval()` call if you're already orchestrating the execution ordering of your jobs?
> 
> @gwarmstrong thanks for the look. Worth clarifying what the warning actually means here because I think it's reading as "drop dep, hope for the best" because that's not quite what's happening.
> 
> The dep IS still enforced on the Ray path; it gets enforced via polling `in _wait_for_dependencies` (`ray_executor.py`) instead of via `exp.add()`'s `dependencies=` list. The PR's `declarative.py` change fires the warning at a layer where the dep was already going to be handed off to the polling path anyway. So I think you're right that the warning is misleading, but the underlying behavior isn't silent ignore, just split between two enforcement mechanisms (Slurm afterok vs Ray polling).
> 
> Two ways to clean this up:
> 
> 1. Drop the warning (or downgrade to INFO): the dep DOES get enforced; the WARNING level is what makes it read as an error coerced to a swallow.
> 2. Push the fix earlier: detect "cross-experiment string dep and Ray executor" at the `nemo_eval` boundary (or one layer up) and either (a) just dispatch to polling silently with one INFO log, or (b) require the caller to pass `_external_run_after=` so the contract is explicit at the public API.
> 
> For my call site (`nvflow/recipes/finance/stages/evaluation/evaluate.py:642`) I can't drop `run_after` outright because it's load-bearing on Slurm (`--dependency=afterok:<jobid>`). I'd need a way to signal "Slurm: use afterok, Ray: use polling" at one place. Option 2(b) would give me that cleanly.
> 
>  Which direction do you want? Happy to refactor.

Closing this one — and thanks for the careful review; you were right to push back.
  
On reflection your read holds up: threading `run_after` through the `executor: ray` branch of `add_task` wasn't the right place. The redundancy you flagged at `eval.py:119` (`run_after=run_after` in `_create_llm_judge_tasks`, already covered by `_task_dependencies=dependent_tasks`) is the cleaner framing of the actual issue.

The bigger reason, though: we've moved our integration to a Ray Jobs API submission path (`backend.name: ray`) rather than the `executor: ray` scheduler path this PR patches. On that path, cross-task ordering is resolved inside the submission queue itself (`task_dependencies` → per-job `dep_task_names`), so the gap this PR worked around doesn't exist there — the change here would effectively be dead code for us.
  
So I'm going to close #1461 rather than rework it. If the legacy `executor: ray` path ever needs `run_after` honored, we'll do the minimal targeted fix at `eval.py:119` as you suggested — no need to carry this branch in the meantime.

Appreciate the time on the review.

PR #1465, opened 2026-05-28
Warn on risky hosted server allocations

tamohannes:
## Summary
Adds hosted-server guardrails for common wasted-compute configurations:
- warn when a partial-node hosted server uses a fixed port
- warn when `exclusive=True` may reserve more GPUs than the server can use
- make random-port behavior aware of cluster `gpus_per_node`
- keep submission allowed; this is warning-only

## Why
Partial-node serving jobs can collide on fixed localhost ports, and exclusive allocations can reserve idle GPUs when `server_gpus` is smaller than the node size. Both patterns can produce idle-GPU waste while the job appears submitted successfully.

## Tests
- `python -m pytest tests/test_pipeline_utils.py::test_should_get_random_port_respects_cluster_gpu_count tests/test_pipeline_utils.py::test_get_cluster_gpus_per_node_known_clusters_and_override tests/test_pipeline_utils.py::test_warn_hosted_server_allocation_partial_fixed_port tests/test_pipeline_utils.py::test_warn_hosted_server_allocation_partial_exclusive tests/test_pipeline_utils.py::test_warn_hosted_server_allocation_random_partial_has_no_port_warning -q`
- `python -m ruff check nemo_skills/pipeline/generate.py nemo_skills/pipeline/start_server.py nemo_skills/pipeline/utils/server.py nemo_skills/pipeline/nemo_rl/grpo.py nemo_skills/pipeline/verl/ppo.py tests/test_pipeline_utils.py`

## Summary by CodeRabbit

* **New Features**
  * New warnings for hosted-server GPU allocation that may waste resources or risk localhost port conflicts
  * Start-server CLI option now supports an “unset” mode that auto-resolves port behavior from cluster GPU layout

* **Improvements**
  * Port-allocation logic now considers cluster GPU counts and exclusive-allocation settings for smarter defaults and fewer port collisions

* **Tests**
  * Added tests covering cluster-aware port selection and allocation-warning behavior

### DIFF
--- a/nemo_skills/pipeline/start_server.py
+++ b/nemo_skills/pipeline/start_server.py
@@ -35,7 +35,6 @@ from nemo_skills.pipeline.utils import (
     resolve_mount_paths,
     set_python_path_and_wait_for_server,
     should_get_random_port,
-    warn_hosted_server_allocation,
 )
 from nemo_skills.utils import get_logger_name, setup_logging
 
@@ -156,20 +155,15 @@ def launch_server(
     log_dir = check_mounts(cluster_config, log_dir, check_mounted_paths=check_mounted_paths)
 
     gpus_per_node = get_cluster_gpus_per_node(cluster_config)
-    if get_random_port is None:
-        get_random_port = should_get_random_port(
-            server_gpus=server_gpus,
-            exclusive=(sbatch_kwargs or {}).get("exclusive") if isinstance(sbatch_kwargs, dict) else None,
-            gpus_per_node=gpus_per_node,
-        )
-    warn_hosted_server_allocation(
+    exclusive = (sbatch_kwargs or {}).get("exclusive")
+    resolved_random_port = should_get_random_port(
         server_gpus=server_gpus,
-        exclusive=(sbatch_kwargs or {}).get("exclusive") if isinstance(sbatch_kwargs, dict) else None,
+        exclusive=exclusive,
         gpus_per_node=gpus_per_node,
-        get_random_port=get_random_port,
         server_port=server_port,
-        context="ns start_server",
     )
+    if get_random_port is None:
+        get_random_port = resolved_random_port
 
     if server_port is None:
         server_port = get_free_port(strategy="random") if get_random_port else 5000
@@ -284,20 +278,13 @@ def start_server(
     """Self-host a model server."""
     cluster_config = get_cluster_config(cluster, config_dir)
     gpus_per_node = get_cluster_gpus_per_node(cluster_config)
-    if get_random_port is None:
-        get_random_port = should_get_random_port(
-            server_gpus=server_gpus,
-            exclusive=exclusive,
-            gpus_per_node=gpus_per_node,
-        )
-    warn_hosted_server_allocation(
+    resolved_random_port = should_get_random_port(
         server_gpus=server_gpus,
         exclusive=exclusive,
         gpus_per_node=gpus_per_node,
-        get_random_port=get_random_port,
-        server_port=None,
-        context="ns start_server",
     )
+    if get_random_port is None:
+        get_random_port = resolved_random_port
 
     server_port = get_free_port(strategy="random") if get_random_port else 5000
     sandbox_port = get_free_port(strategy="random") if get_random_port else 6000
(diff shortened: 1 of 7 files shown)


nemo_skills/pipeline/utils/server.py
```diff
@@ -64,8 +64,74 @@ def get_free_port(exclude: list[int] | None = None, strategy: int | str = 5000)
...
+    if not cluster_config:
+        return default
+
+    for key in ("gpus_per_node", "num_gpus_per_node"):
+        value = cluster_config.get(key)
+        if value is not None:
+            return int(value)
+
+    cluster_name = str(cluster_config.get("_cluster_yaml_name") or cluster_config.get("name") or "").lower()
+    if any(token in cluster_name for token in ("aws-cmh", "aws-dfw")):
```
Kipok: let's not reference any internal infrastructure. We can just add this as explicit argument in cluster config or default to 8 otherwise

nemo_skills/pipeline/verl/ppo.py
```diff
@@ -375,7 +375,16 @@ def ppo_verl(
 
     server_config = None
     if server_type is not None:
-        get_random_port = pipeline_utils.should_get_random_port(server_gpus, exclusive)
+        gpus_per_node = pipeline_utils.get_cluster_gpus_per_node(cluster_config)
+        get_random_port = pipeline_utils.should_get_random_port(server_gpus, exclusive, gpus_per_node)
+        pipeline_utils.warn_hosted_server_allocation(
```
Kipok: this should be done for all pipelines, not just the ones you added it for? E.g. eval, sft, etc.

nemo_skills/pipeline/verl/ppo.py
```diff
@@ -375,7 +375,16 @@ def ppo_verl(
...
+            server_gpus=server_gpus,
+            exclusive=exclusive,
+            gpus_per_node=gpus_per_node,
+            get_random_port=get_random_port,
+            server_port=None,
+            context="ns verl ppo",
```
Kipok: I think this might get stale very fast and not super useful, I'd probably not use context at all

nemo_skills/pipeline/start_server.py
```diff
@@ -152,6 +155,22 @@ def launch_server(
 
     log_dir = check_mounts(cluster_config, log_dir, check_mounted_paths=check_mounted_paths)
 
+    gpus_per_node = get_cluster_gpus_per_node(cluster_config)
+    if get_random_port is None:
+        get_random_port = should_get_random_port(
+            server_gpus=server_gpus,
+            exclusive=(sbatch_kwargs or {}).get("exclusive") if isinstance(sbatch_kwargs, dict) else None,
```
Kipok: sbatch_kwargs should already be resolved to always be dict or none at this point. And we can probably check for exclusive ones

nemo_skills/pipeline/start_server.py
```diff
@@ -263,6 +282,23 @@ def start_server(
     sandbox_tunnel_port: int = typer.Option(6000, help="Local tunnel port for the sandbox server."),
 ):
     """Self-host a model server."""
+    cluster_config = get_cluster_config(cluster, config_dir)
+    gpus_per_node = get_cluster_gpus_per_node(cluster_config)
+    if get_random_port is None:
+        get_random_port = should_get_random_port(
+            server_gpus=server_gpus,
+            exclusive=exclusive,
```
Kipok: I think exclusive can also be passed through sbatch kwargs, so better to check for that one (and it should have explicit exclusive fused already

Kipok (changes requested):

tamohannes:
Updated per review:
- `get_cluster_gpus_per_node` no longer hard-codes cluster names — it reads `gpus_per_node`/`num_gpus_per_node` from the cluster config, defaulting to 8.
- Folded the guardrail into `should_get_random_port` so every server-hosting pipeline (generate/eval, grpo, ppo, start_server) gets it — removed the per-call-site `warn_hosted_server_allocation` calls and the `context` arg.
- Simplified the `sbatch_kwargs`/`exclusive` handling; an explicitly pinned `server_port` is now treated as a fixed port for the collision warning (+ tests).

PR #1466, opened 2026-05-28
Add timeout for tool calls

tamohannes:
## Summary
Adds `GenerationTaskConfig.tool_call_timeout_s` with a default of `300.0` seconds and wraps tool execution in a wall-clock timeout. When a tool times out, generation receives a model-visible tool error instead of waiting indefinitely.

## Why
If retrieval/tool/environment calls hang while an inference server is allocated, GPUs can sit idle until the Slurm job is cancelled by idle-GPU monitoring. A bounded timeout lets the sample fail or continue cleanly.

## Tests
- `python -m pytest tests/test_mcp_clients.py::test_tool_calling_wrapper_times_out_slow_tool_call -q`
- `python -m ruff check nemo_skills/inference/generate.py nemo_skills/inference/model/__init__.py nemo_skills/inference/model/tool_call.py tests/test_mcp_clients.py`

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-tool-call execution timeout (default 300s). Tool calls that exceed the timeout are aborted and return a structured error indicating the call timed out, improving robustness of tool integrations. Initialization validates timeout values and rejects non-positive values.

* **Tests**
  * Added tests verifying timeout enforcement for slow tools and validation that non-positive timeout values are rejected.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -202,6 +202,7 @@ class GenerationTaskConfig:
     schema_overrides: dict | None = field(default_factory=dict)
 
     max_tool_calls: int = -1  # If >= 0, will limit the number of tool calls executed during generation to this number
+    tool_call_timeout_s: float | None = 300.0  # Wall-clock timeout for each individual tool call
 
     # if True, will move full generation to _full_generation key and keep cfg.generation_key without thinking tokens
     # IMPORTANT: do not set this for non-reasoning models as it will make the generations empty!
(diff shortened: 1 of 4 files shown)


Kipok (commented):
@gwarmstrong can you also check this? I think it's probably appropriate to have some client side timeout as currently, e.g. search tools don't see to have any configured (and in principle this can be used with external tool implementations that we don't have control over I guess). But ideally we somehow also enforce this on our own tool implementations so that the tool calls handle timeout appropriately directly, otherwise we might have a bunch of orphan processes

tamohannes:
Addressed the bot findings: validate `tool_call_timeout_s > 0` at construction, and the `SlowTool` test fails fast on unexpected args.

@gwarmstrong on the bigger point — the wrapper-level `wait_for` cancels the await but the underlying tool process can keep running, so a timeout can still leave an orphan. Enforcing timeouts inside the tool implementations themselves is worth a follow-up; happy to scope it.

PR #1467, opened 2026-05-28
Default Python tool to direct provider

tamohannes:
## Summary
Makes `nemo_skills.mcp.servers.python_tool::PythonTool` use the direct in-process implementation by default, while keeping the previous stdio implementation available as `PythonStdioTool`.

## Why
The stdio subprocess transport is unnecessary for the in-repo Python tool and can leave extra subprocess lifecycle failure modes. New users should get the simpler direct provider by default while legacy stdio remains explicit.

## Tests
- `python -m pytest tests/test_mcp_clients.py::test_python_tool_defaults_to_direct_provider -q`
- `python -m ruff check nemo_skills/mcp/servers/python_tool.py tests/test_mcp_clients.py`

## Summary by CodeRabbit

* **Changes**
  * Python tool now uses the in-process direct execution path by default, improving performance and predictability.

* **Documentation**
  * Clarified supported configuration keys and how to reference the in-process provider; legacy transport-specific keys omitted.

* **Tests**
  * Added tests for stateful session behavior, exception reporting, and the default in-process provider.

### DIFF
--- a/nemo_skills/mcp/servers/python_tool.py
+++ b/nemo_skills/mcp/servers/python_tool.py
@@ -26,7 +26,6 @@ from pydantic import Field
 
 from nemo_skills.code_execution.sandbox import get_sandbox
 from nemo_skills.mcp.tool_manager import Tool
-from nemo_skills.mcp.tool_providers import MCPClientTool
 from nemo_skills.mcp.utils import add_config_args, load_mcp_config
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


Kipok (commented):
@gwarmstrong any downside to just removing the mcp implementation completely and then remapping PythonTool to DirectPythonTool for bc?

tamohannes:
Done as suggested: removed the stdio `PythonTool(MCPClientTool)` implementation and remapped `PythonTool` to a thin `DirectPythonTool` subclass for bc. Converted the MCP-vs-direct parity tests to direct-only functional tests and dropped the stdio-specific timeout test.

I kept the generic FastMCP stdio server (`main()`) since `clients.py` and the generic stdio-client test still reference it — happy to remove that too if we're confident nothing else needs the python stdio server.

gwarmstrong:
@Kipok none that I can think of, only risk would be it changes the code path, but I think we've been using DirectPythonTool enough that it should be safe.

Kipok (approved):

PR #1470, opened 2026-05-29
Add direct Tavily browser tool

jiacheng-xu:
latest search tool for ultra 
ultra reached 37+ with this tavily tool + python tool. 

## Summary by CodeRabbit

* **New Features**
  * Per-request tool execution metrics are now aggregated and attached to generated results, including streamed outputs; request-scoped cleanup and metrics retrieval are supported.
  * Tavily: added direct HTTP-based tools with browser-style features (caching, search memory, open-result), Gym endpoints (search/extract/scroll), exclude-domain config, input validation, formatted/truncated outputs, and API-key rotation.

* **Tests**
  * New tests covering metrics inclusion and extensive Tavily behaviors: formatting, caching, validation, rotation, scrolling, and metrics reporting.

### DIFF
--- a/nemo_skills/mcp/servers/tavily_search_tool.py
+++ b/nemo_skills/mcp/servers/tavily_search_tool.py
@@ -98,6 +98,7 @@ DEFAULT_API_BASE_URL = "https://api.tavily.com"
 DEFAULT_HTTP_TIMEOUT_S = 30.0
 DEFAULT_MAX_RETRIES = 2
 DEFAULT_RETRY_BACKOFF_S = 0.5
+DEFAULT_MAX_RETRY_DELAY_S = 30.0
 DEFAULT_GYM_MAX_RESULTS = 10
 DEFAULT_GYM_MAX_RESULT_CHARS = 2000
 DEFAULT_GYM_MAX_QUERY_CHARS = 400
@@ -120,6 +121,8 @@ STATUS_CODE_ERRORS = {
 }
 
 RETRY_STATUS_CODES = {429, 500, 502, 503, 504}
+GYM_TOOL_ERROR_PREFIXES = ("URL ", "Query ", "Search ")
+GYM_NO_CONTENT_RESULT = "No content found."
 
 # These errors should stop the process - no point continuing with bad credentials
 FATAL_STATUS_CODES = {401, 403}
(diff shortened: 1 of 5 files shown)


nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -69,71 +417,42 @@ async def answer(
...
-        return {"error": "Search request failed due to network error"}
-
-    # Handle non-200 responses
-    if response.status_code in FATAL_STATUS_CODES:
-        return {"error": "Search authentication failed", "fatal": True}
-    if response.status_code != 200:
-        error_msg = STATUS_CODE_ERRORS.get(
-            response.status_code, f"Search request failed with status {response.status_code}"
+        result = await _post_tavily_json(
+            api_base_url=DEFAULT_API_BASE_URL,
```
activatedgeek: Why not define these defaults in the definition itself if they are not being modified at all?

nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -35,32 +54,361 @@ class ExecutionResult:
...
+
+    for attempt in range(max_retries + 1):
+        try:
+            async with httpx.AsyncClient(timeout=timeout_s) as client:
+                response = await client.post(url, headers=headers, json=payload)
+        except httpx.TimeoutException:
+            last_error = "Search request timed out"
+            if attempt < max_retries:
+                await asyncio.sleep(delay)
+                delay *= 2
```
activatedgeek: We probably want to clip this value at some `max_delay` if the `retry_backoff_s` is being treated a general function argument instead of a fixed value.

nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -35,32 +54,361 @@ class ExecutionResult:
...
+        return default, None
+    if isinstance(value, bool):
+        return None, f"{arg_name} must be an integer"
+    try:
+        return int(value), None
+    except (TypeError, ValueError):
+        return None, f"{arg_name} must be an integer"
+
+
+def _remember_lru(cache: OrderedDict[str, Any], key: str, value: Any, max_items: int) -> None:
```
activatedgeek, line 360: Python has a [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator that you can directly apply to a target function. Perhaps reuse?

nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -180,6 +497,659 @@ async def execute(self, tool_name: str, arguments: dict[str, Any], extra_args: d
...
+        ]
+
+    async def execute(
+        self, tool_name: str, arguments: dict[str, Any], extra_args: dict[str, Any] | None = None
+    ) -> str | dict[str, Any]:
+        arguments = self._sanitize_arguments(tool_name, arguments)
+        request_id = (extra_args or {}).get("request_id")
+        if tool_name == "web_search":
+            return await self._web_search(arguments, request_id=request_id)
+        if tool_name == "open_result":
```
activatedgeek: Nit: replace subsequent `if`s with `elif`s.

nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -180,6 +497,659 @@ async def execute(self, tool_name: str, arguments: dict[str, Any], extra_args: d
...
+
+        raw_content = _first_raw_content(results)
+        if not raw_content:
+            return None, "No content found."
+
+        max_cached_page_chars = max(1, int(self._config["max_cached_page_chars"]))
+        raw_content, was_truncated = _truncate_text(raw_content, max_cached_page_chars)
+        cleaned = _clean_text(raw_content)
+        page = TavilyBrowserCachedPage(content=cleaned, words=cleaned.split(), truncated=was_truncated)
+        _remember_lru(cache, url, page, max(1, int(self._config["max_cached_pages"])))
```
activatedgeek: URLs will likely be cache miss if any query parameters change (e.g. marketing query params like `utm_*`). Is that a concern?

activatedgeek (changes requested):
Some changes and some behaviors for consideration. But looking very solid!

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: fixed

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: Keeping the explicit `OrderedDict` LRU helper here. `functools.lru_cache` is function-result scoped and does not fit these mutable request/global caches, where we need explicit per-request cleanup, metrics, and bounded eviction of cached page objects.

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: Addressed: `_post_tavily_json` now has `max_retry_delay_s` and clamps both the initial `retry_backoff_s` and subsequent exponential backoff sleeps to that max. `Retry-After` responses are also capped before sleeping.

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: This thread is outdated after the follow-up changes. The direct tool defaults are now centralized in constants plus each tool class config, while the legacy MCP `answer()` path keeps fixed Tavily defaults inline because those are part of that function’s public parameter defaults.

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: Addressed in `884808c0`: cache keys now normalize URLs by stripping common tracking parameters (`utm_*`, `gclid`, `fbclid`, etc.) and fragments while preserving meaningful query params. Tavily still receives the original URL for extraction.

activatedgeek (approved):

nemo_skills/mcp/servers/tavily_search_tool.py
```diff
@@ -35,32 +54,433 @@ class ExecutionResult:
...
+        if overrides:
+            unknown = set(overrides) - set(cfg)
+            if unknown:
+                raise ValueError(f"Unknown {self.__class__.__name__} override(s): {sorted(unknown)}")
+            cfg.update(overrides)
+
+        self._config = cfg
+        self._api_keys = _normalize_api_keys(cfg.get("tavily_api_key"))
+        self._exclude_domains = list(cfg.get("exclude_domains") or [])
+        if cfg.get("exclude_domains_config"):
```
gwarmstrong: I would HIGHLY recommend making this field REQUIRED. Leaving it as optional is a recipe to leak domains that are required to be excluded into post-training data.

gwarmstrong: And rather than having a simple flag to turn off the requirement, it would be better to have to  know how to construct the config file to allow all domains if that is truly what you want. IMO We want to discourage use without the exclusion config as much as possible.

gwarmstrong (changes requested):

nemo_skills/mcp/servers/tavily_search_tool.py
jiacheng-xu: for evaluation, we can use any domain (maybe exclude huggingface.com ish websites); for training, I agree we need to use the exclude_domains_config. Can we publicize the config? if so, the easiest approach would be using it as default.

gwarmstrong: No, we cannot publish the config.

jiacheng-xu: Understood. let me change the code to make this REQUIRED and add some note here.

gwarmstrong (approved):

PR #1478, opened 2026-06-09
fix: honor skip_special_tokens from extra_body on the vLLM completion path

DongjiGao:
## What

The vLLM **text-completion** request builder hardcoded `skip_special_tokens: False` at the top level, while `_build_request_body` could simultaneously carry a user-supplied `skip_special_tokens` inside `extra_body`. The field was sourced twice, so any `extra_body.skip_special_tokens` was silently dropped on the completion path.

This pops `skip_special_tokens` out of `extra_body` into the single top-level field (default `False`), so `inference.extra_body.skip_special_tokens` is honored on the completion path — matching how the chat path already behaves.

## Why

`skip_special_tokens=false` keeps special tokens (e.g. speaker tags) in the decoded output, which multispeaker ASR needs. The chat endpoint already respects this via `extra_body`; this makes the completion endpoint consistent, without adding any new config surface.

Usage (just `extra_body`, no new options):

```
++inference.extra_body.skip_special_tokens=false
```

## Test

`tests/test_vllm_completion.py`:
- default → top-level `skip_special_tokens=False`, absent from `extra_body`
- `extra_body.skip_special_tokens=True` → lifted to the single top-level field, not duplicated

### DIFF
--- a/nemo_skills/inference/model/vllm.py
+++ b/nemo_skills/inference/model/vllm.py
@@ -150,6 +150,7 @@ class VLLMModel(BaseModel):
         assert reasoning_effort is None, "reasoning_effort is not supported for text completion requests"
         assert tools is None, "tools are not supported for text completion requests"
         assert response_format is None, "response_format is not supported for text completion requests"
+        request_extra_body = self._build_request_body(top_k, min_p, repetition_penalty, extra_body=extra_body)
         return {
             "prompt": prompt,
             "max_tokens": tokens_to_generate,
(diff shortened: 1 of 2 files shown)


Kipok:
@DongjiGao please recreate from a branch

DongjiGao:
Recreated as in-repo branch PR #1488 so CI can run (fork PRs don't receive CI secrets). Closing this fork PR in favor of #1488.

PR #1479, opened 2026-06-09
feat: add a Ray Jobs API execution backend

nicgupta-nvidia:
## What
Add an additive **Ray Jobs API execution backend** to NeMo-Skills. With `backend.name: ray` plus a `dashboard_url` (Jobs API) or `endpoint` (Ray Client), a pipeline submits its commands to a pre-provisioned Ray cluster via `JobSubmissionClient` instead of launching through Slurm. It coexists with the existing executors — **no behavior change when unused**. Supports Kubernetes (KubeRay) and precreated clusters, optional entrypoint label selectors / image-derived placement labels, env-var forwarding into each job's `runtime_env`, and a per-experiment Ray Jobs queue with dependency ordering.

## Design
- New `ExecutionBackend` hook interface and `get_execution_backend(cluster_config)` resolver in `pipeline/utils/backends.py`; the default backend preserves existing Slurm/local behavior. Wired into `cli.py` and `generate.py`.
- `RayBackend` (`pipeline/utils/ray_backend.py`) submits, tracks, and cancels jobs via `JobSubmissionClient`. Commands are queued on the experiment (`_ns_ray_jobs_queue`) and submitted/tracked/cancelled centrally in `start_experiment`, with bounded concurrency and cross-experiment dependency ordering.
- Cluster-configured env (`required_env_vars`, etc.) is forwarded into each job's `runtime_env.env_vars`, so endpoint/judge API keys and `HF_TOKEN` reach jobs regardless of the Ray head's environment.
- Explicit `entrypoint_label_selector` keys take precedence over image-derived label selectors (static-wins, matching the sibling `image_label_key` path).
- Backwards compatible: with no backend configured, behavior is unchanged; the legacy `with_ray` flag resolves to the Ray backend.
- Includes a sample precreated-Ray-on-Kubernetes cluster config under `cluster_configs/`.

## Testing
`tests/test_backends.py` (16 tests): backend resolution across config shapes, env-var forwarding, image-label-selector precedence, the dependency resolver, and poll-failure cleanup. `ruff check`/`format` clean; DCO signed on all commits. The Jobs API submission path has been exercised end-to-end against a real pre-provisioned Ray cluster (hello-world submission, data-prep, SFT, and parallel eval fan-out).

## Review feedback
Recent commits address review (CodeRabbit) findings:
- Forward external `run_after` dependencies into the Ray queue (`declarative.py`, `exp.py`) so cross-experiment ordering is preserved.
- Tighten the dependency resolver to require a proven `SUCCEEDED` for in-batch dependencies (no premature start) while still treating truly cross-experiment dependencies as gated upstream — resolving the earlier false "dependency deadlock" without regressing it.
- Stop submitted jobs best-effort on a Jobs-API polling failure instead of leaving them running.
- Treat `dashboard_url` as a precreated-cluster signal; stop downgrading an explicit `precreated_cluster: true`.
- Narrow the experiment-lookup fallback to `FileNotFoundError`.
- Move the sample `preflight` block to top-level config scope (where it is actually read).

## Summary by CodeRabbit

* **New Features**
  * Added a Kubernetes + SLURM sample configuration that targets a pre-created Ray cluster.
  * Introduced backend-driven execution so Ray and non-Ray paths are selected automatically.
* **Bug Fixes**
  * Corrected dependency handling for cross-experiment “finished” inputs in Ray-backed runs.
  * Improved dry-run behavior and backend routing (including sequential execution rules).
  * Enhanced Ray job queuing, status tracking, and best-effort stop on failures.
* **Tests**
  * Expanded unit coverage for backend selection, Ray runtime configuration, and dependency/poll-failure regressions.


nemo_skills/pipeline/utils/exp.py
```diff
@@ -942,11 +1060,12 @@ def run_exp(exp, cluster_config, sequential=False, dry_run=False):
         )
         check_remote_mount_directories(mount_sources, cluster_config, exit_on_failure=exit_if_failure)
 
-    if cluster_config["executor"] != "slurm":
-        exp.run(detach=False, tail_logs=True, sequential=sequential)
-    else:
+    backend = get_execution_backend(cluster_config)
```
gwarmstrong, line 1102: **`--dry_run` now performs live remote I/O on the non-Ray path (high).**

The top-level `if dry_run: return` that used to be the **first** statement in `run_exp` was removed, so the mount check above this line now runs before `dry_run` is honored.

A default-backend `executor: slurm` config with a `mounts:` entry, invoked via `ns generate/eval/run_cmd … --dry_run`, used to return instantly with zero network access. Now `check_remote_mount_directories` opens an SSH tunnel (`get_tunnel`), runs `test -e` on each mount source, and raises `FileNotFoundError` on a missing source. A dry run now requires live cluster connectivity and can hard-fail offline. This affects **all** non-Ray slurm callers (`eval.py`, `convert.py`, `run_cmd.py`, `robust_eval.py`, `verl/ppo.py`, `megatron_lm/train.py`, `nemo_rl/grpo.py`, `nemo_rl/sft.py`).

**Suggested fix:** restore the early dry-run short-circuit at the top of `run_exp` (or gate the mount check behind `not dry_run`).

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -902,10 +921,60 @@ def _allocation_sort_key(entry: Dict) -> Tuple[int, int]:
...
+        # A run_after naming another experiment whose tasks have already finished
+        # resolves to empty handles and falls through as a bare experiment-name string
+        # in internal_deps. nemo-run's exp.add asserts every dependency is a job in THIS
+        # experiment, so drop deps that are not present here -- cross-experiment ordering
+        # is still honored via external_deps and (on Ray) the queued dep resolver. Only
+        # filters when exp.jobs is introspectable, so it never over-drops valid handles.
+        # Only filter when exp.jobs is a concrete list (a real nemo-run experiment); a
+        # mocked or duck-typed exp leaves deps untouched so valid handles are never dropped.
+        exp_jobs = getattr(exp, "jobs", None)
+        if internal_deps and isinstance(exp_jobs, (list, tuple)):
```
gwarmstrong: **Always-run `internal_deps` drop block silently weakens a dependency guarantee (medium).**

This block filters `internal_deps` against `{job.id for job in exp.jobs}` for **every** backend, not just Ray.

Concretely: default-backend `executor: slurm` with `ns eval … --run_after <prior_exp>`, where the prior experiment's tasks have already finished — `get_exp_handles` returns `[]` and the experiment-name string lands in `internal_deps` (the `_reuse_exp` branch). Base behavior: nemo-run's `exp.add` does `assert dep in job_ids` (`experiment.py:580`) and raises `Dependency <x> not found`, surfacing the unresolved ordering. PR behavior: the block sees the string isn't a job id, logs a warning, and **drops it**, so the job submits with no dependency. A loud failure becomes a silent ordering change on the non-Ray path.

**Suggested fix:** scope this filtering to the Ray backend (it's described as needed for the Ray dep resolver), or otherwise preserve the fail-loud behavior for the default backend.

nemo_skills/pipeline/cli.py
```diff
@@ -38,6 +37,13 @@
 from nemo_skills.pipeline.summarize_robustness import summarize_robustness
 from nemo_skills.pipeline.verl.ppo import ppo_verl
 
+try:
```
gwarmstrong: **Unrelated `nemo_evaluator` import guard bundled into this PR (medium / scope).**

Wrapping `from …nemo_evaluator import nemo_evaluator` in `try/except ModuleNotFoundError` is unrelated to the Ray backend.

In any environment missing `nemo_evaluator_launcher`, the `nemo_evaluator` subcommand now silently disappears (`ns nemo_evaluator …` → "No such command") instead of a clear missing-dependency error. This is a behavior change for all users.

**Suggested fix:** split this into its own PR; if kept, emit a warning so the command's absence isn't silent.

nemo_skills/pipeline/utils/declarative.py
```diff
@@ -579,13 +581,27 @@ def _prepare_command(self, command, cluster_config: Dict) -> Tuple[run.Script, D
...
+        submitted to a *remote* cluster whose image and Python version need not match
+        the driver's, so the driver-side absolute install path (e.g.
+        ``/usr/local/lib/python3.10/dist-packages``) will not exist there. In that
+        case rewrite to relative (cwd) paths instead -- mirroring the add_task path --
+        so the command runs from the job's working directory and ``nemo_skills``
+        resolves via the baked image's ``PYTHONPATH``.
+        """
         nemo_repo = get_registered_external_repo("nemo_skills")
-        if nemo_repo is None:
+        is_ray = (self.cluster_config.get("backend") or {}).get("name") == "ray"
```
gwarmstrong: **Inconsistent backend parsing crashes on string-form `backend` (low/medium).**

This site does `(self.cluster_config.get("backend") or {}).get("name")`, but `_normalize_backend_config` (`backends.py`) and `generate.py` both explicitly accept `backend` as a **string**.

A config with `backend: ray` (or `backend: default`) as a bare string plus `executor: none` raises `AttributeError: 'str' object has no attribute 'get'`. The pure no-`backend` default path is safe, but the PR's own supported string-form input is not.

**Suggested fix:** route this through the same `_normalize_backend_config` helper used elsewhere.

nemo_skills/pipeline/utils/exp.py
```diff
@@ -890,10 +914,34 @@ def add_server_tasks():
             )
             commands[idx] = commands[idx].replace("/nemo_run/code", "./")
 
-    if with_ray and cluster_config["executor"] == "slurm":
-        metadata = {"use_with_ray_cluster": True}
-    else:
-        metadata = None
+    backend = get_execution_backend(cluster_config, with_ray=with_ray)
+    first_command_image = command_images[0] if command_images else None
+    should_use_with_ray_cluster = bool(with_ray or getattr(backend, "name", "") == "ray")
```
gwarmstrong: **`with_ray=True` + non-slurm executor now crashes (low; legacy ray flag).**

Old code gated metadata on `with_ray and executor == "slurm"`, returning `None` otherwise. Routing through `stage_metadata` dropped the executor check, so `with_ray=True` + `executor: local/none` now attaches `use_with_ray_cluster: True`, which nemo-run rejects (`packaging.py:236` asserts `SlurmExecutor`).

This is on the legacy `with_ray` flag (so arguably within the Ray path), but it's a regression for that combo where the old code silently no-op'd. Worth restoring the `executor == "slurm"` guard for the legacy-flag case.

nemo_skills/pipeline/utils/exp.py
```diff
@@ -757,6 +778,7 @@ def add_server_tasks():
             with temporary_env_update(cluster_config, main_env_updates):
                 cur_cmd = install_packages_wrap(cur_cmd, installation_command)
                 commands.append(cur_cmd)
+                command_images.append(resolve_container_image(cur_container, cluster_config))
```
gwarmstrong: **Redundant `resolve_container_image` on the shared path (low / efficiency).**

`command_images` is populated by calling `resolve_container_image` for every server/main/sandbox command (also lines 743 and 831), but `get_executor` already resolves the same container (`exp.py:279`), and the list is only consumed by the Ray queue. For normal image strings it's a cheap duplicate; for `dockerfile:` specs it triggers a redundant `docker image inspect` per command on the non-Ray path.

**Suggested fix:** reuse the already-resolved image from the executor, or only build `command_images` when the Ray backend is active.

nemo_skills/pipeline/utils/backends.py
```diff
@@ -0,0 +1,337 @@
...
+            exp.cancel(handle)
+            cancelled_jobs.append(task_name)
+
+        return cancelled_jobs
+
+
+# ---------------------------------------------------------------------------
+# RayBackend lives in ray_backend.py – imported here for backwards compat.
+# ---------------------------------------------------------------------------
+from nemo_skills.pipeline.utils.ray_backend import RayBackend  # noqa: E402  # re-exported
```
gwarmstrong: **Nit:** this module-level `from …ray_backend import RayBackend` means the default path unconditionally imports the ~900-line `ray_backend` module. It's import-safe today (lazy `ray`), but lazy-loading `RayBackend` only when the backend resolves to `ray` would fully decouple it and protect the default path from future top-level additions in that module.

gwarmstrong (changes requested):
Thanks for the contribution. I focused this review specifically on whether the new Ray backend stays off the non-Ray code path (i.e., no behavior change for users who never opt in). Mostly it does — no hard runtime dependency on `ray` is added (the `ray` / `JobSubmissionClient` imports are all lazy), and the default-backend resolution faithfully reproduces old behavior in most spots. But a few ray-motivated refactors leaked onto the shared path.

**Requesting changes** — a few of which are blocking: they are genuine behavior changes on the **non-Ray** path. See the inline comments for specifics.

nemo_skills/pipeline/utils/exp.py
nicgupta-nvidia: Restored the dry‑run short‑circuit: `run_exp` gates the mount check behind `if not dry_run`, so `--dry_run` is offline‑safe again for all the non‑Ray slurm callers. Added `test_run_exp_dry_run_skips_remote_mount_check`.

nemo_skills/pipeline/utils/declarative.py
nicgupta-nvidia: Scoped to `backend.name=="ray"`; the default backend passes deps through unchanged, so nemo‑run still fails loud on an unresolved dep. Reframed the existing test as the Ray case + added a default‑backend "kept" test.

nemo_skills/pipeline/cli.py
nicgupta-nvidia: Agreed. Reverted to the plain import; `cli.py` is now byte‑identical to `origin/main`.

nemo_skills/pipeline/utils/exp.py
nicgupta-nvidia: Only resolved when a Ray backend is active — no docker inspect per command on the non‑Ray path.

nemo_skills/pipeline/utils/backends.py
nicgupta-nvidia: RayBackend now imported lazily (local import + PEP 562 `__getattr__` + `TYPE_CHECKING`); verified the default path no longer imports `ray_backend`.

nemo_skills/pipeline/utils/exp.py
nicgupta-nvidia: Restored the guard — though it turned out the obvious fix wasn't enough on its own: legacy `with_ray` resolves to a `RayBackend (name=="ray")` through the compat path, so gating just the `with_ray` term still left the flag on. It's now `executor == "slurm"` and (`with_ray or backend.name == "ray"`). The precreated Jobs-API path is unaffected since `stage_metadata` already forces embedding off there. Added a test (confirmed it fails on the old behavior) and mirrored the same gate in `declarative.py`.

nemo_skills/pipeline/utils/declarative.py
nicgupta-nvidia: Fixed. Pulled the parsing into a shared `get_backend_name()` helper (built on `_normalize_backend_config`) and used it here and in `generate.py`, so the bare-string form works and the duplication is gone.

tests/test_backends.py
```diff
@@ -0,0 +1,437 @@
...
+    cluster_config = {
+        "executor": "slurm",
+        "backend": {"name": "ray", "dashboard_url": "http://ray-head:8265"},
+    }
+
+    backend = get_execution_backend(cluster_config)
+    assert backend.precreated_cluster is True
+
+    metadata = backend.stage_metadata(container_image="nvcr.io/nvidia/pytorch:25.02-py3")
+    assert not (metadata or {}).get("use_with_ray_cluster")
```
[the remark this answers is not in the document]
nicgupta-nvidia: Addressed in 5f79c5f3 — both assertions (here and the ~216 site) now check `"use_with_ray_cluster" not in (metadata or {})`, so they fail if the key is ever reintroduced with any value (incl. `False`/`None`).

One deviation from the suggestion: I kept the `or {}` guard rather than `not in metadata`. The base `ExecutionBackend.stage_metadata` returns `None` by contract (backends.py:108), so a bare `in` would `TypeError` if the base backend is ever exercised. Both call sites here run on `RayBackend` (always a dict), so this is just future-proofing the absence check. `tests/test_backends.py` → 25 passed; pre-commit clean.

nicgupta-nvidia:
The CPU-tests failure on 31cf2e2 is not from this branch: `vedas==0.0.1` was deleted from PyPI, which breaks the sandbox image build (Build Images step — no tests ran) for every PR/branch repo-wide. Fix: #1509 (3-line requirements removal). Will re-run CI here once it merges.

nicgupta-nvidia:
Merge-blocker status update (September 14):

- This branch contains current `main` (`bcf059af55c20a89f797724598f9908d126153e6`) and is 0 commits behind, with no merge conflicts.
- All 15 review threads are resolved. The requested non-Ray compatibility fixes are implemented in the current code.
- Required checks are green: [CPU tests](https://github.com/NVIDIA-NeMo/Skills/actions/runs/34642971436) (795 passed), pre-commit, copyright, and DCO. The GPU check is intentionally skipped by its workflow rules.
- GitHub reports the current head (`381120415cabeb9d6ba18da1b2e854ceb0465716`) as mergeable.

@gwarmstrong The earlier changes-requested review is now the only remaining merge blocker. A fresh review is already requested; could you please re-review the current head when available?

Issue #1035, opened 2025-11-07
Need to find a way to customize judge setup only for part of benchmarks for groups


Kipok:
more details and real use-case we are missing is here https://github.com/NVIDIA-NeMo/Skills/pull/1022#discussion_r2500971745

PR #1481, opened 2026-06-10
Speed up local vLLM ASR client requests

DongjiGao:
## Summary
- Replace httpcore's async connection-pool assignment hot path (httpcore 1.0.9 is O(N²+N·M)) with a single-pass O(N+M) rewrite ported from encode/httpcore#1035; HTTP/2 pools fall back to the original implementation. This also fixes the two logic bugs that PR documents: the `max_keepalive_connections` idle-count miscount, and the HTTP/1.1 double-assign that can raise `ConnectionNotAvailable`.
- Send local vLLM multimodal audio as `file://` audio URLs by default for local `audio_url` requests, avoiding inline base64 payloads (base64 retained for external APIs and as an explicit fallback).
- Add a shared `make_audio_url_content_block` helper and expand vLLM audio tests (file:// vs base64 content, plus direct `VLLMMultimodalModel.__init__` tests for `audio_as_path` defaulting/validation).

## Test plan
- `python3 -m py_compile nemo_skills/inference/generate.py nemo_skills/inference/model/vllm_multimodal.py nemo_skills/inference/model/audio_utils.py`
- `pytest tests/test_vllm_audio.py` → 16 passed; `ruff check` + `ruff format --check` clean.
- httpcore rewrite validated for assignment-decision equivalence + liveness across 20k randomized pool states (vs the original 1.0.9 implementation).
- Full 8-chunk ASR leaderboard end-to-end: WER (%) 6.21, no throughput regression; controlled traced runs show the patch applied (`max_queued=1`, `max_conn=512`). Residual wall-clock variance traces to server-startup / chunk-start skew, not the request path.

## Summary by CodeRabbit

* **New Features**
  * Added audio_as_path option to multimodal models to represent local audio as file:// URLs instead of inline base64.
  * Added support for producing audio URL content blocks for URL-based audio inputs.
  * Improved async connection pooling to make request assignment more efficient.

* **Tests**
  * Expanded tests to cover audio_as_path defaults, fallback to base64, validation for file:// and data:audio values, and related edge cases.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -273,6 +273,79 @@ cs = hydra.core.config_store.ConfigStore.instance()
 cs.store(name="base_generation_config", node=GenerationTaskConfig)
 
 
+def _patch_httpcore_connection_pool_assignment() -> None:
+    """Patch httpcore's async connection assignment hot path.
+
+    httpcore 1.0.9 repeatedly materializes connection-state lists while
+    assigning queued requests to connections. At high OpenAI-compatible
+    concurrency this can dominate the client asyncio loop. Keep the same
+    request-assignment order, but use scalar checks/generators instead of
+    rebuilding lists on every pass.
+    """
+    try:
+        import httpcore._async.connection_pool as async_connection_pool
+    except ImportError:
+        LOG.debug("httpcore is not installed; skipping async connection-pool patch")
+        return
+
+    assign = async_connection_pool.AsyncConnectionPool._assign_requests_to_connections
+    if getattr(assign, "_nemo_skills_fast_assign", False):
+        return
+
+    def _fast_assign_requests_to_connections(self):
+        closing_connections = []
+
+        # First clean up closed/expired/surplus-idle connections. The original
+        # implementation builds an N-element list for every idle connection just
+        # to take its len(); len(self._connections) is behavior-equivalent.
+        for connection in list(self._connections):
+            if connection.is_closed():
+                self._connections.remove(connection)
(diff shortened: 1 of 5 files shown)


DongjiGao:
@pzelasko

DongjiGao:
Addressed in eca6b2bb (patch) and 701cbf0e (tests).

Rather than only fix the `max_keepalive` count, I replaced the assignment routine with a single-pass O(N+M) port of encode/httpcore#1035. This resolves **both** issues that PR documents:

- the `max_keepalive_connections` miscount — it now enforces the limit on the count of *idle* connections, not `len(self._connections)`; and
- the HTTP/1.1 double-assign (`ConnectionNotAvailable`) bug — each connection is popped from the available bucket as it is assigned, so one connection is never handed to two requests in a single pass.

HTTP/2 pools fall back to the original implementation, since the single-pass path intentionally omits the per-connection stream-capacity tracking #1035 adds for multiplexing.

Validated for assignment-decision equivalence + liveness across 20k randomized pool states, and end-to-end on the 8-chunk ASR leaderboard (WER 6.21, no throughput regression).

For the test nitpick: added direct `VLLMMultimodalModel.__init__` tests covering `audio_as_path` defaulting (by mode/format) and the validation that rejects path mode for external APIs / non-`audio_url` formats.

pzelasko (approved):
LGTM, very welcome speedup, great work!

DongjiGao:
Recreated as in-repo branch PR #1487 so CI can run (fork PRs don't receive CI secrets). Closing this fork PR in favor of #1487.

PR #1487, opened 2026-06-15
Speed up local vLLM ASR client requests

DongjiGao:
@/tmp/tmp.HobmUsFmEc

## Release Notes

**New Features**
- Added `audio_as_path` option to vLLM multimodal models for more efficient handling of local audio files using URL references instead of inline encoding
- New utility function for creating audio URL content blocks for API compatibility
- Comprehensive validation for audio formats with intelligent configuration defaults and automatic fallbacks

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import importlib
 import json
 import logging
 import random
@@ -290,9 +291,17 @@ def _patch_httpcore_connection_pool_assignment() -> None:
     this fast path omits.
     """
     try:
-        import httpcore._async.connection_pool as async_connection_pool
+        httpcore = importlib.import_module("httpcore")
+        async_connection_pool = importlib.import_module("httpcore._async.connection_pool")
     except ImportError:
         LOG.debug("httpcore is not installed; skipping async connection-pool patch")
+        return
+
+    if httpcore.__version__ != "1.0.9":
+        LOG.warning(
+            "Skipping httpcore AsyncConnectionPool patch: validated only on httpcore 1.0.9, found %s",
+            httpcore.__version__,
+        )
         return
 
     original_assign = async_connection_pool.AsyncConnectionPool._assign_requests_to_connections
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -273,6 +273,79 @@ cs = hydra.core.config_store.ConfigStore.instance()
 cs.store(name="base_generation_config", node=GenerationTaskConfig)
 
 
+def _patch_httpcore_connection_pool_assignment() -> None:
+    """Patch httpcore's async connection assignment hot path.
+
+    httpcore 1.0.9 repeatedly materializes connection-state lists while
+    assigning queued requests to connections. At high OpenAI-compatible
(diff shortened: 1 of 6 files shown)


Slyne:
+ comparison with vllm benchmark script
Screenshot 2026-06-15 at 12 17 58 PM

nemo_skills/inference/model/vllm_multimodal.py
```diff
@@ -105,6 +109,11 @@ def __init__(
         if audio_format not in ("audio_url", "input_audio"):
             raise ValueError(f"Unsupported audio_format '{audio_format}'. Use 'audio_url' or 'input_audio'.")
         self.audio_format = audio_format
+        if audio_as_path is None:
+            audio_as_path = not self._external_api_mode and self.audio_format == "audio_url"
```
KunalDhawan: This defaults `audio_as_path=True` for all local audio_url requests, but vLLM rejects `file://` media unless the server is launched with `--allowed-local-media-path` (it's disabled by default for security). `serve_vllm.py` (the command at L44–53) doesn't pass that flag, and it doesn't appear anywhere in the repo, so the default path here will fail at request time on a stock server, the same local-vLLM ASR path this PR is named for. The green e2e run likely had a server with the flag set manually, which hides this.
Could we either (a) have `serve_vllm.py` add `--allowed-local-media-path` (pointed at `data_dir`) when audio is in use, or (b) flip the default to `audio_as_path=False` so it's strictly opt-in, with the server requirement documented? Right now turning it on by default puts the burden on every user to also reconfigure the server.

nemo_skills/inference/generate.py
```diff
@@ -273,6 +273,127 @@ def _get_disallowed_params(self):
...
+    the original because they need per-connection stream-capacity tracking that
+    this fast path omits.
+    """
+    try:
+        import httpcore._async.connection_pool as async_connection_pool
+    except ImportError:
+        LOG.debug("httpcore is not installed; skipping async connection-pool patch")
+        return
+
+    original_assign = async_connection_pool.AsyncConnectionPool._assign_requests_to_connections
```
KunalDhawan, line 318: This function depends on `httpcore 1.0.9` internals `(_connections, _requests, _http2, _max_connections, _max_keepalive_connections, pool_request.is_queued(), connection.can_handle_request())`. The only safeguard is the `_nemo_skills_fast_assign` double-patch marker, but there's no version check. If httpcore bumps, this will either silently change behavior or raise inside the event loop. Could we gate the patch on `httpcore.__version__` (patch only the validated version, log-and-skip otherwise) or add some other guards?

KunalDhawan (changes requested):
Great work @DongjiGao! Added some comments below, other than that LGTM

nemo_skills/inference/generate.py
DongjiGao: Thanks @KunalDhawan , both points make sense.

- For the httpcore patch, I gated it to the validated version only (`httpcore==1.0.9`) and log/skip otherwise, so we don't patch unknown private internals after a dependency bump. I also added a small unit test covering both the apply and skip cases.

- For `audio_as_path`, agreed that defaulting to `file://` is unsafe without `--allowed-local-media-path`. I flipped the default to `False`, kept it opt-in for local `audio_url`, and updated the docstring/tests to call out the server requirement.

PR #1502, opened 2026-07-01
High-throughput async inference: native aiohttp fast-path, scalable async loop, and request-routing keys

gwarmstrong:
## Summary

Makes `ns generate` scale cleanly to high concurrency (5k+ in-flight requests) against OpenAI-compatible endpoints. At high concurrency the old litellm/httpx dispatch was the GIL-bound bottleneck (~95% of a single core in the async loop, per py-spy); this branch moves the hot path onto a C-extension-backed aiohttp transport and removes the per-completion write/postprocess serialization in the generate loop. It also adds optional, provider-agnostic request-routing keys that a proxy/gateway may use (a plain endpoint ignores them).

The fast path is the **default** and is correct for current vLLM (reasoning field rename) and streaming.

## What's included

**Native AsyncOpenAI + aiohttp fast-path** (`model/base.py`)
- Routes non-streaming **and** streaming chat through a native `AsyncOpenAI` client using the `DefaultAioHttpClient` (aiohttp) transport, bypassing litellm's httpx dispatch.
- **Default-on** for OpenAI-compatible providers, gated by a new `SUPPORTS_NATIVE_OPENAI` class attribute → enabled for `openai`/`vllm`/`sglang`; `azure`, `gemini`, `megatron`, and multimodal stay on litellm. Disable with `NEMO_SKILLS_OPENAI_AIOHTTP=0`.
- Reasoning content is read from **both** `.reasoning` (newer vLLM) and `.reasoning_content` (older) — litellm normalized this for us, but the native path sees vLLM's raw field.
- Transient socket-error retries with backoff; extended httpx client TTL patch to survive long high-concurrency runs.

**Scalable async loop** (`inference/generate.py`)
- Bounded worker pool (memory O(in_flight), not O(total inputs)) with a rate-limited connection ramp (`ramp_rate_per_sec`, default 4000) to avoid overflowing the kernel accept queue on prime.
- Dedicated batched **writer task** (flush per batch) instead of a per-completion `fout` lock.
- Separate **postprocess/eval processor pool** (`postprocess_workers`, default 16) so a lane frees its slot the instant the HTTP response returns.
- Installs `uvloop` when available (`NEMO_SKILLS_DISABLE_UVLOOP=1` to opt out).

**Request-routing keys** (`model/base.py`) — optional, provider-agnostic; a plain endpoint ignores them:
- `X-Request-Id`: a per-request id for idempotency/tracing, kept stable across the OpenAI SDK retry loop; distinct calls get distinct ids.
- `prompt_cache_key`: the standard OpenAI/vLLM prefix-cache hint, opt-in via `cache_key=`. Requests sharing a value hint a shared prompt prefix. Kept separate from `X-Request-Id` (shared vs per-request).

**Misc**
- Dumps the full `response.usage` dict + reasoning-token breakdown into the output.
- Auto-passthrough of `NEMO_SKILLS_*` env vars to the inner python launched by the pipeline (`pipeline/utils/cluster.py`).
- `core/requirements.txt`: `openai` → `openai[aiohttp]` so the transport is actually installed (per CONTRIBUTING, inference deps live here).

## Config / env knobs

| Knob | Default | Purpose |
|---|---|---|
| `NEMO_SKILLS_OPENAI_AIOHTTP` | `1` | Native fast-path; `0` forces litellm |
| `NEMO_SKILLS_OPENAI_AIOHTTP_LIMIT` | `65536` | aiohttp max connections |
| `NEMO_SKILLS_TRANSIENT_RETRIES` | `3` | Socket-level retry budget |
| `NEMO_SKILLS_DISABLE_UVLOOP` | unset | Opt out of uvloop |
| `max_concurrent_requests` (cfg) | `512` | Script-level concurrency cap |
| `ramp_rate_per_sec` (cfg) | `4000` | Connection ramp on prime |
| `postprocess_workers` (cfg) | `16` | Postprocess/eval pool size |

## Testing

**Unit** (`tests/test_prompt_cache_routing.py`, `tests/test_native_openai_path.py` — 15 tests): routing-key idempotency/affinity decoupling; reasoning extraction from both field names; stream + non-stream parsing; empty-`choices` chunk guard; per-provider native eligibility; env-var default/opt-out.

**Live parity** (real model on an OpenAI-compatible endpoint, `nvidia/nemotron-3-nano-30b-a3b`):
- native vs litellm (`AIOHTTP=0`), × non-stream × stream → **identical** final answer; `reasoning_content` captured on all four; native transport active only by default and off with `AIOHTTP=0`.
- Streaming yields well-formed chunks (`generation`/`reasoning_content`/`finish_reason`).
- Tool-calling stream via native path → `finish_reason=tool_calls`, reassembled `get_weather({"city":"Paris"})` (the exact chunk shape `ToolCallingWrapper._stream_single` consumes).

> Reviewer note: this touches the core inference loop — worth running the SLURM GPU CI (add the **run GPU tests** label).

## Known non-blocking notes
- Streaming `num_generated_tokens` is `None` on both native and litellm paths (unchanged; `ToolCallingWrapper` counts via tokenizer). I did not enable `stream_options.include_usage`, but added the empty-`choices` guard so it's safe to turn on later.
- The native aiohttp session isn't explicitly closed at process exit (same as litellm's existing `aclient_session`); harmless since the process exits after a run.

## Summary by CodeRabbit

* **New Features**
  * Improved async generation pipeline with bounded in-flight concurrency, smoother ramp-up, and configurable background postprocessing.
  * Native OpenAI-compatible async fast-path for eligible providers, with per-call request id headers and optional prompt-cache routing.
  * Optional uvloop acceleration for async generation.
  * Automatic forwarding of `NEMO_SKILLS_*` environment variables into jobs (excluding SSH tunnel controls).
* **Bug Fixes**
  * More resilient retries, streaming/response parsing (including reasoning + usage), and safer async output restoration.
  * Multimodal and structured-output requests reliably fall back when native mode isn’t appropriate.
* **Tests**
  * Added regression and unit tests for native fast-path selection, routing keys, env forwarding, and async ordering integrity.
* **Chores**
  * Updated the OpenAI dependency to include the aiohttp extra.

### DIFF
--- a/nemo_skills/inference/generate.py
+++ b/nemo_skills/inference/generate.py
@@ -22,16 +22,8 @@ import subprocess
 import sys
 import time
 
-# uvloop is a drop-in replacement for asyncio's default event loop
-# that's ~2-4x faster on CPU-bound asyncio workloads. ns generate's
-# async_loop at high concurrency (5k+ lanes) becomes CPU-bound on
-# the event loop's per-task scheduling overhead -- not on actual
-# Python work in user code. Installing uvloop closes that gap so
-# the bounded pool can refill at the rate the upstream can sustain
-# instead of the rate one Python thread can scheduler-switch.
-#
-# Set NEMO_SKILLS_DISABLE_UVLOOP=1 to opt out (useful if a tool in
-# the stack assumes selector loop semantics; rare).
+# Prefer uvloop when available to reduce event-loop scheduling overhead.
+# Set NEMO_SKILLS_DISABLE_UVLOOP to use the default asyncio event loop.
 if not os.environ.get("NEMO_SKILLS_DISABLE_UVLOOP"):
     try:
         import uvloop
@@ -144,46 +136,11 @@ class GenerationTaskConfig:
     # maximum number of concurrent requests to the server for the async loop
     # if sync loop is used, this is the batch size
     max_concurrent_requests: int = 512
-    # Rate at which the async loop opens NEW HTTP connections during
-    # initial prime (target connections per second). The bounded
-    # worker pool would otherwise spawn `max_concurrent_requests`
-    # tasks in a tight loop, each opening a TCP connection nearly
-    # simultaneously -- the kernel accept queue on the receiving end
-    # (typically capped at net.core.somaxconn=4096) can briefly
-    # overflow and SYN-drop the excess. Limiting the ramp rate so
-    # each ~ramp_rate/4-connection burst is followed by a short pause
-    # keeps the accept queue draining cleanly.
-    #
-    # Default 4000/sec accounts for ns generate's per-task pre-work
-    # (format_prompt, build litellm request, semaphore acquires) --
-    # the time from asyncio.create_task to "TCP request fired" is
-    # significant, so the configured task-creation rate has to be
-    # higher than the SYN rate you actually want at the kernel.
-    # Each burst is still ramp_rate/4 SYNs (1000 at default), well
-    # under net.core.somaxconn=4096.
-    #
-    # Set to 0 (or negative) to disable ramping and prime
-    # instantaneously. The transient-retry logic in
-    # BaseModel.generate_async would still cover any resulting SYN
-    # drops; the ramp just makes them not happen in the first place.
+    # Target task-creation rate while initially filling the async worker pool.
+    # Tasks are created in short batches to limit connection bursts. Values <= 0 disable ramping.
     ramp_rate_per_sec: int = 4000
-    # Number of background coroutines that postprocess + (optionally)
-    # evaluate completed datapoints. The async loop's bounded worker
-    # pool used to keep each lane "alive" (occupying a slot) until
-    # postprocess + evaluate + queue.put all completed -- which
-    # serialized those steps with the next HTTP refill. For
-    # reasoning-heavy workloads the per-lane post-HTTP work is small
-    # in absolute terms (~ms per record) but multiplied across
-    # thousands of lanes it can keep the bounded-pool slot occupied
-    # while real HTTP capacity sits idle. Moving postprocess + eval
-    # to a separate processor pool lets the lane exit the moment
-    # generate_async returns, so the bounded pool refills with the
-    # next datapoint immediately.
-    #
-    # Default 16 is enough headroom for synchronous postprocess (no
-    # eval) at >3k records/sec. If you enable an LLM-judge evaluator
-    # inside generate, raise this to match the eval concurrency you
-    # want.
+    # Number of background coroutines that postprocess and optionally evaluate completed datapoints.
+    # Increase this when per-datapoint evaluation benefits from additional concurrency.
     postprocess_workers: int = 16
     # chunk the dataset into equal sized parts and index into them
     num_chunks: int | None = None  # if specified, will split the data into chunks and only generate for one chunk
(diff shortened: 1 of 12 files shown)


nemo_skills/inference/model/base.py
```diff
@@ -289,17 +360,54 @@ async def generate_async(
             "response_format": response_format,
         }
 
-        # TODO: remove this after we no longer use gpt-oss or it's fixed in vllm
+        # Keep the targeted bad-request retry budget separate from retries for
+        # transient connection failures.
         max_retries = 2
```
Kipok, line 365: could you clean this up please - I think current retry count is never incremented and while loop is basically while True and there is a separate counter for transient retries. Should either have explicity while True and remove these outdated vars or just use retry_count without adding transient_count like it was before

nemo_skills/inference/model/base.py
```diff
@@ -289,17 +360,54 @@ async def generate_async(
...
+                        # litellm.
+                        #
+                        # Exception: a pydantic BaseModel response_format (structured
+                        # output) must go through litellm. The native SDK's
+                        # chat.completions.create() rejects a BaseModel class
+                        # ("must use .parse() instead"); litellm accepts it and
+                        # converts it to a JSON schema. (dict/json_schema
+                        # response_formats are fine on the native path.)
+                        rf = request_params.get("response_format")
+                        needs_litellm_response_format = inspect.isclass(rf) and issubclass(rf, PydanticBaseModel)
```
Kipok, line 388: this looks a bit fragile as a logic for routing to litellm, but I guess we will retire all this code soon anyway, so hopefully ok for now. And gym doesn't use litellm at all

nemo_skills/inference/generate.py
```diff
@@ -15,11 +15,22 @@
...
 import shutil
 import subprocess
 import sys
 import time
+
+# Prefer uvloop when available to reduce event-loop scheduling overhead.
+# Set NEMO_SKILLS_DISABLE_UVLOOP to use the default asyncio event loop.
+if not os.environ.get("NEMO_SKILLS_DISABLE_UVLOOP"):
+    try:
+        import uvloop
```
Kipok, line 29: does this need to be added to requirements?

nemo_skills/inference/generate.py
```diff
@@ -893,15 +976,165 @@ async def async_loop(self, data):
...
+                        return
+                    output, data_point = item
+                    try:
+                        await self.postprocess_single_output(output, data_point)
+                        if self.should_run_evaluation and self.evaluator:
+                            output = await self.evaluate_single_datapoint({**data_point, **output})
+                    except Exception:
+                        pos = data_point.get(self.cfg.async_position_key, "?")
+                        LOG.exception("postprocess/evaluate failed for async_pos=%s", pos)
+                    await self._output_queue.put((output, data_point))
```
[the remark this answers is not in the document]
Kipok: @gwarmstrong not sure if this is correct but if it is, we should probably fail on exceptions

nemo_skills/inference/generate.py
```diff
@@ -893,15 +915,119 @@ async def async_loop(self, data):
...
+            writer_task = asyncio.create_task(_writer())
+
+            # Bound both request concurrency and allocated task state to max_in_flight.
+            # Each task handles its own error so other datapoints continue processing.
+            max_in_flight = self.cfg.max_concurrent_requests
+
+            async def _safe_one(dp):
+                try:
+                    await self._generate_and_save_datapoint(dp, data, fout, pbar)
+                except Exception as e:
```
Kipok, line 980: I guess similar here, why do we want to catch all exceptions?

PR #1515, opened 2026-08-14
senior-swe-bench benchmark evaluation integration

wasiahmad:
## Summary
- Add **DeepSWE** and **Senior SWE-Bench** as Harbor-format coding-agent benchmarks in NeMo-Skills.
- Both reuse existing SWE-bench agent frameworks (`mini_swe_agent` / `swe_agent` / `openhands` / `gold_patch`) and grade with each task’s Harbor `tests/test.sh` (not the SWE-bench harness).
- Extract shared Harbor helpers (`harbor_utils.py`) for tests-dir resolution and reward loading; Senior SWE-Bench applies `model.patch` under `/repo/$REPO_NAME` before the full SSB verifier (verify + LLM judges + validation).

## Details
- **Prepare:** `ns prepare_data deep-swe|senior-swe-bench --data_dir=...` clones the Harbor task repo, writes JSONL + `tasks/`, then deletes the temporary checkout (`REQUIRES_DATA_DIR=True`).
- **DeepSWE:** scores via Harbor `reward.json` (`reward` / `f2p` / `p2p` / `partial`); network-isolates a small set of verifiers; resolves LLM hostnames before container launch.
- **Senior SWE-Bench:** scores via full SSB `test.sh`; injects judge API keys through Apptainer `--env`; treats empty `reward.txt` as an invalid trial (not a solve).
- Docs in `docs/evaluation/code.md`; unit tests for reward parsing; both benchmarks excluded from heavy GPU bulk CI.

## Test plan
- [ ] `ns prepare_data deep-swe --data_dir=... --container_formatter=...`
- [ ] `ns prepare_data senior-swe-bench --data_dir=... --container_formatter=...`
- [ ] Smoke Senior SWE-Bench grading: `ns eval --benchmarks=senior-swe-bench --data_dir=... ++agent_framework=gold_patch` (with judge API keys set)
- [ ] Smoke DeepSWE: `ns eval --benchmarks=deep-swe --data_dir=... ++agent_framework=gold_patch` (or `mini_swe_agent`)
- [ ] Confirm metrics.json includes `issues_resolved` / `reward` (and `invalid_trial` for SSB)
- [ ] Run `tests/test_deep_swe.py` and `tests/test_senior_swe_bench.py`


nemo_skills/inference/eval/swebench.py
```diff
@@ -46,12 +48,104 @@ class SupportedAgentFrameworks(str, Enum):
...
+        model_entry,
+        {
+            "id": model,
+            "name": model,
+            # OpenCode only forwards an agent temperature for models that declare
+            # temperature support.
+            "temperature": True,
+            "tool_call": True,
+            "limit": {
+                "context": 262144,
```
ludwig-n: does this limit context length? if so, 262144 is too low. i don't think opencode should control this at all, can we disable this limit? (or set it to like 1 billion or something) and control context length with vllm args instead

nemo_skills/inference/eval/swebench.py
```diff
@@ -46,12 +48,104 @@ class SupportedAgentFrameworks(str, Enum):
...
+                "output": (
+                    tokens_to_generate if tokens_to_generate is not None else OPENCODE_DEFAULT_OUTPUT_TOKEN_MAX
+                ),
```
ludwig-n, line 227: the other frameworks don't limit output length, so ideally by default there should be no limit here as well, though to be fair 131072 is a lot

nemo_skills/inference/eval/swebench.py
```diff
@@ -46,12 +48,104 @@ class SupportedAgentFrameworks(str, Enum):
...
+            },
+        },
+    )
+    if top_k is not None:
+        model_entry.setdefault("options", {})["top_k"] = top_k
```
ludwig-n: we need to pass `extra_body` here too, otherwise we can't control `enable_thinking` or `reasoning_effort` for example. if opencode doesn't support this natively, then we'll need a fork i guess

nemo_skills/inference/eval/swebench.py
```diff
@@ -46,12 +48,104 @@ class SupportedAgentFrameworks(str, Enum):
...
+    )
+    models = nemo.setdefault("models", {})
+    model_entry = models.get(model, {}) if isinstance(models.get(model), dict) else {}
+    _deep_merge_dicts(
+        model_entry,
+        {
+            "id": model,
+            "name": model,
+            # OpenCode only forwards an agent temperature for models that declare
+            # temperature support.
```
ludwig-n, line 220: does this mean it won't be passed for some models? can we force it to always be passed? (same with other sampling parameters)

nemo_skills/inference/eval/swebench.py
```diff
@@ -896,6 +1070,88 @@ async def _run_openhands(self, data_point, api_base):
...
+            f"echo {shlex.quote(config_json)} >/root/.config/opencode/opencode.json && "
+            "cd /testbed && "
+            "git config --global --add safe.directory /testbed && "
+            "git config --global user.email opencode@nemo-skills.local && "
+            "git config --global user.name OpenCode && "
+            "START_COMMIT=$(git rev-parse HEAD) && "
+            f"mkdir -p /trajectories_mount/trajectories/{shlex.quote(instance_id)} && "
+            f"opencode --model={shlex.quote(model_arg)} run --format=json "
+            f"--thinking --dangerously-skip-permissions -- {shlex.quote(instruction)} "
+            f"</dev/null > /trajectories_mount/trajectories/{shlex.quote(instance_id)}/opencode.txt 2>&1 && "
```
ludwig-n: this redirects both normal output (JSONL format) and stderr into the same file, this means in case of an error the file will be a mix of JSONL and some random plain text tracebacks. perhaps we should redirect stderr into another file?

side note: if we can export a trajectory in openai messages format directly from opencode (e.g. a copy of the query sent to vLLM on the last turn containing the whole conversation history), that would be great, right now opencode outputs its own format which we need to convert separately

nemo_skills/inference/eval/swebench.py
wasiahmad: Default for opencode is 32k. There has to be a number, it cannot be None, if it is None, then it will be 32k. It failed 1-2 rollout generation when I tested Kimi-K3.

nemo_skills/inference/eval/swebench.py
wasiahmad: The hardcoded 262144 is OpenCode’s client-side context metadata, used for output budgeting and compaction, not server enforcement. So, OpenCode compacts as if the limit were 262144, the extra server capacity is wasted.

Your point is valid too. Let me see how can I tackle it.

nemo_skills/inference/eval/swebench.py
wasiahmad: `"temperature": True` is a capability declaration, not the sampling value. It tells OpenCode: “this model supports temperature; include the agent’s configured temperature in requests.”

Current behavior:

- temperature: always passed
- top_p: always passed
- top_k: passed when explicitly configured
- tokens_to_generate: always applied as the output limit
- tool_call: true: unrelated to sampling; advertises tool-call support

Other fields are not currently wired for OpenCode:

- min_p
- repetition_penalty
- random_seed
- top_logprobs
- extra_body

nemo_skills/inference/eval/swebench.py
```diff
@@ -117,7 +118,7 @@ def build_opencode_config(
             "temperature": True,
             "tool_call": True,
             "limit": {
-                "context": 262144,
+                "context": context_window,
```
wasiahmad, line 224: @ludwig-n we can now control it via `++opencode_context_window`

nemo_skills/inference/eval/swebench.py
```diff
@@ -125,8 +126,13 @@ def build_opencode_config(
             },
         },
     )
-    if top_k is not None:
-        model_entry.setdefault("options", {})["top_k"] = top_k
+    if extra_body or top_k is not None:
+        model_options = model_entry.setdefault("options", {})
+        if not isinstance(model_options, dict):
+            raise ValueError("OpenCode model options must be a dictionary.")
+        _deep_merge_dicts(model_options, copy.deepcopy(extra_body))
```
wasiahmad, line 235: @ludwig-n we can now pass the following to OpenCode.

```
"++inference.extra_body.chat_template_kwargs.enable_thinking=$ENABLE_THINKING"
"++inference.extra_body.chat_template_kwargs.reasoning_effort=$REASONING_EFFORT"
```

PR #1532, opened 2026-09-16
fix: raise a clear error on empty eval group config files

### LINKED ISSUE #1497: fix: yaml.safe_load returns none for empty yaml files, propagating as none to callers expecting dict
## Description

`yaml.safe_load()` returns `None` when given an empty YAML file (or one with only comments). `load_config` returns this `None` directly, and `get_eval_group` passes it through unchanged. Any caller doing `config["key"]` will get `TypeError: 'NoneType' object is not subscriptable`. This is a silent misconfiguration bug — an empty or mis-rotated config file produces no useful error message, just a cryptic downstream crash.

**Severity**: `high`
**File**: `nemo_skills/evaluation/utils.py`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

Ashfaqbs:
## Summary

Fixes #1497. `load_config()` in `nemo_skills/evaluation/utils.py` returned `yaml.safe_load(fin)` directly, which is `None` for an empty (or comments-only) YAML file. `get_eval_group()` passed that `None` through unchanged, so any caller doing `config["key"]` got a generic `TypeError: 'NoneType' object is not subscriptable` instead of a message pointing at the actual misconfigured file.

Per this repo's own contributing guidelines ("let the code fail with a clear error" rather than silently defaulting/warning), `load_config()` now raises `ValueError` naming the offending config path as soon as the empty result is detected.

## Out of scope (flagging, not fixing here)

`nemo_skills/prompt/utils.py` has a near-identical `load_config()` with the same `None`-on-empty-file gap (the target file even has a `# TODO: DRY with prompt's load_config function` comment acknowledging the duplication). Left untouched to keep this PR scoped to what #1497 reports — happy to open a follow-up if maintainers want the twin fixed too.

## Test plan

- Added `tests/test_evaluation_utils.py`: empty file, comments-only file, valid file, and `get_eval_group`'s pass-through/empty-file paths. All new tests fail on `main` and pass with the fix.
- `python -m pytest tests/test_evaluation_utils.py -v` — 5 passed.
- `ruff check` / `ruff format --check` on both changed files — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty or comment-only YAML configuration files now produce a clear validation error.
  - Valid YAML configurations are loaded and returned consistently.
  - Evaluation group handling now rejects empty files while preserving dictionary inputs.

- **Tests**
  - Added coverage for empty, comment-only, valid, and dictionary-based configuration inputs.


Ashfaqbs:
Added docstrings to the remaining touched functions (get_eval_group and the new test functions) to satisfy the docstring coverage check.

